PR #268, opened 2014-12-18
Jquery plugin

jbeezley:
This is a redesigned version of the jquery widget that is has been part of tangelo.  I'm putting up this PR to collect comments on the API.  Only the basics are implemented at the moment, but you can take it for a spin locally.  `npm install && grunt serve`.  Then go to [http://localhost:8082/examples/jquery-plugin/](http://localhost:8082/examples/jquery-plugin/) for an example.  Also try the docs at [http://localhost:8082/apidocs/$.fn.geojsMap.html](http://localhost:8082/apidocs/$.fn.geojsMap.html), but ignore the mess on the right side of the page.


jbeezley:
For posterity, the planned option object will look something like this:

```javascript
{
  data: [],
  layers: [
    {
      renderer: "...",
      features: [
        {
          type: "...",
          size: ...,
          position: ...,
          fillColor: ...,
          // etc.
        }
      ]
    }
  ]
}
```

Possibly add names to features/layers and provide a mechanism for feature specific data.

aashish24:
:+1: we will touch base on this soon. Thanks,

jbeezley:
This is nearly done.  I just need to put back in the auto color and size scaling.   The diff looks especially large because:
  1.  I merged in the `event-updates` branch which should go into master soon
  2.  I renamed registered names of renderers from `vglRenderer` to `vgl` and `d3Renderer` to `d3`.  The  `{renderer: 'vglRenderer'}` API was overly verbose.
  3.  I updated jsdoc which now properly supports documenting class hierarchies.  As a result, I added `@extends` tags all over the place.  The API documentation is actually starting to fill out now.

jbeezley:
@jeffbaumes PTAL.  The interface is now what we discussed.  For the data array, I decided to do a search:
  1. Use `data` inside the feature specification
  2. Use `data` inside the layer specification
  3. Use `data` provided in the plugin options
  4. Use `[]`

The features and styles aren't documented terribly well, but I hope to address that in #295.

aashish24:
@jbeezley I will do a detailed review of this today since it has lot of changes, it may take some time. It would be great if @jeffbaumes can give his thumbs up as well.

examples/animation.html
```diff
@@ -168,7 +168,7 @@
 
         // Load image to be used for drawing dots
         var layer = animatedLayer({
-            'renderer' : 'vglRenderer',
+            'renderer' : 'vgl',
```
aashish24: +1

examples/features.html
```diff
@@ -132,7 +132,7 @@
                         center : [0, 0]},
           myMap = geo.map(mapOptions),
           layer = myMap.createLayer('osm'),
-          d3Layer = myMap.createLayer('feature', {'renderer': 'd3Renderer'}),
+          d3Layer = myMap.createLayer('feature', {'renderer': 'd'}),
```
aashish24: d3?

jbeezley: Yup, good catch.

jbeezley:
@aashish24 I fixed the typo in the example.  Also, this branch depends on #285, so the actual diff here is [here](https://github.com/OpenGeoscience/geojs/compare/event-updates...jquery-plugin).  It is big, but most of the changes are related to documentation and renderer renaming.

jbeezley:
Rebased on master.

aashish24:
thanks. LGMT :+1:

PR #278, opened 2015-01-05
A basic legend implementation

jbeezley:
This is a basic implementation of a legend for categorical features.  See: http://opengeoscience.github.io/geojs/examples/legend/


aashish24:
:nice: will review this branch tomorrow

src/ui/legendWidget.js
```diff
@@ -0,0 +1,261 @@
...
+   * rectangle.
+   *
+   * @param {object[]?} categories The categories to display
+   */
+  //////////////////////////////////////////////////////////////////////////////
+  this.categories = function (arg) {
+    if (arg === undefined) {
+      return m_categories.slice();
+    }
+    m_categories = arg.slice().map(function (d) {
```
aashish24: should we think about slicing for other mappings as well?

src/ui/legendWidget.js
```diff
@@ -0,0 +1,261 @@
...
+      .attr('dy', '0.3em')
+      .text(function (d) {
+        return d.name;
+      });
+
+    return m_this;
+  };
+
+  //////////////////////////////////////////////////////////////////////////////
+  /**
```
aashish24: this is awesome. ..at some point we have to get this information from a look up table (once we get it from the cinema). Also, we may have to support gradients but all that could be done later (or may be in a derived class).

aashish24:
LGTM :100:

Issue #57, opened 2014-04-16
Clamp north and south, wrap around east and west


jeffbaumes:
So you never see white boundaries.

jbeezley:
Tile wrapping included as part of #410.

PR #283, opened 2015-01-06
Restricting panning to valid map regions

jbeezley:
This addresses the clamping part of #57.

### DIFF
--- a/src/core/map.js
+++ b/src/core/map.js
@@ -800,6 +800,10 @@ geo.map = function (arg) {
    */
   ////////////////////////////////////////////////////////////////////////////
   this._updateBounds = function () {
+    if (!m_this.baseLayer()) {
+      m_bounds = {};
+      return;
+    }
     m_bounds.lowerLeft = m_this.displayToGcs({
       x: 0,
       y: m_height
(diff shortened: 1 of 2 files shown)


jbeezley:
Yes, I was going to mention that.  The zoom behavior is difficult to manage because it is doing some panning internally.  What I have done allows the zoom to subvert the clamping.  To clamp on zoom as well would require making some fairly drastic changes to the renderer.  I'm hesitant to put in too much effort there because it will touch the same code that @aashish24 has been working on in the gl renderer.

With that said, I thought about adding a bounce back effect (like an over-damped spring) that could both provide some user feedback about hitting the edge and allow me to do the clamping after the renderer performs the zoom.

I guess I need to know to what extent does the current behavior inhibit user experience?  Is it okay for now or do I need to fix it before merging?

jeffbaumes:
I'd say to put the damping in for this because without it I don't really think the feature of restricting the view to just the world bounds (which I believe is what people want) is truly implemented. Some of the current allowed map states as implemented (like only being able to pan one direction) are unintuitive.

jbeezley:
Issue moved to #285.

PR #285, opened 2015-01-12
Event updates

jbeezley:
This PR includes several features needed to wrap up GRITS.
* Feature specific mouse events give more information:
  * a unique ID number
  * whether the current feature is on top or not
* New feature mouse events:
  * `geo.feature.mouseon`: When the mouse pointer is on the marker.  This is like mouseover, but only fires for the marker on top.  This is similar to the DOM's `mouseenter`.
  * `geo.feature.mouseoff`: Fires when the mouse pointer is no longer directly over the marker either because it exited the maker or another marker is now on top.
* Core event documentation done
* Implemented clamping of the map bounds to the visible window via a spring like animation.  I will probably revisit this later to move the map navigation animation to one place.  The current code is fragmented across `mapInteractor` and `map`, but it seems to work well enough for now.

### DIFF
--- a/src/core/map.js
+++ b/src/core/map.js
@@ -37,6 +37,7 @@ geo.map = function (arg) {
       m_interactor = null,
       m_validZoomRange = { min: 0, max: 16 },
       m_transition = null,
+      m_queuedTransition = null,
       m_clock = null;
 
 
@@ -618,23 +619,30 @@ geo.map = function (arg) {
   /**
    * Start an animated zoom/pan.
    *
-   * Options: ::
-   *
+   * Options:
+   * <pre>
    *   opts = {
    *     center: { x: ... , y: ... } // the new center
    *     zoom: ... // the new zoom level
    *     duration: ... // the duration (in ms) of the transition
    *     ease: ... // an easing function [0, 1] -> [0, 1]
    *   }
+   * </pre>
    *
-   * @param {Object} opts
+   * Call with no arguments to return the current transition information.
+   *
+   * @param {object?} opts
    * @returns {geo.map}
    */
   ////////////////////////////////////////////////////////////////////////////
   this.transition = function (opts) {
+
+    if (opts === undefined) {
+      return m_transition;
+    }
+
     if (m_transition) {
-      console.log("Cannot start a transition until the" +
-                  " current transition is finished");
+      m_queuedTransition = opts;
       return m_this;
     }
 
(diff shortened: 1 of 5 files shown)


aashish24:
This is next on my list -:)

jbeezley:
@jeffbaumes Perhaps you could QA the clamping behavior on this.  I'm serving it at [http://192.168.30.129:8082/examples/osm/](http://192.168.30.129:8082/examples/osm/).

I'm not completely happy with it because I had to bake in a number of hacks to prevent instability.  So, the result isn't as natural as it could be.  When I have time, I'll reorganize the code a bit and implement a higher order solver for the spring equation.

aashish24:
@jbeezley One thing we touched base was on implementing behavior like google maps. While we like the spring nature of things, It would be nice to have an option that will dis-allow you to go out (so no spring force is necessary in that). Something that switches between these two options would be nice.

jeffbaumes:
This is looking good - for some reason I'm getting the behavior where the map does not render the first time until I interact. Also, it seems new tiles do not pop in until the spring-back is completely finished. Other than those issues LGTM.

jeffbaumes:
@aashish24 the bounce-back was according to @jbeezley the quickest way to implement avoiding pan/zooming the map out of bounds, so I suggested he go ahead and do it since our project had limited time to get this feature.

jbeezley:
I'll look into the initial render.  I might have to force the osm layer to auto render after loading the initial tiles.

I would like to have the new tiles pop in during the animation as well.  I think the GL code is being too clever and preventing the new tiles from loading until the motion stops.  I tried throttling the frames in the animation, but that wasn't enough.  Perhaps @aashish24 can suggest a way to force the new tiles to render.

jbeezley:
@aashish24 I took out the bounds clamping/spring effect commits and rebased on master so we can get these changes in now.  I'll create a new branch for the rest.

aashish24:
Thanks, I will review this one today as well.

src/core/map.js
```diff
@@ -719,18 +728,34 @@ geo.map = function (arg) {
...
         m_transition = null;
+
+        m_this.geoTrigger(geo.event.transitionend, defaultOpts);
+
         if (done) {
           done();
         }
+
+        if (next) {
+          console.log(next);
```
aashish24: left over?

aashish24:
LGTM to me.. :+1:

PR #286, opened 2015-01-13
Add polygon feature to geojs

aashish24:


aashish24:
@jbeezley as discussed, please add support in d3 and fix issues such as download of the external library as the build time

src/gl/polygonFeature.js
```diff
@@ -0,0 +1,335 @@
...
+        intIndex += 1;
+      });
+
+      //console.log("extRing ", extRing);
+      //console.log("result", PolyK.Triangulate(extRing));
+      triangList = triangulator.triangulate_polygon(extRing);
+      newTriangList = [];
+
+      triangList.forEach(function (newIndices) {
+        newTriangList = newTriangList.concat(newIndices);
```
jbeezley: I think the following might be faster:
```javascript
Array.prototype.push.apply(newTriangList, newIndices);
```
It extends the array, rather than concatenating into a new array.  Although without preallocation, it is hard to say.

testing/test-cases/selenium-tests/glPolygons/testGlLines.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+
+import unittest
+
+from selenium_test import FirefoxTest, ChromeTest,\
+    setUpModule, tearDownModule
+
+
+class glLinesBase(object):
+    testCase = ('glLines',)
```
jbeezley: I'll have to update the names in the file when I generate the screenshots.

jbeezley:
Looks good.  I'll work on fixing the tests tomorrow.

src/gl/polygonFeature.js
jbeezley: Yup looks like push is about 10x faster for this use case.  http://jsperf.com/jb-concat-vs-push

PR #295, opened 2015-01-18
[WIP] Refactoring the feature style/attribute API

jbeezley:
As we add more complicated feature types to GeoJS, it is becoming clear that the current way we are handling style accessors is unmaintainable.  This PR is refactoring concept that should have relatively low impact on the user-facing API, but will make development and maintenance of features much easier.  I believe it will also speed up the rendering (for gl in particular) because it eliminates most of the array copy operations occurring at render time.

The idea here is that methods involving data dependent accessors (i.e. `point.position`, `polygon.polygon`, `feature.style`) will be generated and bound to subclasses via a "property" interface.  For example, `geo.point` would define its `position` method during construction as follows:
```javascript
this._property(
  'position',   // The public method name
  'position',   // The internal storage name
  'coordinate', // The value type
  {x: 0, y: 0}  // The default value
);
```
This would do several things:
1. Create a public `this.position` getter/setter
2. Create an empty internal array `m_property.position`
3. Generate api documentation

There is one breaking change that I want to make.  Accessors will no longer be called lazily at render time; instead they will applied immediately when given and the output from each accessor cached internally in the `m_property` object.  The base feature class should be smart enough to know when to do this in most cases, and it can do so efficiently by only updating the components that changed.  The benefits of this become even more apparent with classes containing subfeatures (such as polygon -> edges/holes -> positions) for which I'm creating a `container` property type (an accessor that returns the datum for a subfeature).

I'm still fleshing out the details, but I wanted to give heads up for anyone interested providing feedback.


aashish24:
@jbeezley seems like a good idea. But let's talk more about it sometime soon.

jeffbaumes:
Even though I am not to familiar with the prior code, this seems a step in the right direction. I see right now this adds a net of over 250 LOC. But will this common property code end us saving us complexity and LOC in the long run?

jbeezley:
It will save major sections of duplicated code in the feature subclasses.  For example, every feature subclass has loops like [this](https://github.com/OpenGeoscience/geojs/blob/master/src/gl/polygonFeature.js#L118) that is responsible for evaluating the style accessors.  The 68 LOC for that loop will go away as well as a similar loop that I would have to write for the d3 version of the polygonFeature.

These loops have also been a major source of bugs, which are hard to test for because they require evaluating them in selenium and comparing screenshots.  By abstracting out the renderer, I can make jasmine tests that will execute in phantomjs.

jeffbaumes:
That loop does indeed look ugly, and I can see how you would need similar loops in several places. From my point of view, carry on!

aashish24:
we may not be able to get rid of the loop in the polygon feature as it does triagulization and for that it has to create a separate data structure.

jbeezley:
For that, I have the `setter` argument where the class can inject custom code into the loop.  I originally put that in to handle flattening objects into arrays as required for GL, but it can be used for triangularization, too.

aashish24:
Closing this one until we get back to it.

PR #299, opened 2015-01-29
Bounds clamping on zoom and pan

jbeezley:
This PR adds proper bounds clamping to the map navigation.  The clamping is enforced on pan, zoom, and resize.  This behavior can be disabled by setting the map option `clampBounds` to `false`.

Depends on #285.  See [here](https://github.com/OpenGeoscience/geojs/compare/event-updates...bounds-clamping) for changes added by this PR.

### DIFF
--- a/src/core/map.js
+++ b/src/core/map.js
@@ -39,7 +39,8 @@ geo.map = function (arg) {
       m_transition = null,
       m_queuedTransition = null,
       m_clock = null,
-      m_bounds = {};
+      m_bounds = {},
+      m_zoomCallback = null;
 
 
   arg.center = geo.util.normalizeCoordinates(arg.center);
@@ -134,6 +135,9 @@ geo.map = function (arg) {
     });
 
     m_this.modified();
+    if (m_zoomCallback) {
+      m_zoomCallback();
+    }
     return m_this;
   };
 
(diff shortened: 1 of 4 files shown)


jeffbaumes:
Do you have a live link for (lazy) me to try out this change?

jbeezley:
[http://192.168.30.129:8083/examples/osm/](http://192.168.30.129:8083/examples/osm/)

jbeezley:
Rebased on master.

jeffbaumes:
Nearly works. The edge case I'm seeing is that the initial load (on a large screen) seems to allow the map to be out of bounds, which gets fixed after the first interaction. This may be a bit tricky to solve, since you will need to resolve this by finding the nearest "good" view based on the user's initial request, which may involve a pan and possibly a zoom if the zoom level requested is too zoomed-out to fill the viewport.

jbeezley:
I tried to handle that... I'll take another look at it.  :weary:

jbeezley:
@jeffbaumes Try it again.  I think I got it this time. :pray:

jeffbaumes:
Huzzah! Seems to be working for me. Can't really speak to implementation details but the result LGTM.

jbeezley:
Great!  Thanks for the QA.

aashish24:
LGTM :+1:

PR #302, opened 2015-02-03
Fixing exit methods so deleted features get garbage collected

jbeezley:
Note, I made `geo.map`'s exit method public because it attaches itself to `window.resize` and won't go out of scope unless the user explicitly deletes it.


src/core/init.js
```diff
@@ -11,37 +11,22 @@ geo.fileReaders = {};
...
-//////////////////////////////////////////////////////////////////////////////
-/**
- * Convenient function to get size of an object
- *
- * @param {object} obj
- * @returns {number} *
- */
-//////////////////////////////////////////////////////////////////////////////
-Object.size = function (obj) { // jshint ignore: line
+geo.inherit.func = function () {
```
aashish24: Do we not use Object.size ?

jbeezley: No, we don't.  But also 1. this is probably a bad idea in general, and 2. this is defined in vgl init anyway.

aashish24: Curious. Why you think its bad. JS supposed be extended. A similar code is suggested here: http://stackoverflow.com/questions/5223/length-of-a-javascript-object-that-is-associative-array

jbeezley: [Mozilla](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Inheritance_and_the_prototype_chain#Bad_practice.3A_Extension_of_native_prototypes) says so.  I suppose people can argue for it, but I don't like it, especially for a common method name like `size`.  Collisions between libraries for things like this would be a nightmare to debug.

jbeezley: I just noticed that this a static method on Object.  In that case, it isn't so bad as the stackoverflow answer mentions.  Still though, I would rather minimize the global footprint.

src/core/renderer.js
```diff
@@ -200,23 +200,6 @@ geo.renderer = function (arg) {
   this._render = function () {
   };
 
-  ////////////////////////////////////////////////////////////////////////////
-  /**
-   * Exit
-   */
-  ////////////////////////////////////////////////////////////////////////////
-  this._exit = function () {
```
aashish24: why not we keep the _exit API on the renderer and use it for the purpose of cleaning up and call it from the deleteCache? Also, having this API is consistent with the rest of the objects.

src/core/renderer.js
```diff
@@ -200,23 +200,6 @@ geo.renderer = function (arg) {
...
-  };
-
-  ////////////////////////////////////////////////////////////////////////////
-  /**
-   * Connect mouse events to the map layer
-   */
-  ////////////////////////////////////////////////////////////////////////////
-  this._connectMouseEvents = function () {
-
-  };
```
aashish24: +1

src/core/init.js
aashish24: Right. For others, after our discussions we think that extending base objects may not be as bad (the JS community has different opinions on this) but we will avoid doing so as much as possible and if we do, we will use underscore to mark it special. Also, we discussed about using underscore library in the future which I think is a low priority for now.

src/core/map.js
```diff
@@ -527,12 +526,15 @@ geo.map = function (arg) {
...
+    m_this.node().off(".geo");
+    $(window).off("resize", resizeSelf);
+    s_exit();
   };
 
   this._init(arg);
 
   // set up drag/drop handling
-  this.node().on("dragover", function (e) {
+  this.node().on("dragover.geo", function (e) {
```
aashish24: geo.dragover? I thought we prefix geo.

aashish24:
Mostly looks good. Just curious about geo suffix (vs prefix).

src/core/map.js
jbeezley: This is how jquery namespaces events.  `geo.dragover` would mean event `geo` with namespace `dragover`.  I changed our event names a while back to use [underscores](https://github.com/OpenGeoscience/geojs/blob/master/src/core/event.js#L49) to avoid the confusion.

jbeezley:
@aashish24 Merged in master and fixed a couple of small conflicts.

PR #303, opened 2015-02-04
Added names to sources.

manthey:
Also, pass a 'dynamicDraw' option to the mapper.

This isn't very useful without the companion changes to vgl.  It should do nothing without them.


src/gl/pointFeature.js
```diff
@@ -279,34 +280,34 @@ geo.gl.pointFeature = function (arg) {
     }
 
     sourcePositions.pushBack(buffers.get("pos"));
-    geom.addSource(sourcePositions);
+    geom.addSource(sourcePositions, "pos");
```
aashish24: @manthey can we do this instead: 

sourcePositions = vgl.sourceDataP3fv("pos") 

Basically at the time of the creation of source array, we pass an argument. If not then internally we should create a name for it (using may be time of the creation)? I think this might be better.

src/gl/pointFeature.js
```diff
@@ -318,6 +319,20 @@ geo.gl.pointFeature = function (arg) {
 
   ////////////////////////////////////////////////////////////////////////////
   /**
+   * Return mapper where actor gets it behavior and data
+   *
+   * @returns {vgl.mapper}
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this.mapper = function () {
```
aashish24: Its probably better if we return actors (as then we can move that API in a common GL class). This would be because a feature like graph may have more than one actor. Also, in the future it is possible that point feature will split the data into many actors (for performance reasons). 

this.actors = function() {
return [m_actor]; 
}

What you think of it? 

Thanks for pushing these changes.

src/gl/pointFeature.js
manthey: @aashish24 I've pushed a change to vgl and to here that allows passing additional arguments to all the sources.  I'm happy to change it if you'd rather have something different.

src/gl/pointFeature.js
manthey: Changed.

jbeezley:
I like the latest changes, I'll try to pull some of this into a `geo.gl.feature` class with my property refactor.  We should merge vgl first to fix the SHA.  For that, I'll defer to @aashish24.

aashish24:
thanks for considering the suggestions. Loosk good to me :+1:

PR #304, opened 2015-02-05
Report the number of vertices used per point.

manthey:
Report the number of vertices used per point so that the internal use can be abstracted.


aashish24:
LGTM I believe 5bba24fabe460807bc8f66c1e250aafd031b1691 VGL sha now points to master?

PR #306, opened 2015-02-11
d3 vector feature and core API

jbeezley:
@aashish24 Here is a first pass at the vector feature.  I decided to implement it via svg `marker` elements because it makes the code extremely simple.  The problem I found is that `marker`'s can't inherit styles yet.  This feature has been defined in SVG2 with `context-fill` with `context-stroke`, but it hasn't appeared in any browsers yet.  For the moment, the vector lines can be colored with `strokeColor`, but the arrows will always be black.  I think I will backfill the `context-stroke` property by using the method used in [this example](http://bl.ocks.org/kenpenn/8d782030e4be9d832be7).

The API works like this:
```javascript
layer.createFeature('vector')
  .data(...)
  .origin(...) // The origin of the vector in world coordinates
  .delta(...) // The vector (dx/dy) in arbitrary coordinates (see explanation)
```

This is the default style
```javascript
{
  strokeColor: 'black',
  strokeWidth: 2,
  strokeOpacity: 1,
  scale: null
}
```

The scale is a factor that is multiplied by each of the vector deltas to get the length in pixels.  A null value indicates that the renderer will decide how to scale based on the current zoom, density of the vectors, etc.  This is designed to be analogous to quiver plots in matplotlib.  We could also add a special value to indicate the delta is in world coordinates as well.

One final thing that I want to add is a style attribute for the arrows.  This will enable creating the sorts of arrows that meteorologists like to use that indicate wind speed.

vectors

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ if (COVERAGE_TESTS)
   file(COPY "${GEOJS_DEPLOY_DIR}/src" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
 else()
 
-  set(SOURCES_INCLUDE <script src=/built/geo.all.js></script>)
+  set(SOURCES_INCLUDE <script src=/built/geo.all.dev.js></script>)
   set(BLANKET_INCLUDE "")
 
 endif()
--- a/sources.json
+++ b/sources.json
@@ -30,6 +30,7 @@
         "pathFeature.js",
         "polygonFeature.js",
         "planeFeature.js",
+        "vectorFeature.js",
         "geomFeature.js",
         "graphFeature.js",
         "transform.js",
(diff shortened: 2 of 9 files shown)


jeffbaumes:
Looks nice. Important question: what are all the arrows pointing at??

ronichoudhury:
I assume the spacecraft buried under @jbeezley's hometown :alien:

jbeezley:
Sssshhh.  No one is supposed to find out about that.

Actually, it is the [test case](https://github.com/OpenGeoscience/geojs/pull/306/files#diff-f326e097931d734cf6c9797e77123608R21) where the vectors always point at the map center.

jeffbaumes:
Clever.

aashish24:
Very nice @jbeezley. Like the API. 

* +1 Adding style for the arrows in the future would be nice. 
* May be a style for tail of arrows (so that it can be drawn curved like in VTK) 

LGTM

PR #309, opened 2015-02-12
Adding a link to the new api docs

jbeezley:
I uploaded the latest jsdoc generated [documentation](http://opengeoscience.github.io/geojs/apidocs/).  There is still a long way to go in writing out the docstrings across the classes and making them consistent, but they are starting to get better.  I also need to look into customizing the output template to both make the docs look better with our class structure.

### DIFF
--- a/examples/index.jade
+++ b/examples/index.jade
@@ -20,7 +20,8 @@ block prepend headerButtons
         li
           a(href="https://github.com/OpenGeoscience/geojs") Source
           a(href="https://github.com/OpenGeoscience/geojs/issues") Bugs
-          a(href="http://geojs.readthedocs.org/en/latest/") Documentation
+          a(href="http://opengeoscience.github.io/geojs/apidocs/") Documentation
+          a(href="http://geojs.readthedocs.org/en/latest/") Developers guide
           a(href="http://my.cdash.org/index.php?project=geojs") Testing
 
 block mainContent


jeffbaumes:
Looks really slick.

ronichoudhury:
This is awesome.  Is it possible to blend this somehow into the Sphinx docs that go up to readthedocs?

jbeezley:
Yes, I started working in that direction a while back.  It is a matter of
creating an output template for sphinx compatible restructured text.  I
decided to use this for now as a fully working sphinx template will take a
while.  In the mean time, I have been watching the development of jsdoc for
internal features that will make the task easier.

ronichoudhury:
Cool.  While I'm being obnoxious about documentation, I thought I might
suggest starting a "cookbook" or tutorials section to the documentation.
The hurricane demo would be a perfect one to start with.  We've got one for
the Tangelo docs and I think it helps to knock out bugs in the user's
experience of creating these things from scratch.

jbeezley:
Agreed, though I think the hurricane one is a bit too complicated.  It is mostly d3 magic anyway.  I have been toying with the idea of creating a bl.ocks style site that will list of bunch of quick one-off demos to show how to do simple things from gists.  I eventually want to reserve the examples page for more full featured applications like the hurricane.

Also, as part of the documentation, I want to add many more examples in the docstrings themselves.  See the [jQuery plugin](http://opengeoscience.github.io/geojs/apidocs/jQuery.fn.geojsMap.html) for what I would like to build to.

jeffbaumes:
Just to throw a wild idea out there - maybe we should have Sitar support both Vega and the GeoJS jQuery plugin input format, which is also essentially JSON though we may need a little function serialization library. The "editor" for GeoJS would just be an ace editor for the JSON text for now. Basically, take @ronichoudhury and @jbeezley's ideas on a bl.ocks-like interfaces for vis and combine them into something awesomer.

ronichoudhury:
This is a great idea @jeffbaumes.

I wonder if we could (well down the road) cook up a "Lyra for GeoJS" based on this idea.  We could name it after a star/constellation that early sailors used for navigation purposes (to keep up with the Vega -> Lyra -> ... name sequence).

jbeezley:
That was kind of the idea behind [this example](http://opengeoscience.github.io/geojs/examples/geoJSON/), but I would be quite happy to make something awesomer.

PR #310, opened 2015-02-12
Adding a hurricane example

jbeezley:
[http://opengeoscience.github.io/geojs/examples/hurricanes/](http://opengeoscience.github.io/geojs/examples/hurricanes/)

I did this a while back.  It is probably worth keeping around as a more complicated example application.  The line highlighting is a bit pokey due to the usual gl buffer issues.  There is a lot more I can add as time permits such as filtering the data by date and name.   For now it is a nice example of lines and legends used together with a custom d3 layer.


ronichoudhury:
I get "Uncaught TypeError: Cannot read property 'enable' of null" in renderer.js:173

jbeezley:
Hmmm... I don't see where that could come from... You have a stack trace with that?

jbeezley:
It's probably [this line](https://github.com/OpenGeoscience/vgl/blob/5bba24fabe460807bc8f66c1e250aafd031b1691/src/renderer.js#L173) in vgl.  I'lll see if I reproduce it somehow.

ronichoudhury:
Well it turns out webgl is not enabled in my browser (no idea why... it used to be enabled).

But, maybe GeoJS should do a check for webgl being available when it first loads?  Or, more likely, the example applications should perform this check so they can show a nice error message letting the client know what's wrong.

In the meantime, I'll see if I can figure out how to get it working for me again.

jbeezley:
Yeah, I'll put in an issue about that.

ronichoudhury:
Meanwhile, it works great on firefox :crying_cat_face:

Cool demo!

aashish24:
Thanks for adding it! Looks great.

PR #318, opened 2015-02-19
Quick fix to minimize pixel distortion

aashish24:
@jbeezley, we will have to update the baselines. On my machine the rendering tests mostly fail. It might be better if you upload the new baselines if approved.

### DIFF
--- a/src/core/osmLayer.js
+++ b/src/core/osmLayer.js
@@ -442,8 +442,7 @@ geo.osmLayer = function (arg) {
       worldDelta = worldDeltaY;
     }
 
-    noOfTilesRequired = Math.pow(2, Math.floor(
-      Math.log2(Math.round(displayDelta / 256.0))));
+    noOfTilesRequired = Math.round(displayDelta / 256.0);
     worldDeltaPerTile = worldDelta / noOfTilesRequired;
 
     /// Minimize per pixel distortion
--- a/src/core/osmLayer.js
+++ b/src/core/osmLayer.js
@@ -446,7 +446,7 @@ geo.osmLayer = function (arg) {
     worldDeltaPerTile = worldDelta / noOfTilesRequired;
 
     /// Minimize per pixel distortion
-    for (i = 20; i > 2; i = i - 1) {
+    for (i = 20; i >= 2; i = i - 1) {
       distWorldDeltaPerTile = Math.abs(360.0 / Math.pow(2, i) - worldDeltaPerTile);
       if (!minDistWorldDeltaPerTile ||
           distWorldDeltaPerTile < minDistWorldDeltaPerTile) {
--- a/src/core/osmLayer.js
+++ b/src/core/osmLayer.js
@@ -446,17 +446,13 @@ geo.osmLayer = function (arg) {
     worldDeltaPerTile = worldDelta / noOfTilesRequired;
 
     /// Minimize per pixel distortion
+    minDistWorldDeltaPerTile = Number.POSITIVE_INFINITY;
     for (i = 20; i >= 2; i = i - 1) {
       distWorldDeltaPerTile = Math.abs(360.0 / Math.pow(2, i) - worldDeltaPerTile);
-      if (!minDistWorldDeltaPerTile ||
-          distWorldDeltaPerTile < minDistWorldDeltaPerTile) {
+      if (distWorldDeltaPerTile < minDistWorldDeltaPerTile) {
         minDistWorldDeltaPerTile = distWorldDeltaPerTile;
         m_zoom = i;
       }
-    }
-
-    if (m_zoom < 2) {
-      m_zoom = 2;
     }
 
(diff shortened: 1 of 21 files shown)


src/core/osmLayer.js
```diff
@@ -441,23 +429,58 @@ geo.osmLayer = function (arg) {
...
+    displayDeltaX = urx - llx;
+    displayDeltaY = lly - ury;
+
+    /// Reuse variables
+    if (displayDeltaY > displayDeltaX) {
+      displayDeltaX = displayDeltaY;
+      worldDelta = worldDeltaY;
+    }
+
+    if (Math.log2) {
```
jbeezley: We already have a polyfill for this.  https://github.com/OpenGeoscience/geojs/blob/master/src/core/init.js#L219

aashish24:
I will push the style fixes momentarily.

src/core/osmLayer.js
aashish24: Thanks, I remembered about it, will push it in my clean up commit.

aashish24:
Force updated a clean version with Math.log2 fix as well.

src/core/osmLayer.js
```diff
@@ -441,23 +429,55 @@ geo.osmLayer = function (arg) {
...
+      worldDelta = worldDeltaY;
+    }
+
+    noOfTilesRequired = Math.pow(2, Math.floor(
+      Math.log2(Math.round(displayDeltaX / 256.0))));
+    worldDeltaPerTile = worldDelta / noOfTilesRequired;
+
+    /// Minimize per pixel distortion
+    for (i = 20; i > 2; i = i - 1) {
+      if (!minDistWorldDeltaPerTile ||
```
jbeezley: This is a minor point because it is unlikely, but if `minDeltaworldDeltaPerTile` happened to be `0` from the previous iteration, it will get reset here.  I usually do something like this in this case:
```javascript
minDeltaworldDeltaPerTile = Number.POSITIVE_INFINITY;
for ( i = 20; i > 2; i = i - 1) {
  minDeltaworldDeltaPerTile = Math.min(
    minDeltaworldDeltaPerTile,
    Math.abs(360.0 / Math.pow(2, i) - worldDeltaPerTile)
  )
}
```

src/core/osmLayer.js
```diff
@@ -441,23 +429,55 @@ geo.osmLayer = function (arg) {
...
+            Math.abs(360.0 / Math.pow(2, i) - worldDeltaPerTile) <
+           minDistWorldDeltaPerTile) {
+        minDistWorldDeltaPerTile =
+          Math.abs(360.0 / Math.pow(2, i) - worldDeltaPerTile);
+        m_zoom = i;
+      }
+    }
+
+    if (m_zoom < 2) {
```
jbeezley: `m_zoom` is always `>= 3` after the above loop.

aashish24: Right, that is intentional. Low level zoom levels are almost useless as they are too pixelated. Also, their at 0 and 1 (as well), the size of the texture is not 256 (but I have to verify again).

jbeezley: My comment was about the if statement.  The condition is always `false`.

aashish24:
@jbeezley don't merge yet, I am still making some tweaks. Just wanted to push it so that I can get some feedback.

jbeezley:
Roger that.  I can update the baselines.  I will just need to increment the revisions in this branch.  I may adjust some of the tests to use a white background so they don't depend so much on tiny differences in the osm tiles.

aashish24:
in OpenStreetMap
osm

in GeoJS
geojs

There is still fuziness, but that is hard to get rid of unless we switch to Orthographic projection or use vector based text rendering.

aashish24:
Okay, now is ready for merge. I will revisit / improve it further in the future.

aashish24:
Also, this fixes the issue of resize. Now when you resize the OSM layer will auto adjust to minimize the sampling artifacts.

jbeezley:
It looks a lot better, but there is an issue when viewing in a small window.

screen shot 2015-02-19 at 10 34 54 am

Could we allow level 2 and maybe level 1 tiles?

aashish24:
Sure level 2 should be fine. 1 is definately worse.. Let me push a fix.

aashish24:
Pushed. Try it now, if needed I can dig down to level 1

jbeezley:
I see what you mean.  You don't get any real improvement at level 1.  It just gets blurry.

screen shot 2015-02-19 at 10 48 11 am
screen shot 2015-02-19 at 10 47 14 am

jbeezley:
Baselines updated.

aashish24:
thansk @jbeezley !

jeffbaumes:
@aashish24 - just what I always wanted! This looks much much better on a high-res monitor than before.

PR #324, opened 2015-02-24
Added a function to change baseUrl.

manthey:
To change the tile set, call updateBaseUrl on the osm layer.  This moves all existing tiles to the zoom level of 'old' so that they will not be used, then updates the layer.

### DIFF
--- a/src/core/osmLayer.js
+++ b/src/core/osmLayer.js
@@ -308,6 +308,7 @@ geo.osmLayer = function (arg) {
     tile.LOADED = false;
     tile.REMOVED = false;
     tile.REMOVING = false;
+    tile.INVALID = false;
 
     tile.crossOrigin = "anonymous";
     tile.zoom = zoom;
@@ -548,6 +549,9 @@ geo.osmLayer = function (arg) {
       m_this.addDeferred(defer);
 
       return function () {
+        if (tile.INVALID) {
+          return;
+        }
         tile.LOADING = false;
         tile.LOADED = true;
         if ((tile.REMOVING || tile.REMOVED) &&
@@ -714,6 +718,7 @@ geo.osmLayer = function (arg) {
         for (x in m_tiles[zoom]) {
           for (y in m_tiles[zoom][x]) {
             tile = m_tiles[zoom][x][y];
+            tile.INVALID = true;
             m_this.deleteFeature(tile.feature);
           }
         }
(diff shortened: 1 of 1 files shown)


jbeezley:
This is great.  I'll add a quick test as part of the osmLayer selenium test already in place.  Thanks @manthey.

aashish24:
Nice. Quick question . Why keep the old tiles? Just wanted to get your thoughts.

manthey:
Because I wasn't certain if removing them would have a negative effect: tiles that are not fully loaded will still call tileOnLoad.  At that point, if they aren't present, we would need logic in that routine to handle it.  By keeping the, the existing garbage collection will handle this.

jbeezley:
Right. I think this is probably the path of least resistance here.

aashish24:
If it is not too much work: I prefer we remove them. You can set the flag on the old ones because if on load gets called it may interfere with the current tiles. But j I havenot reviwed the branch very carefully

jbeezley:
I'll look into it a bit later.

manthey:
I'll refactor to remove the tiles later today or tomorrow.

aashish24:
Sounds good, thanks. If it looks good then please merge and create new issue for it. I agree it's least resistance for now

aashish24:
thanks @manthey. I am reviewing it now

aashish24:
Looks good @manthey. I will just wait for the travis and then will merge it.

manthey:
@aashish24 looks like Travis is stuck -- it passed, but isn't sending the results back to github.

jbeezley:
Well, that's a new one.  I say merge it.

aashish24:
yeah seems like it.. merging it fornow.

jeffbaumes:
Absolutely. When in doubt, merge! :smile:

aashish24:
@manthey one thought I had. Currently we list only one baseurl: m_baseUrl = "http://tile.openstreetmap.org/". Would it be possible for you to add list of good URL's (with one being the default). In that case, user can query list of possible URL's and use them appropriately. It will be helpful to provide this list for new users of GeoJS. The number of lines of code will be few. Thoughts?

jbeezley:
I was planning to add an example with a widget for switching between tile servers.  I would probably do #279 and #281 along with this.

aashish24:
@jbeezley that would be great.

PR #330, opened 2015-02-26
Added dynamic drawing capability to line features.

manthey:
Added a verticesPerFeature() method.

Refactored the code which generated the pair of triangles per line segment to use an array of information.  This allows a caller to determine which vertices of the triangles belong to which end of the line.  I expect that I will need this to dynamically change one end of a line without changing the other.

Added code to avoid computations on hidden lines.

### DIFF
--- a/src/gl/lineFeature.js
+++ b/src/gl/lineFeature.js
@@ -54,6 +54,12 @@ geo.gl.lineFeature = function (arg) {
 
         'void main(void)',
         '{',
+        /* If any vertex has been deliberately set to a negative opacity,
+         * skip doing computations on it. */
+        '  if (strokeOpacity < 0.0) {',
+        '    gl_Position = vec4(2, 2, 0, 1);',
+        '    return;',
+        '  }',
         '  const float PI = 3.14159265358979323846264;',
         '  vec4 worldPos = projectionMatrix * modelViewMatrix * vec4(pos.xyz, 1);',
         '  if (worldPos.w != 0.0) {',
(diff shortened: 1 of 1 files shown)


src/gl/lineFeature.js
```diff
@@ -225,22 +240,17 @@ geo.gl.lineFeature = function (arg) {
 
     i = 0;
     k = 0;
+    var order = m_this.featureVertices();
```
aashish24: just to be consistent, can we move the declaration at the beginning? Also, same for the addVert method as well.

src/gl/lineFeature.js
```diff
@@ -225,22 +240,17 @@ geo.gl.lineFeature = function (arg) {
...
-        addVert(prev[i - 1], position[i - 1], next[i - 1], -1,
-                strkWidthArr[i - 1], strkColorArr[i - 1], strkOpacityArr[i - 1]);
-
-        addVert(prev[i - 1], position[i - 1], next[i - 1], 1,
-                strkWidthArr[i - 1], strkColorArr[i - 1], strkOpacityArr[i - 1]);
-        addVert(prev[i], position[i], next[i], 1,
-                strkWidthArr[i], strkColorArr[i], strkOpacityArr[i]);
-        addVert(prev[i], position[i], next[i], -1,
-                strkWidthArr[i], strkColorArr[i], strkOpacityArr[i]);
+        for (v = 0; v < order.length; v += 1) {
```
aashish24: nice optimization!

src/gl/lineFeature.js
```diff
@@ -274,6 +284,28 @@ geo.gl.lineFeature = function (arg) {
...
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Return the number of vertices used for each line segment.
+   *
+   * @returns {Number}
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this.verticesPerFeature = function () {
+    return this.featureVertices().length;
```
aashish24: +1

aashish24:
Thanks @manthey. Looks good to me :+1:

PR #331, opened 2015-02-27
Updated VGL to access bin skipping feature

aashish24:


aashish24:
@jbeezley when you get a chance, please review it. It requires https://github.com/OpenGeoscience/vgl/pull/69

src/core/osmLayer.js
```diff
@@ -55,6 +59,10 @@ geo.osmLayer = function (arg) {
     m_imageFormat = arg.imageFormat;
   }
 
+  if (arg && arg.displayLast !== undefined) {
```
jbeezley: I'd change this to  `(arg && arg.displayLast !== undefined && displayLast)`.

aashish24: Sure.. you want to check if this check (arg.displayLast) results to true? (arg && arg.displayLast !== undefined && arg.displayLast. So does it mean that arg.displayLast could result to false as well? Just wanted to make sure I understand your intentions.

jbeezley: Yeah, I think that would be the reasonable expectation.  If it isn't there, then default to true, otherwise use it as a boolean.

jbeezley:
Falsy is used by Douglas Crockford to describe a javascript object that evaluates as false in a conditional.  I seem to have adopted it into my vocabulary.

jbeezley:
It looks fine to me.  I don't notice any performance improvements, but it couldn't hurt.

aashish24:
Sounds good. I see. It fixes this issue for me. 
ist-geojs

and now 
ist2

aashish24:
On this note, I seems faster on my machine. I am assuming that you clear the cache?

PR #333, opened 2015-02-27
Speed up point rendering.

manthey:
Speed up point rendering.

If some points are hidden because they have neither stroke nor fill, or because they have no opacity or size, then set gl_Position to outside of the view to skip the bulk of the computation for it.

Optionally use an equilateral triangle surrounding a point rather than a square.  This reduces the number of vertices.

Have the fragment shader terminate when outside of the radius.

I ran two test cases on my hardware to get some speed numbers.  Load time is the time it takes the point layer and a map layer to render.  I have an animation which can hide some of the points.

First test: 300,000 points, radius 5 pixels, no stroke, showing roughly 150,000 - 200,000 points per animation frame.
Second test: 100,000 points, radius 20 pixels, stroke 5 pixels, showing all 100,000 points on each animation frame (only the fill opacity was changed between frames).

Point rendering mode | Test 1 Load | Test 1 FPS | Test 2 Load | Test 2 FPS
-----|-----|-----|-----|-----
Original | 5.7s | 19 fps | 1.9s | 3.8 fps
New - squares | 5.7s | 24 fps | 1.9s | 4.0 fps
New - triangles | 3.3s | 30 fps | 1.1s | 4.5 fps

I expect that with large enough points, eventually squares will out-perform triangles, but they would have to be very big points.

The measured framerate was the best the animation could achieve without *any* dropped frames.  For the first test, I measured to the nearest 1 fps, for the second to the nearest 0.1 fps.

### DIFF
--- a/src/gl/pointFeature.js
+++ b/src/gl/pointFeature.js
@@ -53,18 +53,32 @@ geo.gl.pointFeature = function (arg) {
       "varying float strokeVar;",
       "void main(void)",
       "{",
-      "  unitVar = vec3 (unit, 1.0);",
+      "  strokeWidthVar = strokeWidth;",
+      "  // No stroke or fill implies nothing to draw",
+      "  if (stroke < 1.0 || strokeWidth <= 0.0 || strokeOpacity <= 0.0) {",
+      "    strokeVar = 0.0;",
+      "    strokeWidthVar = 0.0;",
+      "  }",
+      "  else",
+      "    strokeVar = 1.0;",
+      "  if (fill < 1.0 || rad <= 0.0 || fillOpacity <= 0.0)",
+      "    fillVar = 0.0;",
+      "  else",
+      "    fillVar = 1.0;",
+      /* If the point has no visible pixels, skip doing computations on it. */
+      "  if (fillVar == 0.0 && strokeVar == 0.0) {",
+      "    gl_Position = vec4(2, 2, 0, 1);",
+      "    return;",
+      "  }",
       "  fillColorVar = vec4 (fillColor, fillOpacity);",
       "  strokeColorVar = vec4 (strokeColor, strokeOpacity);",
-      "  strokeWidthVar = strokeWidth;",
-      "  fillVar = fill;",
-      "  strokeVar = stroke;",
+      "  unitVar = vec3 (unit, 1.0);",
       "  radiusVar = rad;",
       "  vec4 p = (projectionMatrix * modelViewMatrix * vec4(pos, 1.0)).xyzw;",
       "  if (p.w != 0.0) {",
-      "    p = p/p.w;",
+      "    p = p / p.w;",
       "  }",
-      "  p += (rad + strokeWidth) * ",
+      "  p += (rad + strokeWidthVar) * ",
       "vec4 (unit.x * pixelWidth, unit.y * pixelWidth * aspect, 0.0, 1.0);",
       "  gl_Position = vec4(p.xyz, 1.0);",
       "}"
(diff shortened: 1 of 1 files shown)


jbeezley:
This is awesome! Thanks @manthey.  I'll leave it to @aashish24 to review the GL code.

jeffbaumes:
+1 for slick markdown table. The performance improvements look good too :)

aashish24:
thanks @manthey this is very much aligned to what we have discussed. Reviewing it now.

aashish24:
@manthey the table looks good!. How did you measure the FPS? Did you start the counter leaving some of initial frame times (as that would include the load time as well).

src/gl/pointFeature.js
```diff
@@ -26,6 +26,7 @@ geo.gl.pointFeature = function (arg) {
       m_pixelWidthUniform = null,
       m_aspectUniform = null,
       m_dynamicDraw = arg.dynamicDraw === undefined ? false : arg.dynamicDraw,
+      m_primitiveShape = arg.primitiveShape || "triangle",
```
aashish24: Can we document above the object definition that arg can have keys such as triangle and square (and may be check if none of the above two but something else, then it should be an error)

src/gl/pointFeature.js
```diff
@@ -53,18 +54,32 @@ geo.gl.pointFeature = function (arg) {
       "varying float strokeVar;",
       "void main(void)",
       "{",
-      "  unitVar = vec3 (unit, 1.0);",
+      "  strokeWidthVar = strokeWidth;",
+      "  // No stroke or fill implies nothing to draw",
+      "  if (stroke < 1.0 || strokeWidth <= 0.0 || strokeOpacity <= 0.0) {",
```
aashish24: +1

src/gl/pointFeature.js
```diff
@@ -136,21 +145,37 @@ geo.gl.pointFeature = function (arg) {
...
-        x + w, y + h
-      ];
+  var pointPolygon = function (x, y, w, h) {
+    var verts;
+    switch (m_primitiveShape) {
+      case "triangle":
+        /* Use an equilateral triangle.  While this has 30% more area than a
+         * square, the reduction in vertices should help more than the
+         * processing the additional fragments. */
+        verts = [
```
aashish24: Looks correct +1

src/gl/pointFeature.js
```diff
@@ -136,21 +145,37 @@ geo.gl.pointFeature = function (arg) {
...
-        x + w, y + h,
-        x - w, y - h,
-        x + w, y - h,
-        x + w, y + h
-      ];
+  var pointPolygon = function (x, y, w, h) {
+    var verts;
+    switch (m_primitiveShape) {
+      case "triangle":
+        /* Use an equilateral triangle.  While this has 30% more area than a
```
aashish24: Could we please change it to // vs /* (just to be consistent)

aashish24:
@manthey mostly cosmetic requests.. other than than it looks good!

src/gl/pointFeature.js
aashish24: I am okay if we are just checking for triangle argument only 

if (arg.primitiveShape && (arg.primitiveShape == "square" || ... "triangle") {
} else {
// default to triangle
}

manthey:
To measure the fps: in the taxi map program, I have a target FPS.  After the first frame is drawn (which encompasses all the load time and setup for the animation), it starts keeping track of when it should render a frame.  It sets a `window.setTimeout` to callback at the appropriate time (rounded to the nearest ms).  In the callback, it adjusts values and renders the frame.  After the render, it checks how much time is available before it should render the next frame.  If this is less than zero, then it has missed a frame (and that frame is skipped).  I log how many frames were rendered and how many skipped.

This means that if the javascript is doing something, I can miss a frame through no fault of the rendering speed (for instance, garbage collection will cause skipped frames, so it shows if memory is being allocated and deallocated).

For this report, I check each integer to 1/10th integer frame rate to see what frame rate could be maintained without *any* drops.

I intend to make an automated test of this sort.

aashish24:
a test in the repo would be great!

aashish24:
LGTM :+1:

PR #337, opened 2015-03-03
Do less work to convert constant colors.

manthey:
When we access a style, we always return a function.  Colors are passed through the util.convertColor function.  This was always wrapped util.covertColor(util.ensureFunction(color)).  However, if the color was not a function, it is more efficient to use util.ensureFunction(util.covertColor(color)), as we know the conversion only needs to be performed once.

### DIFF
--- a/src/core/feature.js
+++ b/src/core/feature.js
@@ -307,14 +307,20 @@ geo.feature = function (arg) {
       }
       return all;
     }
-    out = geo.util.ensureFunction(m_style[key]);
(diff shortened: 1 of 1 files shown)


jbeezley:
LTGM.

This is one of the reasons why I started PR #295.  It becomes too much of a burden to store the accessors like this.  My goal is to move to a convention where functional accessors are evaluated immediately and thrown away.

jeffbaumes:
+1 for immediately-evaluated accessors.

PR #342, opened 2015-03-06
Improve npm metadata

jbeezley:


jbeezley:
I just noticed the notice in the README states that the license is Apache 2.  I put BSD 3 clause in the package.json because that is what the content of LICENSE.txt is.  I don't particularly care which we use, but we should probably clarify.

aashish24:
good catch. I would recommend Apache 2 because that's what some of the organizations that we want to work prefer. I will fix the License.txt  and pacakage.json in another PR.

aashish24:
:+1:

PR #343, opened 2015-03-06
Don't create needless arrays.

manthey:
We create arrays using the vgl.DataBuffers.  Except for positions, which need to be transformed, load this data buffers directly.  Prior to this, the data was loaded into javascript arrays, and then that was copied to the data buffers.

Fill buffers much more efficiently by doing it directly instead of calling buffers.write and buffers.repeat

Move code from createGLPoints to _init and from createGLLines to _init.  Don't do things more than once if they can be done just once.

The point vertexShader wasn't explicitly asking for highp floats.

Added an early-exit clause to convertColor to speed up handling colors that are already in the preferred format.

Update the reference to vgl.

I will add speed test for this in geojs in another PR.

This vastly speeds up my tests in the geoapp taxi application:

Test | Loading 1,000,000 Points | Loading 1,000,000 Lines
---|---|---
Before changes | 5.7s average | 27.6s average
After changes | 0.8s average | 4.0s average

### DIFF
--- a/src/gl/lineFeature.js
+++ b/src/gl/lineFeature.js
@@ -189,27 +189,26 @@ geo.gl.lineFeature = function (arg) {
         vert[1].strokeWidth = strkWidthFunc(lineItemData, j, lineItem, i);
         vert[1].strokeColor = strkColorFunc(lineItemData, j, lineItem, i);
         vert[1].strokeOpacity = strkOpacityFunc(lineItemData, j, lineItem, i);
-        if (!j) {
-          continue;
-        }
-        for (k = 0; k < order.length; k += 1, dest += 1, dest3 += 3) {
-          v = vert[order[k][0]];
-          posBuf[dest3]     = position[v.pos];
-          posBuf[dest3 + 1] = position[v.pos + 1];
-          posBuf[dest3 + 2] = position[v.pos + 2];
-          prevBuf[dest3]     = position[v.prev];
-          prevBuf[dest3 + 1] = position[v.prev + 1];
-          prevBuf[dest3 + 2] = position[v.prev + 2];
-          nextBuf[dest3]     = position[v.next];
-          nextBuf[dest3 + 1] = position[v.next + 1];
-          nextBuf[dest3 + 2] = position[v.next + 2];
-          offsetBuf[dest] = order[k][1];
-          indicesBuf[dest] = dest;
-          strokeWidthBuf[dest] = v.strokeWidth;
-          strokeColorBuf[dest3]     = v.strokeColor.r;
-          strokeColorBuf[dest3 + 1] = v.strokeColor.g;
-          strokeColorBuf[dest3 + 2] = v.strokeColor.b;
-          strokeOpacityBuf[dest] = v.strokeOpacity;
+        if (j) {
+          for (k = 0; k < order.length; k += 1, dest += 1, dest3 += 3) {
+            v = vert[order[k][0]];
+            posBuf[dest3]     = position[v.pos];
+            posBuf[dest3 + 1] = position[v.pos + 1];
+            posBuf[dest3 + 2] = position[v.pos + 2];
+            prevBuf[dest3]     = position[v.prev];
+            prevBuf[dest3 + 1] = position[v.prev + 1];
+            prevBuf[dest3 + 2] = position[v.prev + 2];
+            nextBuf[dest3]     = position[v.next];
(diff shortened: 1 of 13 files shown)


aashish24:
Very nice @manthey :+1: . Will review in detail as soon as I get some time but looks like this along the lines we discussed. Thanks.

manthey:
The added speed tests for selenium may need their thresholds adjusted depending on the dashboard computer(s) that run those tests.  I've tried to pick conservative values.

aashish24:
thanks @manthey. I will get the reviews done today.

src/gl/lineFeature.js
```diff
@@ -123,162 +123,107 @@ geo.gl.lineFeature = function (arg) {
...
+        geom = m_mapper.geometryData();
+
+    for (i = 0; i < data.length; i += 1) {
+      lineItem = m_this.line()(data[i], i);
+      numSegments += lineItem.length - 1;
+      for (j = 0; j < lineItem.length; j += 1) {
+        pos = posFunc(lineItem[j], j, lineItem, i);
         if (pos instanceof geo.latlng) {
-          position.push([pos.x(), pos.y(), 0.0]);
+          position.push(pos.x());
```
aashish24: Just curious what was the motivation from moving away from pushing an array vs pushing a single value?

manthey: Two reasons: (1) the transform is faster on a single array than on an array of arrays, and (2) less memory use.  A single array uses less memory than an array of arrays.  Also, with the array or arrays we end up creating one array per point initially, then one array per point in the transform.

src/gl/lineFeature.js
```diff
@@ -123,162 +123,107 @@ geo.gl.lineFeature = function (arg) {
...
-      lineSegments.push(lineItem.length);
-      lineItem.forEach(function (lineItemData) {
-        pos = posFunc(lineItemData, lineItemIndex, item, itemIndex);
+        buffers,
+        posBuf, nextBuf, prevBuf, offsetBuf, indicesBuf,
+        strokeWidthBuf, strokeColorBuf, strokeOpacityBuf,
+        dest, dest3,
+        geom = m_mapper.geometryData();
+
+    for (i = 0; i < data.length; i += 1) {
```
aashish24: Is for loop is preferred vs forEach?

manthey: Because we need to use an index variable (what had been `lineIndex` and `lineItemIndex`), using a for loop is mildly more efficient.  If you don't need the index variable, the difference is very slight (but the for loop is still slightly faster).  It isn't a big time difference, but it is there.

jbeezley: I get that the for loop is faster because it avoids the function calling, but are you saying there is a difference between these?
```javascript
A.forEach(function (d) {});
```
```javascript
A.forEach(function (d, i) {});
```

manthey: My memory was that `for (var i=0; i<A.length; i++) { <action> }` had been faster than `forEach`.

I just tested this in the latest version of Chrome and Firefox.  The for loop is no longer faster (or I'm just misremembering -- though I found some historic data that supports my memory).  Some actual results:

`A.forEach(function (d, i) { <action> } );` is the fastest in both Chrome and Firefox
`var i=0; A.forEach(function (d) { <action>; i += 1; }` is slower (it has roughly 0.4 microseconds overhead per iteration in my version of Chrome and 0.05 microsecords overhead per iteration in Firefox).

The plain `for` loop ends up between the two or slower depending on what is inside the loop.

Therefore, I really should switch back to `A.forEach(function (d, i) {}` for maximum speed.

jbeezley: Interesting result [here](http://jsperf.com/for-vs-foreach/9).  For me, `for` is 75% faster than `forEach` in chrome, but they are roughly the same speed in firefox.

manthey: I just tried out switch the lineFeature to use forEach instead of for.  In my taxi data tests, the results are conclusive.  `for` is faster than `forEach` (averaging three tests for each in an A-B-A-B-A-B manner) by 17% overall.

jbeezley: Nice!  I'll keep that in mind in the future.

aashish24: So in summary: 

forEach is slower and for is faster in general for most platforms we care?

manthey: Yes.  forEach is substantially slower.

PR #347, opened 2015-03-11
Reuse source buffers

manthey:
This is a further extension of the `allocate-less-memory` branch, but the changes introduced in the last commit are independent of that, so I'm adding this as a separate PR.  If this is reviewed directly, both can be satisfied with this PR.

Rather than using the DataBuffers structure, allocate source buffers directly, reusing them if they haven't changed.  Also, for points and lines, we require the vertex indices to exist, but we don't do anything with them, so create but don't fill those arrays.

This shouldn't speed up the first load of data much, but when data is reloaded with the same number of points or lines, it improves the speed.  Compared to allocating the buffers every time, this reduces the time to load 1,000,000 points from 0.8 s to 0.7 s and for 1,000,000 lines from 4.0 s to 3.5 s (roughly 12 % improvement).

### DIFF
--- a/testing/test-cases/selenium-tests/glLinesSpeed/testGlLinesSpeed.py
+++ b/testing/test-cases/selenium-tests/glLinesSpeed/testGlLinesSpeed.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+
+import math
+import os
+import unittest
+
+import selenium_test
+from selenium_test import FirefoxTest, ChromeTest, ThresholdException
+
+
+setUpModule = selenium_test.setUpModule
+tearDownModule = selenium_test.tearDownModule
+
+
+class glLinesSpeedBase(object):
+
+    testCase = ('glLinesSpeed',)
+    testRevision = 1
+
+    def loadPage(self):
+        self.resizeWindow(640, 480)
+        self.loadURL('glLinesSpeed/index.html')
+        self.wait()
+
+    def testGlLinesSpeed(self):
+        self.loadPage()
+
+        testName = 'drawGlLinesSpeed'
+        self.loadTimeTest(testName)
+        self.framerateTest(testName)
+
+    def loadTimeTest(self, testName, revision=None):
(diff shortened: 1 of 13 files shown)


testing/test-cases/selenium-tests/glLinesSpeed/testGlLinesSpeed.py
```diff
@@ -0,0 +1,64 @@
...
+        self.loadTimeTest(testName)
+        self.framerateTest(testName)
+
+    def loadTimeTest(self, testName, revision=None):
+        # Threshold is in milliseconds.  We need the value to be less than the
+        # threshold.
+        el = self.getElement('#loadResults')
+        value = float(el.get_attribute('results'))
+        threshold = float(os.environ.get('LOAD_SPEED_THRESHOLD', '1000'))
+        print 'Average load time %1.0f ms (must be at least %1.0f ms)' % (
```
jbeezley: s/at least/less than/

testing/test-cases/selenium-tests/glLinesSpeed/testGlLinesSpeed.py
```diff
@@ -0,0 +1,64 @@
...
+        if value > threshold or math.isnan(value):
+            raise ThresholdException({'value': value, 'threshold': threshold})
+
+    def framerateTest(self, testName, revision=None):
+        # Threshold is in frames-per-second.  We need the value to be greater
+        # than the threshold.
+        el = self.getElement('#framerateResults')
+        value = float(el.get_attribute('results'))  # in milliseconds
+        threshold = float(os.environ.get('FRAMERATE_THRESHOLD', '10'))
+        print 'Average framerate %4.2f fps (must be less than %4.2f fps)' % (
```
jbeezley: s/less than/at least/

jbeezley:
The test look great.  I'll leave the review of the GL to @aashish24.

aashish24:
Other than that one suggestion, looks great to me! I am merging it :+1:

PR #351, opened 2015-03-17
Polygon colors and opacities could be wrong.

manthey:
If you have multiple polygons, their colors and opacities would be wrong, because the would reference the wrong array index.

Also, made it so that 'inner' data is not required.


aashish24:
Looks great! Thanks.

PR #353, opened 2015-03-18
Lines were not taking into account aspect ratio.

manthey:
Angle at the ends of lines weren't quite right.  If a map was not square, the line itself was not a consistent width, but depending on its orientation.

Switched the arctangents to calculate only when needed, and then use the atan(y, x) form rather than the convolute if-else statements and atan(y / x).


jeffbaumes:
Go Math! :heavy_plus_sign: :heavy_multiplication_x:

aashish24:
Looks great to me :heavy_division_sign:  :smile: :

PR #354, opened 2015-03-19
Updated point sprite feature

aashish24:


aashish24:
We should get better performance compare to squares and triangles. @jbeezley I may need your help to upload a new baselines. Tests always fail on machine. @manthey would be nice to capture some performance numbers for taxi  using sprites. 
* I will add the image option later once I have more time.

jbeezley:
I suspect they fail because the image sizes are different.  The problem is that you give the window size in selenium, but the screenshot is of the viewport only.  The difference is in the decorations provided by the window manager, so the baseline images are really only valid per computer.

When my new computer is in, I will set up a selenium server that we can all use to run the tests.  Until then, you just need to have me generate the baselines.

manthey:
One issue and some suggestions:

The `verticesPerFeature` needs to return 1 for sprites:

    this.verticesPerFeature = function () {
        if (m_primitiveShape === "sprites") {
          return 1;
        }
        var unit = pointPolygon(0, 0, 1, 1);
        return unit.length / 2;
      };

We are calling the `m_primitiveShape` either `triangle`, `square`, or `sprites`.  It would be better if they were all singular or plural.

We should change the default to sprites.

There are a number of minor style issues (commas at the end of lists, for instance).

For the taxi data set, the sprites are faster.  Load time is 0.52 seconds for 1,00,000 points.  For triangles load time is 0.75 seconds for 1,000,000 points.  The frame rate is not substantially different.  With triangles, I can load 5,000,000 points.  With sprites, I can load 15,000,000 points (perhaps 17,000,000).

If you want, I can push a change to add the verticesPerFeature fix and the style fixes.

aashish24:
Okay, sounds good. Thanks,

aashish24:
* yes, I will fix the this.verticesPerFeature...
* I haven't fixed the style (saw them earlier on the dashboard), will push a fix. Wasn't ready for merge.
* The numbers look good! thanks. I think the throughput is impressive. 
* I will push a fix. I have to make sure my commit count it high :smile:

aashish24:
sorry, missed two things:

* I am okay with triangles or making sprites to sprite, it just that sprites is more common term even when it is referred in singlular context. 
* default to sprite sounds good. I will make the change.

aashish24:
@manthey verticesPerFeature does return 1. If you look at the code, 
case "sprite":
        /* Point sprite uses only one vertex per point. */
        verts = [x, y];
        break;

and then we do:

var unit = pointPolygon(0, 0, 1, 1);
    return unit.length / 2;

Did you mean that I change this to just return 1?

manthey:
Somehow I thought verticesPerFeature wasn't returning 1 for sprite.  My error -- don't change it.

manthey:
You removed a small code snippet from the fragment shader:
    if (rad > 1.0)
      discard;
This helps speed things up and should be put back in.

aashish24:
Ah good catch. That got lost when I split up the shader code. I will put it back in next few minutes.

aashish24:
@jbeezley if this goes to master, then I think we all three agreed that we can cut 0.2 release. If anyone has any objections to it, then wen can discuss it over the email. 

I will push a branch for licensing and readme fix.

jbeezley:
:checkered_flag: :clap:

aashish24:
this is ready for merge..let me know if I missed something.

manthey:
LGTM.

aashish24:
thanks @manthey

PR #357, opened 2015-03-25
Add a OSM layer and planeFeature opacity options.

manthey:
Add a mapOpacity option to the OSM layer.  Add an opacity style for planeFeatures.

I added a second test case for planes, which will need a new reference image.


aashish24:
LGTM :+1: . What I am hoping down the road that plane feature will draw multiple planes not just one and could possibly use polygonFeature internally.

PR #358, opened 2015-04-01
Changes to the selenium test runners

jbeezley:
This is a small change to the selenium testing that does the following:
  1. Requires starting the test server manually, but allows running the tests in parallel
  2. Fixes running tests on a remote selenium node.
  3. Adds an example in the documentation of running on the selenium node I'm hosting for geojs testing within Kitware

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ script:
     - git submodule update
     - npm install
     - grunt default docs
+    - grunt serve-test &
     - mkdir _build
     - cd _build
     - cmake ..
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js:
 
 before_install:
   - pushd "${HOME}"
-  - curl "http://www.cmake.org/files/v3.2/cmake-3.2.0-rc2-Linux-x86_64.tar.gz" | gunzip -c | tar x
+  - curl "http://www.cmake.org/files/v3.2/cmake-3.2.1-Linux-x86_64.tar.gz" | gunzip -c | tar x
   - cd cmake-*/bin && export PATH="${PWD}:${PATH}"
   - popd
   - cmake --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ script:
     - npm install
     - grunt default docs
     - grunt serve-test &
+    - sleep 1
     - mkdir _build
     - cd _build
     - cmake ..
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ node_js:
 
 before_install:
   - pushd "${HOME}"
-  - curl "http://www.cmake.org/files/v3.2/cmake-3.2.0-rc2-Linux-x86_64.tar.gz" | gunzip -c | tar x
+  - curl "http://www.cmake.org/files/v3.2/cmake-3.2.1-Linux-x86_64.tar.gz" | gunzip -c | tar x
```
aashish24: +1

CMakeLists.txt
```diff
@@ -242,10 +239,10 @@ if(SELENIUM_TESTS)
   endforeach()
 
   set_property(TEST "selenium:glPointsSpeed" APPEND PROPERTY ENVIRONMENT "LOAD_SPEED_THRESHOLD=1000")
-  set_property(TEST "selenium:glPointsSpeed" APPEND PROPERTY ENVIRONMENT "FRAMERATE_THRESHOLD=10")
+  set_property(TEST "selenium:glPointsSpeed" APPEND PROPERTY ENVIRONMENT "FRAMERATE_THRESHOLD=5")
```
aashish24: why reduce threshold?

Gruntfile.js
```diff
@@ -431,9 +431,30 @@ module.exports = function (grunt) {
...
-      'use the --port option to override the default port',
-      ['express', 'watch']
+    'serve',
+    'Serve the content at http://localhost:8082, ' +
+    'use the --port option to override the default port',
+    ['express', 'watch']
+  );
+
+  grunt.registerTask(
+    'serve-test',
```
aashish24: Nice addition

aashish24:
Also, how does it fixes running tests on a remote selenium node?

CMakeLists.txt
jbeezley: To make it more likely the test would pass on my dashboard machine when it is under load.  I don't think this test will ever be able to catch minor regressions in speed.

jbeezley:
It was possible before, but not as straightforward.  The test host [now](https://github.com/OpenGeoscience/geojs/pull/358/files#diff-af3b638bc2a3e6c650974192a53c7291R27) defaults to the hostname rather than localhost and is no longer marked as advanced.

aashish24:
I see. thanks. LGTM

PR #359, opened 2015-04-02
Add logo

aashish24:


aashish24:
@jbeezley its ready for review

jbeezley:
:+1:

ronichoudhury:
Why is the typeface in this logo different from the one [here](http://www.tangelohub.org/)?  I think for branding purposes we should consider the uniform logo-typeface combo for the TangeloHub line of platforms.

jbeezley:
I wouldn't mind another pass over the logo style.  To me, the color are a bit light.  I think I like [this version](http://opengeoscience.github.io/geojs/images/logo.pdf) better.  Also, I attempted to generate some [favicons](https://github.com/OpenGeoscience/geojs/tree/gh-pages/favicons) from the symbol.  They look okay, but it would be better to generate some with transparent backgrounds.

aashish24:
I was planning to talk to steve about another pass on the logo but didn't want to wait as he is busy at the momement. I will ask him to generate some options.

jbeezley:
@ronichoudhury and I just had a conversation about updating the logo to be more consistent with the other tangelohub logos.  In particular, we should use the same font and limit it to a single base color with multiple shades.  When Steve has time, we should get together and go over what we want.

aashish24:
yup that was the idea Steve and I had. The logo on the tangelohub is close but not perfect. I will send few options to the mailing list for the voting

PR #363, opened 2015-04-04
Expose a method to set map extent

jbeezley:
It is only an approximation for now.  To compute the extents exactly, we will need to do some refactoring of the base layer/base renderer api.  Two new methods supported:

```javascript
// go directly to the new extent
map.bounds({
  lowerLeft: {x: 10, y: 10},
  upperRight: {x: 20, y: 20}
});
```
or
```javascript
// compute the new center and zoom for given bounds
var nav = map.zoomAndCenterFromBounds({
  lowerLeft: {x: 10, y: 10},
  upperRight: {x: 20, y: 20}
});

// this allows animated transitions
nav.interp = d3.interpolateZoom
map.transition(nav);
```

You can try it [here](http://jsbin.com/zoticu/3/edit).

### DIFF
--- a/src/core/map.js
+++ b/src/core/map.js
@@ -859,23 +859,71 @@ geo.map = function (arg) {
 
   ////////////////////////////////////////////////////////////////////////////
   /**
-   * Get the locations of the current map corners as latitudes/longitudes.
-   * The return value of this function is an object as follows: ::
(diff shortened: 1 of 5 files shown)


aashish24:
thanks. LGTM :+1:

PR #369, opened 2015-04-07
Use pre-blended alpha.

manthey:
For post-blending alpha, we need to use the blendFuncSeparate call.  However,
some computers (notably MacBook Pros) don't accelerate blendFuncSeparate, so it
is faster on those machines to use pre-blended alpha.

This requires the companion vgl change.

### DIFF
--- a/src/gl/lineFeature.js
+++ b/src/gl/lineFeature.js
@@ -110,7 +110,7 @@ geo.gl.lineFeature = function (arg) {
       'varying float strokeWidthVar;',
       'varying float strokeOpacityVar;',
       'void main () {',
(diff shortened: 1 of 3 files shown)


manthey:
@jbeezley When I ran the tests against your selenium server, most of them passed, but the jquery Plugin test failed.  I don't see why these changes would have affected that, but if they did, I'll probably need some help tracking down the cause.

jbeezley:
Don't worry about it.  The jquery plugin failed in yesterday's nightly run, too.  I haven't gotten a chance to look into yet.

aashish24:
I am closing this pull request unless we verify if this really solving the issue @manthey is seeing on this particular system.

PR #371, opened 2015-04-10
Drop initialize check

jeffbaumes:
This does not seem to be needed anymore and in fact causes an error on the `delete window.gl` line for me.

### DIFF
--- a/src/plugin/jquery-plugin.js
+++ b/src/plugin/jquery-plugin.js
@@ -288,16 +288,6 @@
         // when called multiple times on a single element, do nothing
         return;
       }
-      if (initialized) {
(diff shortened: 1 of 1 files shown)


aashish24:
LGTM but before we merge it it would be nice if @jbeezley can have a look at it.

jbeezley:
It's baffling to me why `delete window.gl` would cause an error, but I'm will to accept the change.

jeffbaumes:
FYI for some reason it said that window.gl was undefined. Perhaps I was double-initializing before the gl field was created again?

jbeezley:
Maybe, but even if `gl` was undefined, it shouldn't cause an error.

PR #384, opened 2015-04-24
Only calculate the range tree as needed.

manthey:
The range tree was being prepared ahead of time coupled to the style call.  This would often (always?) calculate the range tree multiple times even when the data had not changed.  Since calculating the range tree is expensive (multiple seconds for a few hundred thousand points), this is undesirable.

Now, the range tree has a date stamp, and is only calculated when needed and stale.

Also, refactored the RangeNode so that it does lazy memory allocation and sorting.  This speeds up the initial construction substantially, plus uses less memory in all but a few degenerate cases.

For reference, I was seeing _updateRangeTree called 2 or 3 times when constructing an initial point set.  For 250,000 points this took roughly 5 seconds per call.  Now, it is only called once (and possibly not even then, if the mouse never tracks across the map), and the initial call takes 300 ms.  Of this 300 ms, roughly 60% of the time is in sorting the points along the x coordinate and 40% is in constructing the RangeNodes.  It now feels responsive on a 250,000 point set.

Lastly, I noticed the vgl reference was for the pull request.  I've updated it to the current master sha (there is no change, but now you can browse the master branch without it appearing as a difference to git).

### DIFF
--- a/src/core/pointFeature.js
+++ b/src/core/pointFeature.js
@@ -23,8 +23,8 @@ geo.pointFeature = function (arg) {
   var m_this = this,
       s_init = this._init,
       m_rangeTree = null,
+      m_rangeTreeTime = geo.timestamp(),
       s_data = this.data,
-      s_style = this.style,
       m_maxRadius = 0;
 
   ////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


jbeezley:
Looks good.  :+1:

jeffbaumes:
Way to go @manthey!

Issue #365, opened 2015-04-06
size option in geojsMap should accept constants as well as functions


ronichoudhury:

aashish24:
:+1:

PR #389, opened 2015-04-28
Allow constants for size argument in jquery plugin

jbeezley:
Also fixes errors in the docstrings.  Addresses #365.


aashish24:
LGTM :+1:

PR #391, opened 2015-05-08
Small doc fix - trying to escape brackets too

jeffbaumes:


jbeezley:
:+1:

src/plugin/jquery-plugin.js
```diff
@@ -256,9 +256,9 @@
      *   Describes layers added to the map
      * @property {boolean} [autoresize=true]
      *   Resize the map on <code>window.resize</code> (initialization only)
-     * @property {string} [tileServer]
+     * @property {string} [tileUrl]
      *   The open street map tile server spec default:
-     *   <code>http://tile.openstreetmap.org/<zoom>/<x>/<y>.png</code>
+     *   <code>http://tile.openstreetmap.org/&lt;zoom>/&lt;x>/&lt;y>.png</code>
```
aashish24: <code>http://tile.openstreetmap.org/&lt;zoom>/&lt;x>/&lt;y>.png</code>

shouldn't it be:
 <code>http://tile.openstreetmap.org/&lt;zoom&gt;/&lt;x>/&lt;y&gt;.png</code>

but it may be that I missing something.

jbeezley: I'm not sure about exact rules for html escape, but this seems to work.

screen shot 2015-05-08 at 11 40 58 am

aashish24:
Interesting.  I think I know why. It has to something to do with parser trying to match < and > together. If you replace < with &lt; then that breaks and then it treats the other > as any other character.

PR #392, opened 2015-05-08
Improve point default color

aashish24:


aashish24:
@jbeezley as discussed, I will let you fix the bugs.

src/core/pointFeature.js
```diff
@@ -242,14 +242,14 @@ geo.pointFeature = function (arg) {
     var defaultStyle = $.extend(
       {},
       {
-        radius: 10.0,
+        radius: 5.0,
         stroke: true,
-        strokeColor: { r: 0.0, g: 1.0, b: 0.0 },
+        strokeColor: { r: 0.851, g: 0.604, b: 0.0 },
         strokeWidth: 2.0,
```
jeffbaumes: Can we change this to 1.0?

aashish24: I pixel might be too small but I will see how it looks and if it looks reasonable then I am okay with it.

jbeezley:
I fixed the tests.  Here is some comparisons of stroke widths `1`, `1.25`, `1.5`, and `2` with both renderers.  I think `2` is too wide especially for this size of point.  In the past, I have noticed bad antialiasing for `1` pixel wide circles in svg.  I checked various browsers and the antialiasing doesn't look too bad anymore, but I think `1.25` might look a bit better.

http://output.jsbin.com/tikota/13

aashish24:
thansk @jbeezley. If I zoom way in, then you will see that 1.25 and 1.5 both look good. I think  1.5 looks slightly better though but folks may prefer 1.25.

aashish24:
Okay, lets go with 1.25 then.

jbeezley:
Fixed and ready for merge.

aashish24:
:+1:

PR #397, opened 2015-06-02
WMS layers should be in the same projection as the map/baselayer

mbertrand:
…this case, EPSG:3857.


aashish24:
Great! I will test it locally and have it merged. Thanks @mbertrand

aashish24:
Works perfectly now! thanks @mbertrand. Merging it.

PR #398, opened 2015-06-03
For osmLayer, make tile.crossOrigin configurable 

mbertrand:
For osmLayer, this patch makes tile.crossOrigin configurable (default "anonymous, or "use-credentials") by passing a boolean  'useCredentials' argument.  This is necessary for loading map tiles from WMS servers that require authentication.

Example:
     var wms = map.createLayer('osm', {useCredentials: true});

### DIFF
--- a/src/core/osmLayer.js
+++ b/src/core/osmLayer.js
@@ -47,7 +47,8 @@ geo.osmLayer = function (arg) {
     m_updateDefer = null,
     m_zoom = null,
     m_tileUrl,
-    m_tileUrlFromTemplate;
+    m_tileUrlFromTemplate,
+    m_crossOrigin = "anonymous";
 
   if (arg && arg.baseUrl !== undefined) {
     m_baseUrl = arg.baseUrl;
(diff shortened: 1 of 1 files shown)


aashish24:
thanks @mbertrand. How do you pass those credentials?

mbertrand:
Credentials are read from cookies when tile.crossOrigin = 'use-credentials'; the cookies are not passed in the request header if tile.crossOrigin='anonymous'.  

For example, if you go to http://geodata.epidemico.com/geojs, you will see an OSM basemap but not a WMS country boundaries layer it tries to load (because it requires authentication).  Go to http://geodata.epidemico.com and log in with temporary account username 'geojs_user', password 'geojs' - then revisit the geojs page and you should see the WMS country boundaries layer appear.   Without this patch, the WMS layer will not appear even after logging in.

aashish24:
Got it. That what's I thought but just wanted to confirm. I didn't know that cookies will be be passed by like that. Thanks for the explanation. LGTM :+1:

PR #407, opened 2015-06-30
Remove geo.latlng

jbeezley:
This class was never used consistently and most transform methods using it were broken.  Enforcing getters and setters will also kill performance unnecessarily.  If functional getter/setter is necessary, it is far better to use ES5 properties on a case-by-case basis.

This is partially in preparation for a refactor of the transform and osm layer API that I am working on.  Also, -274 LOC :scissors:.


aashish24:
@jbeezley would it mean that coordinates are referred by the projection string if set on the layer?

jbeezley:
I'm not certain what your asking.  All I did here was remove the latlng class, which was largely not being used anyway.

aashish24:
I meant to say that if user provided geo.latlng instance we knew the projection but without it, we need to know the projection the numbers are coming in (if it is required). I will look at the changes may that information is not needed.

jbeezley:
There is a need for handling transformations between coordinates systems in general, yes, but latlng's were never actually used that way.  The entire projection/coordinate transformation codebase needs to be rewritten anyway.  I'm not terribly concerned about these issues until that happens.

aashish24:
Okay sounds good. It would be great if we can throw an idea to me on transformation before you actually do it. In the mean time, I will review this branch.

aashish24:
Overall the branch looks good to me. My only concern is the folks that are using google maps or other API, will expect object like latlng as that makes it easy to pass data to geojs and folks don't have to think about lat is Y and lng is X. On the other hand, yes, this change makes our code more general with fewer lines. I am wondering what others think about this.

jbeezley:
I would be okay with having a specific class to provide coordinates if we enforced it uniformly across the whole library.  The problem with magic detection of arguments like we currently have is that you end up with if blocks like [this](https://github.com/OpenGeoscience/geojs/blob/master/src/core/osmLayer.js#L200) (and [this](https://github.com/OpenGeoscience/geojs/blob/master/src/core/transform.js#L161) and [this](https://github.com/OpenGeoscience/geojs/blob/master/src/gl/vglRenderer.js#L69), etc.) all over the code.

If we did have a `latlng` class, I would want to get rid of the function call `latlng.protoype.lat()` and use a bare attribute because we have found that function calls can introduce a lot of overhead.

aashish24:
+1 for latlng class with bare attribute. But even with that, I guess we will have to detect the argument type. Or may be you suggested something and I missed that?

jbeezley:
For the purposes here, I'm advocating for reducing the complexity of the argument checking.  Enforcing a single latlng class would mean at most starting these methods with
```javascript
if (not arg typeof geo.latlng) {
    throw new Error('Wrong argument type')
}
```
But the would be a big change to the API, so we would need to consider it more carefully.

aashish24:
Okay, let;s go ahead and merge this branch :+1: We can collect some feedback from other folks and see if they miss a latlng object. 

LGTM

ronichoudhury:
+1 for having a single latlng class and simplifying those if-elseif-then
blocks.  Those make me nervous because they multiply the chance of an error
appearing.  And I've said before, I wasn't sure of the utility of allowing
users to say either "x/y" or "lat/long" at will, when the cost is a more
complex API and a more bug-prone codebase.  As for bridging to other
mapping suites, a collection of adapter classes, one per mapping library,
could easily take care of that, right?

jbeezley:
@ronichoudhury Yes, that is where I would like to head in the future.  We need to work out some details before we go about implementing it though.  For example, how to we encode the coordinate system?  Is it a property of the specific point class or do we just assume the user provided the values in the correct coordinates?  All of this needs to take into account the memory and computational overhead that it requires because we want to be able to support arrays of millions of these objects.

aashish24:
@jbeezley I am merging this since I think we covered discussion pretty well here.

PR #408, opened 2015-07-01
Remove outdated and broken examples

jbeezley:


aashish24:
@jbeezley are we going to replace them with updated / newer examples?

jbeezley:
For the most part, these have been superseded by new examples or tests already.  The only exception is the feature demo, which was written using the old feature api, so would have to be totally rewritten anyway.  I could redo this one, but I don't have time in the near future to do it.

aashish24:
Sounds good. I created an issue for that (https://github.com/OpenGeoscience/geojs/issues/409). Other than that :+1: LGTM

aashish24:
I am wondering why garant is not running?

jbeezley:
http://my.cdash.org/index.php?project=geojs

looks like it is to me.

On Wed, Jul 1, 2015 at 10:01 AM, Aashish Chaudhary <notifications@github.com

aashish24:
Now i see it, thanks.

aashish24:
:+1:

PR #410, opened 2015-07-01
[WIP] Refactoring transform API and OSM layer

jbeezley:
This is a start of an overhaul of the OSM layer to make it more generic to handle nonstandard projections as well as non-geospatial applications.  I will update later with a detailed list changes I'm planning to make.


aashish24:
@jbeezley 

I prefer we do not delete mercator and eliipsoid as they contain useful functionality and devs will expect them. Can you please describe the plan?

jbeezley:
I will describe the plan in detail before going on, but in regard to removing code:

1.  I didn't delete `ellipsoid`, I only removed it from the build.  I agree it could be useful at some point, but at the moment it is untested, unused, not integrated with anything, and bloating the library.  If you want to add a test, I will be happy to put it back in the build.
2.  `mercator`, I deleted because the mercator projection is provided by proj4, and there is no need for a separate implementation.  The mercator class also provides some of the OSM hierarchical tiling methods, but my plan in general is to abstract that away to apply to an arbitrary rectilinear coordinate system.

Going forward, I'm trying to improve the quality of this library.  Part of that mean reducing the cruft.  The minified library is **way** too big by web standards, and the testing coverage is pretty bad (<50%).  We need to stop including untested (usually broken) code.

aashish24:
@jbeezley agree that we need more testing and coverage but I would suggest that we add test for that as like you suggested. the mercator code is not same as the the one provided by proj4. It is more specific to OSM for most part and does not use the 3857 scale (minor and major axis). I thought we are using it for osmLayer but I could be wrong here. As far as the ellipsoid goes, I will add a test (or you can) but that would sometime next week.

jbeezley:
Actually, now that I look more closely at the ellipsoid code, we need to delete it as well.  All it is doing is datum calculations which proj4 supports natively.  There is absolutely no need for it.

Your assertion about proj4's version of mercator is not correct.  Here is the definition they use:
```javascript
>>> proj4.defs['EPSG:3857']
{ title: 'WGS 84 / Pseudo-Mercator',
  projName: 'merc',
  a: '6378137',
  b: '6378137',
  lat_ts: 0,
  long0: 0,
  x0: 0,
  y0: 0,
  k0: 1,
  units: 'm',
  datumCode: 'none',
  no_defs: true }
```
That is a spherical earth (radius `6378137`)... i.e. pseudo-mercator, the same datum used by Google earth, OSM, and every other standard tile set.

aashish24:
Also, code like below will be useful. Plus, once we add support for 3D globe, we will need definition of ellipsoid. We can take out the transform functions from it and that would be fine (assuming that we can get the non-scaled version from PROJ)

this.computeGeodeticSurfaceNormal = function (lat, lon) {
    if (typeof lat === 'undefined' || typeof lon === 'undefined') {
      throw '[error] Valid latitude and longitude is required';
    }

    var cosLatitude = Math.cos(lat),
        result = vec3.create();

    result[0] = cosLatitude * Math.cos(lon);
    result[1] = cosLatitude * Math.sin(lon);
    result[2] = Math.sin(lat);

    vec3.normalize(result, result);
    return result;
  };

jbeezley:
> Plus, once we add support for 3D globe, we will need definition of ellipsoid.

That's exactly what a datum is.

> We can take out the transform functions from it and that would be fine (assuming that we can get the  non-scaled version from PROJ)

The units of the projection is an orthogonal issue.  The transform class is specifically for converting between curvilinear coordinate systems.  The affine transformations related to scale, translation, and rotation only make sense relative to the viewport being displayed, which a function of the renderer.

But, since you mention it, we also should begin using meters for `EPSG:3857` as well.  Any normal dataset given in that projection will be provided in meters because that's what every other mapping library uses.  See [openlayers](https://github.com/openlayers/ol3/blob/master/src/ol/proj/epsg3857projection.js#L109) and [leaflet](https://github.com/Leaflet/Leaflet/blob/45034155bdf8d958554b43ed8dbd7020df9ca6ab/src/geo/projection/Projection.SphericalMercator.js#L9) implementations for example.

> ```javascript
> this.computeGeodeticSurfaceNormal = function (lat, lon) {
>
> var cosLatitude = Math.cos(lat),
>    result = vec3.create();
>
> result[0] = cosLatitude * Math.cos(lon);
> result[1] = cosLatitude * Math.sin(lon);
> result[2] = Math.sin(lat);
>
> vec3.normalize(result, result);
> return result;
> };
> ```

That's not even the correct formula for the normal from a geodetic angle.  It is a perfect example of why we should start removing dead code from the library.  In any case, this method is [provided](https://github.com/proj4js/proj4js/blob/master/lib/datum.js#L95) by proj4.  Note that proj4's implementation takes care to avoid numerical instabilities in the calculation as well.  It is not an easy problem, this is exactly why we should be relying on other libraries to do it.

aashish24:
@jbeezley datum is not same as ellipsoid. Datum uses a ellipsoid (or starts with it) but they are not two names for one thing; https://en.wikipedia.org/wiki/Geodetic_datum

If the method is provided by proj4 great! that's why we added that in the first place. The ellipsoid this class defines is wgs84 (I could be wrong though, it has been some time). What I am trying to emphasize is on the concepts. May we can rename it to wgs84 ellipsoid unless it can be done via PROJ4. I am not objecting to removing code but at the same time I don't want us to loose some concepts that geospatial community will expect to be there in the library.

jbeezley:
Actually my understanding of that article is that a datum is a more general concept than an ellipsoid.  The datums I have used in the past (spherical, NAD83, and WGS84) are all ellipsoidal, which is why I assumed that they all were.  But that just means that this code is even less useful than I thought.

The geospatial community will expect proj4 because that is the standard.  Let's not reinvent the wheel here.

aashish24:
Datum is a reference. Multiple datums can use the same ellipsoid but they can use a different reference point as per their need. WGS84 is datum and ellipsoid both and that's why it could lead someone that datum and ellipsoid are the same thing but they are not. If proj4 has ellipsoid definitions  or objects then yes we don't need our own ellipsoid but if it does not then we will need it.

aashish24:
@jbeezley on this related note: Have a look at this blog: http://ahocevar.net/2014/07/10/proj4js-2-2-x-with-ol3.html (from OL developer).

aashish24:
Also,I am not concerned about ellipsoid as much for now. I would like to see the new tile layer and annotations if possible. If possible, it would be great if we can keep the two behaviors and default to the new behavior that we will be building.

jbeezley:
Yes, this thread has become somewhat pedantic and is largely irrelevant to my justification for removing the files.  I'll provide a plan for the new tile layer interface once I work out in my head.  It is unlikely to be done before I head out for vacation next week.

aashish24:
Sounds great @jbeezley

jbeezley:
I was originally trying to limit this PR to just a refactor of the transform API, but the osm layer is just too heavily coupled with the mercator code to make it an independent change.  The transform API is more or less done.  I think we might want to add an accelerated method that works on an `ArrayBuffer`-like object to speed up transforms inside gl, but I would want that to be a new method to keep things simple.

From here, I need to fix the osm layer and probably generalize the concept of coordinates in the gl renderer.  I will work with @manthey on the renderer.  For the osm layer, I want to use [openseadragon](https://github.com/openseadragon/openseadragon) as a template for how to generalize the image pyramid code.  In particular, I want to have a few new classes:

1. `tile` to describe an abstract interface to draw tiles on the map
  * `imageTile` for tiles loaded from `Image` objects
  * `vectorTile` for tiles rendered locally from vector data
2. `tileCache` for caching the tiles in memory
3. `tileLayer` as a generalization of `osmLayer` (like [`TileSource`](https://github.com/openseadragon/openseadragon/blob/master/src/tilesource.js) in openseadragon)

aashish24:
Thanks @jbeezley. Please arrange a meeting with me and @manthey for coordinates in GL sometime next week. As far as the new classes are concerned, 1., 2. and 3. sounds reasonable. 
https://guides.github.com/features/mastering-markdown/

jbeezley:
Commenting to remind myself to add attribution text to the open street map layer.

jbeezley:
I am at a point on this branch that I need to define the coordinate transformation/camera API.  I have a rough roadmap of where I want to go with it.  These are the core classes involved and the coordinates systems they work with:

* `feature`: The coordinate system of the source data.  For the examples we have so far, this would always be `EPSG:4326`.  Vector-like features (not raster) can be transformed in place using PROJ4.  Raster features (like OSM tiles) for the moment will not be transformable.
* `layer`: All layers on a map would use a common coordinate system (`EPSG:3857` for osm type maps).  Layers are also allowed to define an "internal" coordinate system relative to the common coordinate system that is exposed allowing advanced users to draw directly onto the layers' contexts.  So that differently layers align properly, we need to ensure that these functions are well conditioned and preferably affine.
* `map`: Enforces the common coordinate system shared by the layers.  To handle the issues with numerical precision at high zoom levels, the map will also define an origin that will be "near" the map center in some sense.  The coordinates that the layers use internally then will be offset from standard definitions of coordinates in GIS.  This translation will occur transparently in much the same way as `toLocal/fromLocal`.  In coordinating this, there will need to be an `updateOrigin` event that will be triggered once the map center deviates too far from the origin relative to the zoom level.  As a first approximation this check can occur solely during a zoom in event.
* `camera`: The camera projects the 3D map coordinates (relative to the map origin) into 2D pixel coordinates.  The projection matrix managed by the camera will be provided to the renderers to do the usual rendering calculations.

Here is a quick picture of the coordinate systems and classes involved in terms of the OSM-like applications we have used so far.
geojs-cs
Note that this effectively eliminates the need to have a special `baseLayer`/`baseRenderer`, which means headless testing will be much easier.

manthey:
I like where this is going.

I know there has been some desire to have a world map wrap when zoomed out
(or just when zoomed in at 180 degrees of longitude).  I just mention it so
we can hopefully support it in the future.

I still would love to see us able to render a Goode homolosine projection,
not because it is useful, but because it would be cool (but it violates the
affine requirements).

On Fri, Jul 31, 2015 at 2:57 PM, Jonathan Beezley <notifications@github.com>
wrote:

> I am at a point on this branch that I need to define the coordinate
> transformation/camera API. I have a rough roadmap of where I want to go
> with it. These are the core classes involved and the coordinates systems
> they work with:
>
>    - feature: The coordinate system of the source data. For the examples
>    we have so far, this would always be EPSG:4326. Vector-like features
>    (not raster) can be transformed in place using PROJ4. Raster features (like
>    OSM tiles) for the moment will not be transformable.
>    - layer: All layers on a map would use a common coordinate system (
>    EPSG:3857 for osm type maps). Layers are also allowed to define an
>    "internal" coordinate system relative to the common coordinate system that
>    is exposed allowing advanced users to draw directly onto the layers'
>    contexts. So that differently layers align properly, we need to ensure that
>    these functions are well conditioned and preferably affine.
>    - map: Enforces the common coordinate system shared by the layers. To
>    handle the issues with numerical precision at high zoom levels, the map
>    will also define an origin that will be "near" the map center in some
>    sense. The coordinates that the layers use internally then will be offset
>    from standard definitions of coordinates in GIS. This translation will
>    occur transparently in much the same way as toLocal/fromLocal. In
>    coordinating this, there will need to be an updateOrigin event that
>    will be triggered once the map center deviates too far from the origin
>    relative to the zoom level. As a first approximation this check can occur
>    solely during a zoom in event.
>    - camera: The camera projects the 3D map coordinates (relative to the
>    map origin) into 2D pixel coordinates. The projection matrix managed by the
>    camera will be provided to the renderers to do the usual rendering
>    calculations.
>
> Here is a quick picture of the coordinate systems and classes involved in
> terms of the OSM-like applications we have used so far.
> [image: geojs-cs]
> <https://cloud.githubusercontent.com/assets/31890/9014228/0d0dbcf8-378f-11e5-9f3c-684e25715acb.png>
> Note that this effectively eliminates the need to have a special baseLayer
> /baseRenderer, which means headless testing will be much easier.
>
> —
> <https://github.com/OpenGeoscience/geojs/pull/410#issuecomment-126784448>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

jbeezley:
There aren't strictly any affine "requirements".  A rule of thumb for internal layer coordinate systems.  The map itself could be any CS in PROJ4, we just wouldn't be able to draw raster features on it unless we define a nongridded interpolation method.  One that works for non-surjective coordinate transforms for the general case.

jeffbaumes:
But poor Greenland and Antarctica @manthey! @jbeezley seems like a good approach to have map object manage the projection, rather than a base layer.

aashish24:
Thanks @jbeezley I will provide my feedback on this soon. In a quick look, I am liking where its going.

aashish24:
@jbeezley few pointers related to this branch: 

1) 
feature: The coordinate system of the source data. For the examples we have so far, this would always be EPSG:4326. Vector-like features (not raster) can be transformed in place using PROJ4. Raster features (like OSM tiles) for the moment will not be transformable.

> I would not make an assumption that feature projection always be EPSG 4326. Its just that we don't have many use cases for other projection but we had one or two in the past. Features should be responsible for GCS conversion (how they do it and not necessarily need to have the methods or API). 

>  Camera / API should transform and projects in the target projection system. I believe that's the case but just wanted to make sure. 

> I am assuming that we will allow the Camera class to be over-ridden by different renderers.

jbeezley:
>I would not make an assumption that feature projection always be EPSG 4326. Its just that we don't have many use cases for other projection but we had one or two in the past. Features should be responsible for GCS conversion (how they do it and not necessarily need to have the methods or API).

I don't assume the source data is 4326, but it is the default.  What the gcs of the feature means at an API level is that all coordinates passed in as data will be relative to that coordinate system.  When the coordinate system of the feature is different than the map's, there is an API available through `geo.transform` to transform the coordinates, which can also be extended.

My plan is to have a method in `geo.feature` that checks if the gcs is compatible with the map when created.  The default would check for strict equality because there is no generic way of deciding what is a coordinate.  There are also issues of co/contra-variant vectors that are unknown.  For something simple like a point feature, it is pretty easy to allow an arbitrary transformation that passes the point coordinates through `geo.transform` in the base class.

> Camera / API should transform and projects in the target projection system. I believe that's the case but just wanted to make sure.

I'm not sure what you mean by the "target projection system".  The camera works in the map's coordinate system and transforms (through the view matrix) to the pixel coordinates.

> I am assuming that we will allow the Camera class to be over-ridden by different renderers.

The camera class can be overridden, but I don't think it should be dependent on the renderer.  We could have camera "compatibility" with renderers, but if the renderer is providing anything for the camera, then we get back into the realm of base layers and such, which is what I'm trying to avoid.

jbeezley:
@aashish24 DId you edit my comment? o_O

jbeezley:
> a SVG would not use projection matrix (ortho or perspective). Unless you are willing to put all the methods in the base class whether they are needed or not, we will need derived camera class.

Why not?  CSS [supports](http://www.w3.org/TR/css3-3d-transforms/#three-d-transform-functions) view matrices natively.  It will even do perspective transformations.

> Plus unless we have interactors (like VTK), a camera class will define how we move in the map space. Depending on the navigation type (navigation for globe like visualization would be different than navigation in 2D projection space). In that case also, we will need a derived camera object.

We do have an [interactor](https://github.com/OpenGeoscience/geojs/blob/master/src/core/mapInteractor.js).  We will certainly have to revisit the API and the interaction with the camera at some point, but I'm deferring those decisions as much as possible.

aashish24:
@jbeezley even though css supports perspective matrix, it won't be useful in our case unless I am missing something. I guess what I am trying to say is that we need to have camera switchable whether the renderer does it or someone else. It will be hard to have one camera class that does all sort of navigation. We can always have a default camera (the base). If we allow camera class to be overridden and have them set on the map, we should be good. As far as renderers are concered, they would have to get the data from the base camera and set the values internally.

aashish24:
The idea of not having dependency on renderer sounds good to me.

jbeezley:
For the camera, I have something like [this](http://stack.gl/packages/#mikolalysenko/3d-camera-core) in mind.  The base camera handles computing and caching the view/projection matrices.  Derived classes can add helper methods to manipulate the matrices, such as `lookAt`, `pan`, etc.

CSS's `matrix3d` transform will accept the matrices stored in the camera class and do the same calculations on the coordinates as is done in GL.  I'm not sure what the confusion is.

aashish24:
For the camera, I have something like this in mind. The base camera handles computing and caching the view/projection matrices. Derived classes can add helper methods to manipulate the matrices, such as lookAt, pan, etc.

+1 but remember we may also need to modify projection matrix for devices such as oculus VR which uses a fish eye projection (for the future). 

CSS's matrix3d transform will accept the matrices stored in the camera class and do the same calculations on the coordinates as is done in GL. I'm not sure what the confusion is.

No confusion. As far as we can set the camera and override stuff, we should be good.

aashish24:
> @aashish24 DId you edit my comment? o_O

Lack of sleep does it for you

jbeezley:
I hear having babies as that effect.

manthey:
@jbeezley I've made it so this works with the null renderer in the simple case.  I made some changes that could have been done in different manners (the tile offset, especially), and won't be offended if you'd rather do something else.

jbeezley:
Awesome, thank you!  You definitely saved me some gray hairs here.

aashish24:
This is great progress @jbeezley @manthey

manthey:
@jbeezley We may want to avoid drawing tiles that can be purged immediately (these are tiles that were asked to be loaded and then navigation changed such that they aren't desired any more).  Similarly, if navigation is rapid, we can end up with a lot of tiles in the browser's queue to be loaded that we know won't be necessary.  It would be nice to cancel these before the browser fetches the image, as that way it would be less delay for the images we actually need (and less network activity).  I don't think you can abort a promise, so some other mechanism for canceling tile loading would be required.

jbeezley:
I am aware that the performance is not great for that very reason.  I am planning to add a debounce somewhere (probably in the mouse interactor).  In any case, we could add a `cancel` method on `geo.tile`.  The question is exactly what it should do.  Just cancel the standard callbacks completely?  Should it call the error callback?  There needs to be away to resolve the promise even when it is canceled.  I'm thinking just adding a `cancelled` property to the tile object that the callback can query would be sufficient.

manthey:
@jbeezley I was talking with Charles about SlideAtlas.  In that program, he only renders the tiles via html, and doesn't have the gaps.  His method of preventing gaps was to *always* show the lower zoom level tiles.  This means that these tiles are in the background of the top tiles, and fill the gaps.  It also means that when you pan to a new area, there is always *some* low-resolution tile to be displayed (even if it is only the level 0 tile), which makes it so you never have an empty, unrendered area.  Obviously, you then could have roughly twice as many tiles loaded as with only the current zoom level.

manthey:
I just tried this out, and it works very well.  It also helped me hunt down a bug.  Should it be an option, or should I just switch to it?

aashish24:
> Should it be an option, or should I just switch to it?

Option would be nice but default should be what we had in SlideAtlas, always keep the low resolution tiles.

jbeezley:
I think we can switch to it.  There isn't that much overhead involved (25% to be exact).

manthey:
Pedantically, the overhead is more, because tiles can be partially off of the edges and we have to keep them for all zoom levels.  For instance, on a 1920x1280 screen, you might need 9*6 = 54 tiles, but keeping all lower zoom levels from zoom level 18 might require 145 tiles (4 at each zoom level is possible since we could be at a tile boundary, for instance).

jbeezley:
Okay, I didn't realize we needed all the lower zoom levels.  That is a bit worse.  In terms of loading the tiles, I don't think it will be a big deal because we want to do that anyway.  Rendering all the extra layers will probably take a noticeable performance hit especially on lower end hardware.

manthey:
I'm making it an option.  A fancier option (which I'm not adding) would be to allow setting the number of zoom levels, but by keeping all the zoom levels, you *always* have something to show when panning, except when wrapping.

aashish24:
> I'm making it an option. A fancier option (which I'm not adding) would be to allow setting the number of zoom levels, but by keeping all the zoom levels, you always have something to show when panning, except when wrapping.

Can't we just keep lower resolution tile that is one off compare to the current one? I also, didn't realize that slide-atlas keeps all the low res tiles.

jbeezley:
No, the lower resolution will still have gaps that are aligned with those at the current resolution (just roughly 1/2 as many).  The only way to guarantee that all of them are gone is to use all lower layers.

manthey:
If you pan, then as soon as you pan to an area that doesn't have a tile loaded and shown, you'll see the background color.  If you keep the *lowest* zoom level tile (zoom level 0), then there is guaranteed to be *something* other than the background.  If you only keep one level back, then you have an improvement for panning *if* you don't cross a power-of-two boundary, but not if you do.

You could just keep level 0 and the current level, but then the panning is very low res and doesn't look particularly nice.  Keeping all the levels definitely results in a nicer looking result.

Even better would be to also load tiles for some distance off the screen (based on how quickly someone is likely to pan and how long it takes to load a tile).

All of the extra tiles do have a performance hit, but I do not notice it on my desktop *and* it "seems" faster because there is already something there.  Here the perception is that the response is better, even if you can measurably show it to be worse.

aashish24:
> You could just keep level 0 and the current level, but then the panning is very low res and doesn't look particularly nice. Keeping all the levels definitely results in a nicer looking result.

that was the other option I had in mind. I think if you are not seeing much performance hit (we have to try with some real - use cases), then I think keeping them all (default) sounds good to me. On low performing system, user can turn it off (or may be in the future, we can determine the FPS and turn if off automatically).

manthey:
I've push the changes that have this setting, in case you want to try it out or see what I've done.

aashish24:
@manthey can you also add new / updated option for tileURL in your code as well? Please see here: https://github.com/OpenGeoscience/geojs/pull/463

manthey:
@jbeezley In the _getTile call to the `url` function, you have in the comments that `url` takes `x, y, z` as parameters, but you pass the `index` object with `{x: x, y: y, level: z}` instead.  I'm extending this so that subdomains can be used, and would rather match the documentation in the comments and have this function take `x, y, z`, but if you think it is better to take the object with `level` instead of `z`, I could be convinced.

jbeezley:
It sounds like an oversight on my part.  My intention was to include the ability to have both `level` and `z` where `level` is the zoom level.  `z` is *supposed* to be some other degree of freedom (possibly height, but not necessarily).  This is almost certainly not consistently used, as you've found.  Feel free to change it however you think is best.

manthey:
@jbeezley This now has a working d3 tile layer.  See my comments on the last push for my concerns.  I'd love to get some feedback on the choices I've made (or open criticism -- I'd be happy to refactor some of what I just did).

@aashish24 The url templating is in the last push.

jbeezley:
@manthey I'm with you in disliking the adjust layer methods, but I don't think it is particularly harmful.  Probably the worst thing from a user perspective is that `geo.d3.tileLayer() instanceof geo.d3.tileLayer` is false.

manthey:
@jbeezley How would we expect a user to specify the coordinate system they want to use for map centering and bounds?  Specifically, `boundsFromZoomAndCenter` is currently expecting coordinates in a meter-like coordinate, whereas in the past they were expecting to use degrees.

Perhaps there should be a way for the user to pass a coordinate system to bounds functions in case they want to use a different coordinate system than their set default.  This would also allow internal calls to make sure it is in whatever coordinates we want.

jbeezley:
@manthey I struggled with this as well.  Whether or not the map methods use the base layer's coordinate system in master isn't really specified and has caused bugs in the past.  In this branch, this fact is more obvious because the coordinate system is completely different.  I started to address the issue by adding an optional `gcs` argument to the coordinate conversion methods (for example [worldToGcs](https://github.com/OpenGeoscience/geojs/blob/refactor-projection/src/core/map.js#L478)).

Part of the issue is that we can't just enforce that the map navigation methods use `4326` because we want to use this in non-geospatial contexts.  There could be another configuration parameter on the map that determines what the input/output of coordinate systems are.

Another outstanding issue here, is how to set the global map extent within the map class.  At the moment, this information is provided by the tile layer, but it doesn't get propagated to the map.  I think it would be overly cumbersome to require the user to provide this information when creating the map object, but it is also conceivable that the map and each individual layer could also use different bounds.

aashish24:
> Perhaps there should be a way for the user to pass a coordinate system to bounds functions in case they want to use a different coordinate system than their set default. This would also allow internal calls to make sure it is in whatever coordinates we want.

Makes sense to me. I guess we would have to define our our coordinate class for this purpose as gcs may be too specific and domain specific vs picking a our own class that defines a set of coordinate systems (gcs, metric, etc..) and then internally we could do the conversion to whatever is the chosen coordinate system of the map when user constructed the view.

manthey:
Internally, we call the map transform parameter `gcs`.  I think we need an interface transform as well, which could default to `EPSG:4326`.  For non-geographic data, both of these transforms could be the empty string (which proj4 will properly treat as the identity).  Any opinions on what an interface transform parameter should be called?

manthey:
How much do we want to preserve backwards compatibility?  Specifically, the wsm example sets the projection by calling the `gcs` method on the osm layer (we now have this method on the `map`, not the `layer`), and uses a function called `tileUrl` also on the osm layer.  This example further uses some `geo.mercator` functions that don't exist any more.

I think in this example, reworking it makes more sense then trying to support all of the old interface, but I'm happy to support any set of functions and methods for legacy use, if appropriate.

jbeezley:
I don't think we should worry about backward compatibility.   It is broken in a number of ways by this branch implicitly.

aashish24:
> I don't think we should worry about backward compatibility. It is broken in a number of ways by this branch implicitly.

+1 I agree with Jon completely.

manthey:
Currently, many of the examples are broken.  Some are small looking issues (the legends widget looks like the css is weird), others are large issues (the hurricane example has interaction and coordinate system issues).  We can work through and fix these.

Otherwise, the only other issues (that I know of) is when you are viewing high zoom in the html or d3 renderers, the math precision fails on Chrome at level 18 and above, and on FireFox at level 17 and above.  We intend to set a new origin periodically to work around this.  For VGL, I see some subtle tile edge artifacts at zoom levels 20 and higher.

jbeezley:
@manthey I rebased onto the current master.  As far as I can tell, most things still work.  It introduced a bug rendering tiles in GL.  I'm sure it is trivial to fix, but I couldn't easily find the problem.  I'll start working on some of the simple CSS issues that cropped up in the examples.

aashish24:
thanks @manthey for the update! Would it be possible for you to add a new test to demonstrate biomedical  data viewing capability? Once all other tests are passing. It does not have to be in this branch.

manthey:
@aashish24 Yes.  For a stand-alone test, we want to serve tiles from an image statically (like we do for some of the map tests).  Do you have an image that we can share with the world that would be appropriate?

aashish24:
We could ask brian if he can send us something.

manthey:
I see two problems with when tiles are drawn or hidden.  Both are shown well by the wms example.
  
- When `keepLower = false`, we need to discard more tiles.  The `tileLayer.isCovered` method is used to determine when to hide tiles.  If you zoom in, the map at the edges of the display often have the `level - 1` tiles.   With partially transparent tiles these tiles are still visible.

- On the vgl renderer, if a tile is removed after the screen has rendered, it isn't taken away until the screen needs to be refreshed for another reason.

For the first of these, I think the solution is to purge tiles that are covered AND purge tiles at the non-current level when all tiles are successfully loaded.

For the second, we need to figure out when this occurs and rerender.

jbeezley:
Hey look, passing tests!

I'm okay with merging this in as is.  I know @manthey is still working on some performance issues, but those can possibly go in as a separate PR.

manthey:
I think all other changes should be separate PRs.  LGTM.

jbeezley:
No major objections have arisen in the last 5 months; I'm going to go ahead and merge this so that we can all move on with our lives.    Any additional concerns can be addressed in new issues.

aashish24:
Hurray!! Great job

PR #414, opened 2015-07-16
Point feature clustering

jbeezley:
*Imported from kwgitlab*

This adds the algorithm from [leaflet](https://github.com/Leaflet/Leaflet.markercluster) to perform hierarchical clustering for point features.  The algorithm itself is something like *O(n lg(n))* in the number of points and looks to work well.  To cluster the full cities dataset (30100 points) it takes just over 2 seconds.  So to do 100,000 would be on the order of 10 seconds.

The integration with the point feature has a number of issues that remain to be fixed.

1. The API for styling the clusters is ugly and doesn't provide a way to set defaults.  Currently when clustering is on the cluster objects will be passed through the standard style accessors along with the point data, so you have to look for a `d.__cluster` to determine if it is a cluster or a point.  See how it is done in the [test](https://kwgitlab.kitware.com/gumbo/geojs/blob/clustering/testing/test-cases/selenium-tests/pointClustering/include.js#L15).  Handling this would be much easier with the changes proposed in [295](https://github.com/OpenGeoscience/geojs/pull/295).  Specifically, the lazy evaluation of style accessors makes this almost impossible to handle independent from the renderer.

2.  There needs to be a way to style point features with text labels to show the total points contained in a cluster.  I have no idea how to do this in GL.

3.  The computation should be done in a web worker, or at least asynchronously.  When I tried to do it asynchronously the osm layer misbehaved terribly.

4.  At higher zoom levels, all points and clusters are added to the layer rather than just those that are visible.  I thought about detecting the viewport bounds and only adding points that were visible, but this seems like an optimization that should be done globally rather than just for this purpose.  Also, the viewport bounds really isn't handled well at the moment, and it is probably too renderer specific to do much about in general.

5.  The clustering radius, which is essentially how far apart points have to be in order to be clustered together is computed in world coordinates rather than pixel coordinates.  The side effect of this is that the clustering radius depends on the viewport size.  There is really no way for me to fix this with the current renderer API.  We need a function that provides conversion from world coordinates to pixel coordinates at arbitrary zoom levels rather than just the current zoom level.  [Leaflet](http://leafletjs.com/reference.html#map-project) provides this with the `map.project` and `map.unproject` methods.  In truth, the latlng to pixels ratio should be independent of the viewport size.  This is what other mapping libraries do.

See a demo of the clustering here: http://jsbin.com/setiba/6/edit?js,output


aashish24:
:cool: Will review it

src/core/pointFeature.js
```diff
@@ -25,7 +27,96 @@ geo.pointFeature = function (arg) {
...
+      d.__cluster = true;
+      d.__data = [];
+      d.obj.each(function (e) {
+        d.__data.push(m_allData[e.index]);
+      });
+      data.push(d);
+    });
+    m_ignoreData = true;
+    m_this.data(data);
+    m_this.layer().map().draw(); // replace with m_this.draw() when gl is fixed
```
aashish24: m_this.layer().draw() should work as well.

src/util/clustering.js
```diff
@@ -0,0 +1,265 @@
...
+      that._add(c);
+    });
+  }
+
+  /**
+   * Add a point or cluster as a child to the current cluster.
+   * @param {object} pt A ClusterTree or point object
+   * @private
+   */
+  ClusterTree.prototype._add = function (pt) {
```
aashish24: this class seems to be following a prototype based design. I am not against that but curious about the design decision.

aashish24: or is it because of the memory efficiency we get with prototype methods.

src/core/pointFeature.js
```diff
@@ -25,7 +27,96 @@ geo.pointFeature = function (arg) {
...
+      // short cut when there is nothing to do
+      return;
+    }
+
+    m_lastZoom = z;
+
+    var data = m_clusterTree.points(z).map(function (d) {
+      return m_allData[d.index];
+    });
+    m_clusterTree.clusters(z).forEach(function (d) {
```
aashish24: I have some idea what's going on here (just looking at the code, didn't run).  It would be nice to have some descriptive text here here for the future.

src/core/pointFeature.js
jbeezley: It should, but it doesn't currently.

src/util/clustering.js
jbeezley: Yeah, this is how it was implemented in the leaflet plugin.  This is definitely more common in the javascript world either directly or with the backbone-like `Class.extend` helper method.  It also is more compatible with the ES6 class syntax.

src/core/pointFeature.js
jbeezley: I added some more detailed comments to this file.

src/util/clustering.js
aashish24: Yeah, I was reading about the ES6 Class keyword. Looks interesting but I haven't looked into how it solves some of the most complicated inheritance in JS. I think it would be nice to think about refactoring our class syntax when it comes out though. For few reasons, I am not a big fan of prototype methods and Class seems to be a better option.

jbeezley: babel transpiles for you, so there isn't really a good reason to wait.  I'm not sure what you mean by complicated, but the class syntax is really only syntatic sugar for the normal prototypal patterns.

aashish24: Yup babel is one option. Reason for wait would be other features we need in short term (mostly at the feature or layer level). That's what I was referring to whether the Class just provides the syntactic sugar or something more powerful like access to superclass methods (based on your description it does not sounds like that).

aashish24:
Looking good but seems like clustering tests if failing now.

src/util/clustering.js
jbeezley: It does have `super` methods, but that is just syntatic sugar for `SuperClass.prototype.method.apply(this, ...)`

jbeezley:
That is a performance test that failed.. travis is slow, so I'm reducing the dataset size.

src/util/clustering.js
aashish24: Great, thanks, this is good to know.

aashish24:
Cool! thanks.

aashish24:
LGTM :+1:

aashish24:
Great job @jbeezley

PR #415, opened 2015-07-16
Update developer guide

jbeezley:
@danlamanna This should address the issues you had getting started.  Let me know if anything isn't explained well enough.

### DIFF
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -2,6 +2,11 @@
 Developer's guide
 =================
 
+.. note::
+
+    This guide assumes you have cloned and built the geojs repository
+    according to the :ref:`Quick start guide <project-setup-guide>`.
+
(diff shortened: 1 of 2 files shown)


aashish24:
LGTM :+1: @danlamanna please approve

danlamanna:
Looks good to me! I did this change on a fork because I'm unsure what level of verbosity you want - but this would also work for the checkout:
https://github.com/danlamanna/geojs/commit/9a8151bd15d0f09f07334c926ed64e8be77aa112

jbeezley:
I'm okay with more verbosity in the setup guide.  Most people are likely to check out the repository first and then realize that the submodule is needed.  I'll have another PR in a minute that will throw an error when you try to build without vgl checked out.  This has come up a few times lately.

BTW, I added you to developers here, so you can commit to this repo.

PR #416, opened 2015-07-17
Gruntfile improvements

jbeezley:
This make a few improvements to the gruntfile for building.
1. Tests the file with `jscs` and `jshint`
2. Reports an error when trying to build when `vgl` is not checked out.
3. Adds an option (`--env=production|development`) to switch between minified or concatenated source for the examples.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ script:
     - git submodule init
     - git submodule update
     - npm install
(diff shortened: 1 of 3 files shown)


aashish24:
Nice!!! specifically for 3)

aashish24:
LGTM :+1:

PR #422, opened 2015-07-27
Fixing position bugs

manthey:
This fixes the layer offset bug.

@aashish24, since you are away for a while, I cleaned up your code and added a test.  Also, this expects the parallel projection code to be merged in.  

@jbeezley This should pass once we fix the parallel projection selenium tests since it includes that branch.


manthey:
@jbeezley Do you know why this failed in the buildbot tests?

jbeezley:
I only looked at it briefly so far.  There are some changes here that could cause the map to display differently than before, but I'm pretty sure there is a bug in here somewhere as well.

jbeezley:
I fixed most of the issues with a one line change.  What is left is a problem with glTwoRenderers that I need to look into still.  The rest just need updated baselines.

aashish24:
@manthey no problem. Thanks, I was going to look at this later this week but appreciate your help here. 

@jbeezley let me know if you still have issues with two renderers. I am planning to do some work later this week. 

- aashish

aashish24:
@manthey let me rewrite my commit messages unless someone else has done it.

manthey:
@aashish24 Rewrite as you see necessary.

aashish24:
Done. Please see #426

Issue #418, opened 2015-07-20
Consider using ES6 (EMCA6) feature using Babel 


aashish24:
This is a low priority item but we should plan on doing this in next 2-3 months. At that time it would be good to discuss class design in GeoJS.

zachmullen:
Now that it's 3 years later, I think this has become more important -- using modern language features can drastically improve the development experience. At least, I would vote to prioritize this, especially since it's so easy to support this from where geojs is right now.

manthey:
I hacked this into a development environment, and the only issue I saw was with jQuery global usage (so addressing #862 would make this trivial).

PR #425, opened 2015-07-29
Widgets API

danlamanna:
This is the beginning of the high level widgets API, there are a number of things in this code that need to be cleaned up - but I think this is what we're looking for from an infrastructure standpoint.

This breaks the existing legend/slider widgets until we solidify this and rewrite those to conform.

### Overview

* Every widget is created as a top level element within the UI Layer, which used to be impeded by the renderer

* Widgets have basic positioning functionality, sticky or not, absolute positioned on a map according to viewport or GCS

* Widgets can tell whether they are in the viewport or not

* Widgets can be nested within each other

### Example Usage

    var widget = uiLayer.createWidget('dom', {
        sticky: true,
        positionType: 'gcs',
        position: {
            x: -73.7572,
            y: 42.6525
        }
    });
    
    widget.$el.attr('id', 'sticks-to-albany-div');

### Todo

* Canvas/el/$el is pretty confusing
@jbeezley - How do we feel about access to a jQuery object? In terms of convention, the UI layer canvas returns the DOM element so I suppose we should stick to that?

* svgWidget needs to inherit from domWidget

* svgWidget addChild
This needs to be implemented so that adding a child to an svgWidget correctly handles the grouping of SVG elements.

* Passing arg to widget init
I think I'm doing unnecessary variable passing - still need to look into it.

* Tests, documentation, updated examples
I think this will warrant a good amount of documentation and testing - waiting for the design decisions to settle before starting them.


src/ui/uiLayer.js
```diff
@@ -30,12 +30,15 @@ geo.gui.uiLayer = function (arg) {
...
       widgetName, m_this, m_this.renderer(), arg);
 
-    m_this.addChild(newWidget);
-    newWidget._init();
+    // We only want top level widgets to be a child of the uiLayer
+    if (!('parent' in arg)) {
+      m_this.addChild(newWidget);
+    }
+
+    newWidget._init(arg); // @todo shouldn't this be maintained in the object?
```
jbeezley: The `_init` and `_exit` methods are fairly inconsistently managed.  The convention has largely been that the parent object calls them for the child.  I would like to revisit this api once we get around to restructuring the classes for #418.

jbeezley:
This is looking great!

I would prefer to pass raw DOM elements through the public API.  I might consider removing the jquery dependency at some point.

danlamanna:
@jbeezley The slider widget mostly works. It doesn't appear at first because I create svg elements with no height/width specified - is there a reason for it being the whole viewport by default?

This code will generate a slider that works with the exception of parts disappearing on mouseover, still looking into that.
```
  var slider = uiLayer.createWidget('slider');

  $(slider.canvas()).attr({
    height: 200,
    width: 100
  });
```

danlamanna:
Left to do:
- [ ] Fix slider widgets mouseover events
- [ ] Make legend widget use the new widget API
- [ ] Add proper nesting (with svg groups) to svg widget
- [ ] Fix tests, add documentation

jbeezley:
> The slider widget mostly works. It doesn't appear at first because I create svg elements with no height/width specified - is there a reason for it being the whole viewport by default?

No, I don't think so.  You can probably get the actual size of the widget using a `getBBox` call and use that to set the svg size.

danlamanna:
0e653ff definitely broke a few things and will have to be fixed, but I want to make sure I'm not going down a wrong path by moving the functionality from the ```d3Renderer``` to the ```svgWidget```.

Aside from refactoring and ensuring tests pass, finding bugs, etc - the main problem left is handling the nesting of SVG widgets.

Nesting in the general sense of:
```
<svg id="foo">
    <svg id="bar"></svg>
</svg>
```

Already works by specifying a parent widget, since ```svgWidget``` inherits from ```domWidget```. So what's left is re-introducing the logic which allows the nesting of widgets as svg group elements.

@jbeezley Ping me if anything sounds far gone.

src/ui/widget.js
```diff
@@ -65,5 +70,95 @@ geo.gui.widget = function (arg) {
...
+    return ((position.left >= 0 && position.top >= 0) &&
+            (position.left <= map.width() && position.top <= map.height()));
+  };
+
+  if (arg &&
+      arg.hasOwnProperty('sticky') &&
+      arg.sticky &&
+      arg.hasOwnProperty('positionType') &&
+      arg.positionType === 'gcs') {
+    this.layer().geoOn(geo.event.pan, function () {
```
jbeezley: Are you doing anything with `zoom`?

jbeezley:
It looks like you can actually nest `svg` elements.  I don't really have a problem with doing that instead of making it a group element if it is easier.  The most likely use case I can imagine is nesting multiple svg widgets inside a dom widget anyway.

src/ui/svgWidget.js
```diff
@@ -0,0 +1,242 @@
+geo.gui.svgWidget = function (arg) {
```
jbeezley: This contains a lot of code from the d3 renderer.  Is this something you are in the process of refactoring?  I would want to think of a way to avoid the code duplication if possible.

danlamanna:
I think this is ready for review - 

The only "breaking" change this may have to applications is that the UI layer is now positioned relatively.

##### Testing
There are [some basic jasmine tests](https://github.com/OpenGeoscience/geojs/blob/dc0f1117471644509e26c90f9d507af42291599c/testing/test-cases/jasmine-tests/widgetApi.js) setup and passing, they could certainly be more exhaustive. @jbeezley Do you think there is any point at all to making browser-based (non-jasmine) tests since we're at a pseudo-shift in the way we test?

##### Documentation
I think the code I've added has pretty good documentation (note places you find it lacking) - but do we want an additional example, or something on readthedocs?

src/ui/widget.js
```diff
@@ -16,7 +16,23 @@ geo.gui.widget = function (arg) {
...
+      m_layer = arg.layer,
+      m_canvas = null;
+
+  this.parseArgs = function (args) {
+    var defaults = {
+      sticky: false,
+      positionType: 'viewport'
+    };
+
+    args = $.extend(defaults, args || {});
```
jbeezley: This actually mutates the defaults.

jbeezley:
I still need to go over the code, but it looks pretty good.  It would be nice to have an example of some sort.  Also, could you go over the [class overview](http://geojs.readthedocs.org/en/latest/users.html#class-overview) on rtd and update it for the new API?

aashish24:
@jbeezley @danlamanna if its not urgent, would it be okay if I look at the code as well before we merge this one. I am hoping to get it done by tomorrow.

jbeezley:
@danlamanna I made one small change to this branch that will make the example look better when it is merged with master.

src/ui/svgWidget.js
```diff
@@ -0,0 +1,78 @@
...
+
+  geo.gui.domWidget.call(this, arg);
+
+  var m_this = this,
+      m_renderer = null;
+
+  this._init = function (arg) {
+    var d3Parent;
+    arg = m_this.parseArgs(arg);
+    m_this.args = arg;
```
jbeezley: I would rather get rid of `widget.prototype.parseArgs`, move its code into `widget.prototype._init`, and then call that from here.  This should be calling the superclass `_init` anyway.  You can either place the options directly on `m_this` or add a get/set method for each like we do in [other classes](https://github.com/OpenGeoscience/geojs/blob/master/src/core/layer.js#L101).

jbeezley:
Other than my comment above, I'm happy this this.  Very nice!  I'll let @aashish24 have his say before final approval though.

danlamanna:
@aashish24 Nope not urgent - let me know what you think. 

I've caught a few small bugs myself that I still need to work on.

examples/widgets/main.js
```diff
@@ -0,0 +1,89 @@
...
+    .data(coordinates)
+    .style('radius', 5)
+    .style('fillColor', function () { return 'red'; })
+    .position(function (d) { return d; });
+
+  // Create a ui layer
+  var ui = map.createLayer('ui');
+
+  // Create a time series plot widget
+  var widget = ui.createWidget('dom', {
```
aashish24: Nice +1 for consistent API

src/ui/svgWidget.js
```diff
@@ -0,0 +1,78 @@
...
+      arg.parent.addChild(m_this);
+
+      // Tell the renderer there is an SVG element as a parent
+      d3Parent = arg.parent.canvas();
+    }
+
+    m_this._createCanvas(d3Parent);
+
+    m_this.canvas().addEventListener('mousedown', function (e) {
+      e.stopPropagation();
```
aashish24: I can see some possible use-cases where we don't want to do this but for now it will work.

src/ui/uiLayer.js
```diff
@@ -20,8 +20,14 @@ geo.gui.uiLayer = function (arg) {
   geo.layer.call(this, arg);
 
   var m_this = this,
+      s_init = this._init,
       s_exit = this._exit;
 
+  this._init = function () {
+    s_init();
+    m_this.node().css('position', 'relative');
```
aashish24: I am assuming that we will take care of this in the CSS refactor at some point.

aashish24:
LGTM :+1: good stuff, only had some minor suggestions some of them are beyond the scope of this branch.

src/ui/uiLayer.js
jbeezley: Yes, we will want to manage most css properties using classes like `.geo-map`, `.geo-layer.geo-ui-layer`, `.geo-widget`, etc.

jbeezley:
[LGTM](http://lgtm.in/i/KTGMqLMTp)

aashish24:
Nice image you got @jbeezley!.. very cool.

src/ui/svgWidget.js
danlamanna: The ```svgWidget``` constructor is [different enough](https://github.com/OpenGeoscience/geojs/blob/2b72de6a4162ebd392b0bdf86896cbcb5c87eebf/src/ui/svgWidget.js#L35-L42) that I didn't want it to call the superclass ```_init```. The only really common operation among all of the constructors is the argument parsing. It could probably be done more cleanly by actually setting them on ```m_this``` - but if possible, I would prefer to keep the svg/dom constructors separate.

danlamanna:
@jbeezley Take a look at 2e578db - getting it to "stick" to the top right throughout resizes is a bit cumbersome.

Since it's easier to just do ```uiLayerWidth - widgetWidth - margin``` do you think we should allow position to be some sort of function that takes the uiLayer node as an argument?

src/ui/svgWidget.js
jbeezley: This inherits from `widget` though.  `svgWidget` and `domWidget` are sibling classes.  You could even move the options parsing out of `_init` and put it directly in the constructor.  This is what is done in [`geo.map`](https://github.com/OpenGeoscience/geojs/blob/master/src/core/map.js#L68).  The subclasses all get the same `arg` object anyway so as long as you mutate it, the subclasses will automatically get the changes.

If I'm missing something, I don't have a huge problem with how this is done now, but I would want to save `args` as `m_this._args` instead to imply that it is a protected value.

jbeezley:
There may be a word definition problem here.  Sticky in the layer API means that it sticks to the map not to the screen.  What *should* happen in this case is that the legend will stay in place without having to reposition on pan.

There is a complication in that for core, the `sticky` parameter goes on the layer, but here it is on the widget.  What happens when the layer is sticky, but he widget is not?  Does it stay fixed relative to the layer or the map?

aashish24:
> There is a complication in that for core, the sticky parameter goes on the layer, but here it is on the widget. What happens when the layer is sticky, but he widget is not? Does it stay fixed relative to the layer or the map?

I think sticky here should refer to the fact that it is stick w.r.t layer or not. So if you move the layer does the widget moves with it or not. What you think?

ronichoudhury:
Is there a reason widgets would need a stickiness property independently of
the layers?  Why not just rely on the application writer to put widgets
into sticky/non-sticky layers as needed?  It seems to really complicate the
semantics of stickiness otherwise.

aashish24:
> Is there a reason widgets would need a stickiness property independently of
> the layers?  Why not just rely on the application writer to put widgets
> into sticky/non-sticky layers as needed?  It seems to really complicate the
> semantics of stickiness otherwise.

I could go either way. I think having a stick property for the widget would be nice but not required. It does complicate things. Cost of creating a dom layer shouldn't be that high so a more simple approach would be fine I guess. I don't think we will be creating thousands of widgets. It will be in the range of 1-20 for most applications.

danlamanna:
> Is there a reason widgets would need a stickiness property independently of the layers? Why not just rely on the application writer to put widgets into sticky/non-sticky layers as needed? It seems to really complicate the semantics of stickiness otherwise.

The initial idea was that since layers are somewhat heavy, being able to have only one UI layer with any amount of widgets, positioned however they each want to be positioned, was more desirable.

jbeezley:
I could go either way as well.  I can see enabling a layer that is fixed to the screen, but a widget that moves with the map.  The other way around is more difficult because you need to invert the transform of the parent, which will invariably lead to errors due to numerical precision at some point.  So the widget would wobble back and forth +/- 1 pixel (or more maybe).

For the moment, I'm in favor of doing what is the least amount of work as long as it doesn't limit us too severely at an API level.

aashish24:
> For the moment, I'm in favor of doing what is the least amount of work as long as it doesn't limit us too severely at an API level.

+1

I would suggest that here sticky should use layer as the reference point like I suggested earlier if that's not the case already.

aashish24:
@danlamanna looks like one of the style tests is failing (sliderWidget). Can you fix that? thanks

aashish24:
thanks! @jbeezley  can we merge this now? I am ok with removing stick from widgets.

jbeezley:
LGTM (for reals this time).

vgl
```diff
@@ -1 +1 @@
-Subproject commit 283eed5272f2fa311014a64416bf412cbcc0d068
+Subproject commit 013fe89bb16a21b75286c72bff9cdfd508cf9167
```
jbeezley: You need to reset this back to master.  `283eed5272f2fa311014a64416bf412cbcc0d068`

PR #426, opened 2015-07-29
Fixing position bugs update

aashish24:
Branch with cleaned up history.


aashish24:
Ref: #422

jbeezley:
I think I fixed the issue with the two renderers test.  It is terribly hacky because the camera doesn't have a "real" initialization method.  That whole code path is being updated in my branch so I didn't attempt to make the fix pretty.

jbeezley:
All tests passing now.  I'm good with this if others are.

manthey:
Looks good from here.

aashish24:
does resize not trigger when in testing? Otherwise looks good to me +1

jbeezley:
It does, but it needs to be done several times because various steps of the resize are skipped before the base layer is added.

aashish24:
Thanks. Seems like we need to improve this logic a bit. I will have a look at it.

jbeezley:
I wouldn't do any coding on it, I'm touching all the code related to cameras and coordinate conversions in my branch.  I'm cleaning up issues like this as I go.

aashish24:
Sounds good.

Issue #427, opened 2015-08-03
Contour features do not wrap around map


kotfic:
currently contour features do not wrap for full map datasets,  see example from add-contour branch of minerva:

![selection_001](https://cloud.githubusercontent.com/assets/669959/9046390/4f9e010c-39f8-11e5-8f47-23b64d66ef16.png)

Sample file here:

https://gist.githubusercontent.com/kotfic/b68f2585114e8ac5d2aa/raw/b5bdc9e31cdaa1131f1ce54e5ec4f3cb2dc18307/regridded_1deg_pr_amon_gfdl-esm2g_rcp45_r1i1p1_200601-210012.json

PR #428, opened 2015-08-04
Add a wrapLongitude option to the contourFeature.

manthey:
I am defaulting this to true, such that if stride positioning is used (x0, y0, dx, dy), if the data spans 180 degrees (plus or minus), it will be wrapped around so that part of the contour is at negative longitude and part is at positive longitude.  If the grid appears to span 360 degrees, join the ends (but only if wrapping was necessary).

In the general case, we might want to check the map coordinate system and only default to true if the x coordinate system is degree based (or even adjust it to the coordinate system so you could use gradians).

Added a jasmine test to make sure wrapping behaves as expected.

This fixes issue #427.

### DIFF
--- a/src/core/contourFeature.js
+++ b/src/core/contourFeature.js
@@ -173,10 +173,14 @@ geo.contourFeature = function (arg) {
         rangeValues = contour.get('rangeValues')(),
         valueFunc = m_this.style.get('value'), values = [],
         stepped = contour.get('stepped')(),
+        wrapLong = contour.get('wrapLongitude')(),
+        calcX, skipColumn, x, origI, /* used for wrapping */
+        gridWorig = gridW,  /* can be different when wrapping */
         result = {
           minValue: contour.get('min')(),
           maxValue: contour.get('max')(),
           stepped: stepped === undefined || stepped ? true : false,
+          wrapLongitude: wrapLong === undefined || wrapLong ? true : false,
           colorMap: [],
           elements: []
         };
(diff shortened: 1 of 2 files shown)


src/core/contourFeature.js
```diff
@@ -194,9 +198,55 @@ geo.contourFeature = function (arg) {
     if (gridW * gridH > data.length) {
       gridH = Math.floor(data.length) / gridW;
     }
+    /* If we are not using the position values (we are using x0, y0, dx, dy),
+     * and wrapLongitude is turned on, and the position spans 180 degrees,
+     * duplicate one or two columns of points at opposite ends of the map. */
+    usePos = (x0 === null || x0 === undefined || y0 === null ||
+        y0 === undefined || !dx || !dy);
+    if (!usePos && result.wrapLongitude && (x0 < -180 || x0 > 180 ||
+        x0 + dx * (gridW - 1) < -180 || x0 + dx * (gridW - 1) > 180) &&
```
aashish24: Can we think of a general function for this wrapping? This will be really useful for plane, graph, contour, polygon feature.

manthey: In the general case, this is pretty complicated.  For instance, points are easy, but what should lines do?  You don't necessarily want to change from drawing a line from 179E, 0N to 181E, 0N to drawing to 179E, 0N to 179W, 0N, as suddenly you've draw a line that is 358 degrees long rather than 2, and that could be very misleading.  Similarly, a polygon representing an island at straddling 180 degrees would become the width of the world.

Rather, I think the more correct action would be to *repeat* features every 360 degrees for all positions within the viewport.

Contours have the special case that they can wrap the globe, and therefore in certain instances, the grid of points should be 'closed'.  @kotfic has a case where he has a grid that is 360 data points wide with a 1 degree spacing.  We don't want to have a gap of 1 degree in the contour.  I'm not sure how that logic would extend to anything else.

So, yes, we should have a general way of repeating features so that we can zoom out, but I think it really should be instantiated in a different manner (as a completely different issue and PR).  When we do that, we can revisit contours and stop wrapping them (but still extend one extra column for a world-wide data set).

jbeezley: Agreed, what is done here is quite specific to the contouring algorithm.  I hadn't thought about just repeating features every 360 degrees, but that is a good idea.  I was already adding in my branch a concept of periodicity to the tile layer, but I can expose that through the map so that it can be used here.

src/core/contourFeature.js
```diff
@@ -357,6 +410,12 @@ layer.createFeature('contour', {
     value array>,
   dy: <the distance in the y direction between the 0th and (gridWidth)th point
     in the value array>,
+  wrapLongitude: <boolean (default true).  If true, AND the position array is
+    not used, assume the x coordinates is longitude and should be adjusted to
+    be within -180 to 180.  If the data spans 180 dgrees, the points or squares
```
jbeezley: s/dgrees/degrees/

src/core/contourFeature.js
```diff
@@ -194,9 +198,55 @@ geo.contourFeature = function (arg) {
...
+      }
+    }
+    /* Calculate the value for point */
     numPts = gridW * gridH;
     for (i = 0; i < numPts; i += 1) {
-      val = parseFloat(valueFunc(data[i]));
+      if (skipColumn === undefined) {
+        val = parseFloat(valueFunc(data[i]));
+      } else {
+        j = parseInt(Math.floor(i / gridW));
```
jbeezley: Did you mean `Math.floor(i / parseInt(gridW))`?  I don't think `parseInt` serves any purpose as is.

manthey: My bad.  Here I keep wishing javascript had explicit data types and want to ensure it is an integer.

jbeezley:
LGTM!

src/core/contourFeature.js
aashish24: > In the general case, this is pretty complicated. For instance, points are easy, but what should lines do? You don't necessarily want to change from drawing a line from 179E, 0N to 181E, 0N to drawing to 179E, 0N to 179W, 0N, as suddenly you've draw a line that is 358 degrees long rather than 2, and that could be very misleading. Similarly, a polygon representing an island at straddling 180 degrees would become the width of the world.

Couldn't agree more. What I was referring to the idea that when we figure out the winding we split the cell (and copy data as necessary) and start it on the other end. It is pretty common use-case in climate packages such as UV-CDAT and NCL.

jeffbaumes: For a datapoint, it is interesting to see how the "My Maps" in Google Maps works. Lines are always shown as geodesics (which makes it consistent what it means to go from point X to point Y - the lines are not straight on the map but are consistent when panning) and are repeated if you zoom out far enough.

PR #429, opened 2015-08-05
Added utf-8 charset to the example template

jbeezley:
@danlamanna It is very strange that the charset issue suddenly occurred, but this should solve the problem.


aashish24:
It is. I never ran into this issue before. What browser/version os/version you are using Dan?

jbeezley:
@danlamanna PTAL

danlamanna:
LGTM/Works!

PR #430, opened 2015-08-06
Add attribution notice

### LINKED ISSUE #281: OSM layer should provide an attribution notice

jbeezley:
This is long overdue.  It adds the ability to provide attribution notices per layer.  The map just combines them together and places them in the lower left corner as usual.  The osm layer is the only one that will have an attribution by default.

Fixes #281.

### DIFF
--- a/src/core/map.js
+++ b/src/core/map.js
@@ -596,6 +596,7 @@ geo.map = function (arg) {
       throw 'Map require DIV node';
     }
 
+    m_node.css('position', 'relative');
     if (arg !== undefined && arg.layers !== undefined) {
       for (i = 0; i < arg.layers.length; i += 1) {
         if (i === 0) {
@@ -1049,7 +1050,6 @@ geo.map = function (arg) {
         right: '0px',
         bottom: '0px',
         'padding-right': '5px',
-        'padding-left': '5px',
         cursor: 'auto',
         font: '11px/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif',
         'z-index': '1001',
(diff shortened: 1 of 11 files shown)


src/core/map.js
```diff
@@ -1021,6 +1022,63 @@ geo.map = function (arg) {
...
+        right: '0px',
+        bottom: '0px',
+        'padding-right': '5px',
+        cursor: 'auto',
+        font: '11px/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif',
+        'z-index': '1001',
+        background: 'rgba(255,255,255,0.7)',
+        clear: 'both',
+        display: 'block',
+        'pointer-events': 'auto'
```
manthey: I see you've created a separate issue #431 to convert this into a css file.  I feel that is highly desirable so that the user can easily override the style to change which corner the attribution appears in.  Since this will clobber any changes any time the attribution changes, until then, a program which changes attribution has to manually fiddle the css every time.

Is it worth adding an interim solution of passing an alternate css dictionary or a class with the attribution?

jbeezley: It isn't just here, there are a number of places we are injecting style onto the DOM.  I completely agree that it is important to allow the user to override these things.  That is why I added the issue.  Making that change is more work that I feel that I can allocate to it right now.  I think this PR is important to get in now because it is actually a legal issue.

manthey:
Since the css is a general issue that will eventually be addressed, this LGTM.

aashish24:
:clap:

src/core/map.js
aashish24: +1 @manthey @jbeezley

PR #434, opened 2015-09-01
Import Basic Choropleth Feature

dcjohnston:
To use, run grunt serve and then naviate to http://localhost:8082/examples/choropleth. Note that I've currently disabled multipolygon rendering (line 169 of core/choroplethFeature.js), to demonstrate the performance issues that arise between including and excluding multipolygons.


jbeezley:
Looking good, thanks.  We have some strict style rules [enforced](http://my.cdash.org/testDetails.php?test=20739000&build=817195) by jscs you'll need to keep in mind.  You should probably also rebase out https://github.com/OpenGeoscience/geojs/commit/06c8ce30804f526a7b8a913c247b4c0e0e7655ca (a reverted commit from my PR).

I'll fix the selenium tests (a firefox update I think :angry:) and look into the performance issues.

dcjohnston:
Gotcha, I'll make those style fixes + rebase now.

src/core/choroplethFeature.js
```diff
@@ -0,0 +1,247 @@
...
+                  {r: 0.656658579, g: 0.707001303, b: 1},
+                  {r: 0.821573924, g: 0.837809045, b: 1},
+                  {r: 0.943467973, g: 0.943498599, b: 0.943398095},
+                  {r: 1, g: 0.788626485, b: 0.750707739},
+                  {r: 1, g: 0.6289553, b: 0.568237474},
+                  {r: 1, g: 0.472800903, b: 0.404551679},
+                  {r: 0.916482116, g: 0.236630659, b: 0.209939162}
+                ],
+                scale: d3.scale.quantize(),
+                value: function (geoJsonFeature) {
```
aashish24: do you meant geojsFeature?

src/core/choroplethFeature.js
```diff
@@ -0,0 +1,247 @@
...
+   * @private
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  var m_this = this,
+      s_init = this._init,
+      s_data = this.data,
+      m_choropleth = $
+      .extend({},
+              {
+                minColor: 'black',
```
aashish24: The API is looking pretty good. Just wanted to make sure I am on the same page, what is the idea behind min and max here?

aashish24:
Looking very good @dcjohnsto. Just has some minor questions.

src/core/choroplethFeature.js
dcjohnston: Hey @aashish24, thanks for the feedback! 
(1) For the value method, it's input is a geoJSON feature, not the geoJS feature class. It defines the accessor used to pull a value from the input data, which as I have designed it will always be geoJSON. Is this correct? 
(2) I think minColor and maxColor are useless, we should just use the color range for all values. It was a holdover from the contour class that doesn't really serve a purpose here, so I'll go ahead and remove those values. I also haven't implemented opacity scaling, not sure if that's useful for choropleths, what do you think?

src/core/choroplethFeature.js
aashish24: @dcjohnston why we need geojsonFeature here? There is no class such as geojsonFeature may be @jbeezley added it somewhere. It would be nice if this feature is independent of what geojson is?

jbeezley: What? That is just a dummy argument name for the function.  This particular function is the default accessor that returns the "value" of the polygon.

That said the default here assumes a geojson-like object structure, which differs from the other feature classes.  For consistency, I think it would be better to make the default something like this:

```javascript
function (d) {return d.value;};
```

src/core/choroplethFeature.js
```diff
@@ -0,0 +1,245 @@
...
+   */
+  ////////////////////////////////////////////////////////////////////////////          
+  this._featureToPolygons = function (feature, fillValue) {
+
+    var arrayOfPolygonFeatures = [];
+    
+    if (feature.geometry.type === 'Polygon'){
+      arrayOfPolygonFeatures = [
+        m_this
+          ._addPolygonFeature(feature.geometry.coordinates, fillValue)
```
jbeezley: Because our interface is more general than geojson, this should be wrapped in an accessor that returns the array of coordinates like the `value` accessor.

src/core/choroplethFeature.js
aashish24: @jbeezley that's what was my initial comment but @dcjohnston replied this earlier (now it got deleted because of his commit): 

(1) For the value method, it's input is a geoJSON feature, not the geoJS feature class. It defines the accessor used to pull a value from the input data, which as I have designed it will always be geoJSON. Is this correct? 



> That said the default here assumes a geojson-like object structure, which differs from the other feature classes. For consistency, I think it would be better to make the default something like this:
> 
> function (d) {return d.value;};

+1

jbeezley:
@dcjohnston The feature classes are designed to have a d3-like interface rather than enforce any specific storage semantics.  Much like d3, we want to support even things like this:
```javascript
polygon
  .data(d3.range(10))
  .coordinates(function (_, i) {
    // return the ith polygon from some external or derived data source
  });
```

I hadn't thought about this issue before now, but really choropleths have been supported since the polygon feature was added.  Now that I look into this PR more closely, I think we could do one of two things:
  1. Add a choropleth feature that derives from polygon and just simplifies the interface.  I'm not sure that doing this would be of much value though because there isn't much in the way of simplification possible except for providing different default styles.
  2.  Move this code over to a "reader" or a utility method of some sort.  It would make sense to default to geojson for this and provide coloring algorithms as options.

aashish24:
> Add a choropleth feature that derives from polygon and just simplifies the interface. I'm not sure that doing this would be of much value though because there isn't much in the way of simplification possible except for providing different default styles.

Same here. I think not having it derived from polygon feature is okay and I am somewhat inclined towards it. 

> Move this code over to a "reader" or a utility method of some sort. It would make sense to default to geojson for this and provide coloring algorithms as options.

+1

src/core/choroplethFeature.js
```diff
@@ -0,0 +1,245 @@
...
+      ];
+      
+    }
+    if (feature.geometry.type === 'MultiPolygon') {
```
jbeezley: I see now the issue with multi-polygon performance.  One thing that isn't obvious from the class names is that geojs features **are** actually arrays of features.  The `polygonFeature` is an array of one or more polygons, so there isn't a need to do anything special form multi-polygons here.

Having a lot of instantiated feature objects (as this loop does) will kill the performance (particularly with the gl layers).  We definitely need to do some optimization of the feature classes in the future, but here you can get around it by instantiating a single polygon object.

ronichoudhury: I've always felt the word "feature" isn't being used quite correctly here.  To me, it's more like a "layer" that contains several "features" (i.e., "point feature" to me sounds like a single point on the map, not a bunch of points, bound to data, etc.).

jbeezley: I agree, but that convention has existed in the library long enough to be engrained in just about everything at this point.

ronichoudhury: Sure, but we're still in the 0.x series.  I think it's worth at least considering a change if the fallout of keeping it is persistent new user confusion.

dcjohnston: Want me to refactor into a method on the geoJSONReader class?

aashish24: > I've always felt the word "feature" isn't being used quite correctly here. To me, it's more like a "layer" that contains several "features" (i.e., "point feature" to me sounds like a single point on the map, not a bunch of points, bound to data, etc.).

@ronichoudhury it is typical in GIS and Graphics packages not to use plural. For example vtkPolyData would contain more than one point or lines or polygons. The idea is that a layer can contain multiple features (for efficiency or need point of view). Also, having multiple data points inside a feature is efficient from rendering point of view.

ronichoudhury: But "data" is already plural ;).

My issue is not with the implementation containing multiple objects, it's
naming the entire collection with the word "feature".

aashish24: I understand but feature is a common term used in this context in GIS world For example: http://support.esri.com/en/knowledgebase/GISDictionary/term/feature%20class

jbeezley: FWIW In leaflet, what we call `feature` is called a `featureGroup`, while a point is a single coordinate.

> Also, having multiple data points inside a feature is efficient from rendering point of view.

It doesn't have to be; it is entirely dependent on implementation details.

ronichoudhury: This defines a "feature class" to be a "collection of geographic
features".  Unless I'm missing something, that means a "point feature
class" on some map would be a bunch of "point features".  I am just
pointing out that we're not using the word "feature" in GeoJS consistently
either with common usage in English, or even with this GIS definition.

aashish24: > This defines a "feature class" to be a "collection of geographic
> features".  Unless I'm missing something, that means a "point feature
> class" on some map would be a bunch of "point features".  I am just
> pointing out that we're not using the word "feature" in GeoJS consistently
> either with common usage in English, or even with this GIS definition.

pointFeature is a class in GeoJS too. I think its verbose to name it pointFeatureGroup. I believe its just a matter of understanding GIS context and then it would make sense. I names classes keeping in mind that GIS folks would expect.

aashish24: > It doesn't have to be; it is entirely dependent on implementation details.

what I mean here is that we cannot have a class called pointFeature that will just store information for one point. That would be too in-efficient.

ronichoudhury: I don't mean to belabor the point but... even in a GIS context, "feature"
is in fact the wrong word to describe a collection of objects right?  Your
GIS definition marks "feature class" for that meaning, and "feature" for
the individual object.

aashish24: @ronichoudhury if I am not mistaken, in protovis, we had Mark classes to represent more than one mark. Please see here: http://mbostock.github.io/protovis/jsdoc/symbols/pv.Mark.html

> I don't mean to belabor the point but... even in a GIS context, "feature"
> is in fact the wrong word to describe a collection of objects right?  Your
> GIS definition marks "feature class" for that meaning, and "feature" for
> the individual object.

feature in GIS context means that collection of entities that share same or similar data. Also, this is not new in graphic packages as well. Renaming pointFeature to pointFeatures would be possible but somehow that has a odd feeling to it. pointFeature and point or dot has two distinct meanings.

aashish24: "Marks are associated with data: a mark is generated once per associated datum, mapping the datum to visual properties such as position and color. Thus, a single mark specification represents a set of visual elements that share the same data and visual encoding."

@ronichoudhury how about we discuss in person some other time if that's Okay? I do not intend to disregard this suggestion but probably something like this worth discussing at some other time.

ronichoudhury: Sure thing :)

dcjohnston:
Reading through all of that, now sure if I should be generalizing the choropleth feature to accept data formats of arbitrary structure (as opposed to only geojson), or bumping the class to some method on the geojson reader, thereby implying that the data is to be of a specific structure.

aashish24:
> Reading through all of that, not sure if I should be generalizing the choropleth feature to accept data formats of arbitrary structure (as opposed to only geojson), or bumping the class to some method on the geojson reader, thereby implying that the data is to be of a specific structure.

geojs code does not expect a particular format (format agnostic) and I think we should keep it that way as we may need to support other formats and not just geojson in the future. The default function for value can expect d.value and if the user wants to change it then can do so. We can always have adapters that can make it easy for the users.

dcjohnston:
Agreed, OK I'll generalize using the advice of @aashish24 + @jbeezley

jbeezley:
As I mentioned, we already have a generic choropleth class by way of the polygon.  Because this is a common use case, I think this would be more useful as part of a geojson reader with sane defaults for choropleth coloring and interaction.

aashish24:
@jbeezley that is a interesting point. GeoJSON reader produces points, lines and polygons but it does not really create a look up table of anysort for coloring. Also, Choropleth as a high level (as a feature will be useful from the user point of view). If we sub-class from polygon feature, wouldn't that take care of code duplication problem? I understand that user can put pieces together now but I think it would be easier if we have a feature that I can use with minimal effort.

jbeezley:
Ideally a lookup table should be available as a utility method or class.  It shouldn't be feature specific.

I think features should be generic and configurable (i.e. not just geojson), this implies using them will take more effort.  What I'm suggesting is a method that accepts geojson containing either an array of polygon geometries or a multipolygon geometry.  There can be some basic options as well for coloring algorithms and lookup tables, but it shouldn't be more generic than that.

This method might best be placed as a `reader` which is an overloaded term in the current code as readers can generate features from objects as well.  Another possibility is a utility method in `geo.util` or even as a set of "simple" methods under a new namespace.  I'm open to other ideas, but I don't think creating a new feature (subclassed or not) for this purpose is a good idea.

aashish24:
> Ideally a lookup table should be available as a utility method or class. It shouldn't be feature specific.

+1

> This method might best be placed as a reader which is an overloaded term in the current code as readers can generate features from objects as well. Another possibility is a utility method in geo.util or even as a set of "simple" methods under a new namespace. I'm open to other ideas, but I don't think creating a new feature (subclassed or not) for this purpose is a good idea.

@jbeezley since @dcjohnston already has a feature, lets go ahead and finish it. We can think about more general option as a next task. This is something we need to deliver to one of our client so I need this soon.

jbeezley:
In that case, I will echo @dcjohnston's confusion on what exactly it is that you want.  I have concerns pushing code just to check a box on a task list with the intention of completely changing it later.

aashish24:
> In that case, I will echo @dcjohnston's confusion on what exactly it is that you want. I have concerns pushing code just to check a box on a task list with the intention of completely changing it later.

My only suggestion for this branch is not to use assume geojson feature (for value accessor). And we can fix the performance issue than that would be great. I think this feature is useful as it is. The question of sub-classing or having some code in util is something we can discuss later.

aashish24:
Also, if we can put some of these function in util class and this feature uses them, that also sounds good to me as well.

jbeezley:
The geometries in this class assume the data is given as geojson, and those aren't configurable.  Why would the default value accessor assume anything different?  You can always change it using the `choropleth` method if it is.

aashish24:
okay, @dcjohnston.How about this, let you, @jbeezley and me talk on monday? If that's okay. Until then hold on working on this branch. I think we need bit more thinking.

dcjohnston:
Sounds good! Sorry for the confusion.

aashish24:
Oh please don't :smile: I think what you have done is great plus you know GeoJS now more than before. I have an idea that will address some of the concerns we discussed here and hopefully won't require much work. It is important to discuss some of these things (this is true for new features in general).

dcjohnston:
Quick update: I reimplemented the multipolygon rendering, so that each multipolygon creates only one new feature, however the performance is about as bad as before. I was wondering, would it make sense to group all the polygons that have the same fill value (based on choropleth scale), and then have each of these groups of identically-colored polygons be added to a single polygon feature? This seems like the path to fewest features, which could help with performance.

jbeezley:
I think you implemented it correctly.  Can you post or a link the geojson file you are testing with?

dcjohnston:
http://eric.clst.org/wupl/Stuff/gz_2010_us_040_00_20m.json

dcjohnston:
The performance gap i mentioned above is that given two states, for example, that have the same fill value (i.e. same scalar value), they each get their own polygon feature. It seems like it would be better for those two states to be passed into a single new polygon feature, so that there would be exactly one polygon feature per color.

jbeezley:
I'm looking into it now.  I think I know what you mean, but you do have to be careful with how you use the term feature to avoid confusion.  A polygon feature is one or more geojson polygon geometries.  I think you are talking about merging two polygon geometries into a single geometry if they are displayed with the same color.  That would probably cause a few problems internally, and I doubt it would do much to improve the performance.

I'll let you know what I figure out.

jbeezley:
I merged this branch with master because there were some incompatibilities with the `vgl` you were using.  With the new branch head, I'm not seeing any performance issues with the file you linked.  Could you give another try?  Make sure to `git submodule update` to ensure you are using the correct version of `vgl`.

dcjohnston:
Blazing fast :fire: ! Working great with the updated VGL. I'm plugging in the .scalar method to separate geodata from the scalar values used for coloring, and then I'll look at core/gl/choroplethFeature, since that definitely needs work. Thanks

aashish24:
This is awesome @dcjohnston!

dcjohnston:
Quick question: for the .scalar method, do we want to allow non bijective correspondance between scalar data and geo-data? I ask because, while it seems like it would offer more flexibility from the end-user's perspective, I'm not sure if there's a consistent best solution for how to aggregate the scalar values for the given feature. For instance, if we allows a "state" to have 10 scalar data points, do we determine it's color based on the average of these, the median of these, the mode? We could set average as default, but at this point, by adding a separate method for inputting scalar values, we're going to have a ton of customizations that the user will need to be aware of when the defaults aren't appropriate. I don't think there's anything inherently bad with that, but it seemed like the point of the feature was to make a quick and out of the box choropleth.

aashish24:
> Quick question: for the .scalar method, do we want to allow non bijective correspondance between scalar data and geo-data? I ask because, while it seems like it would offer more flexibility from the end-user's perspective, I'm not sure if there's a consistent best solution for how to aggregate the scalar values for the given feature. For instance, if we allows a "state" to have 10 scalar data points, do we determine it's color based on the average of these, the median of these, the mode? We could set average as default, but at this point, by adding a separate method for inputting scalar values, we're going to have a ton of customizations that the user will need to be aware of when the defaults aren't appropriate. I don't think there's anything inherently bad with that, but it seemed like the point of the feature was to make a quick and out of the box choropleth.

Can we not have the flexibility so that user can see all of those 10 scalars for a particular boundary? so 
for a csv like this: id scalar1 scalar2 scalar3 we get scalar1, scalar2, scalar3 in the function and then let user do whatever they want to do with that (via object).

dcjohnston:
Well, I think we can add a "custom aggregator" no problem. But it's necessary to define some sort of aggregation method, because otherwise how would you define the value to put into the color scalefor the region? 
In that last commit I'm just using average, but we could just make it another conf. option, defaulting to 'mean', but accepting 'median', and 'custom', which would allow the user to pass in a custom function to determine the aggregation method (maybe certain subsets of the scalar data could be more heavily weighted, for instance).

aashish24:
> In that last commit I'm just using average, but we could just make it another conf. option, defaulting to 'mean', but accepting 'median', and 'custom', which would allow the user to pass in a custom function to determine the aggregation method (maybe certain subsets of the scalar data could be more heavily weighted, for instance).

Having another conf option or second argument to scalar might work. What you think of seconfd argument to scalar? If none provided you can assume average.

aashish24:
@dcjohnston the tests are failing because of the baseline missing is that correct? @jbeezley are you going to upload new baseline?

jbeezley:
There are also some style issues, but I can take care of the baseline.

aashish24:
@dcjohnston @jbeezley any update on the baseline and style issues?

jbeezley:
I'll take a look at it today.

aashish24:
The changes and API looks great to me. Thanks @dcjohnston @jbeezley. I updated the branch with current master. Let's wait for the buildbot and then if all good, we can merge it.

aashish24:
LGTM :+1: :clap: @dcjohnston I am going to merge it (not sure if you are around).

dcjohnston:
Wonderful! I'm happy to get some of my code in such a cool project. Looking forward to building out more features soon!

jbeezley:
Thanks @dcjohnston, we're happy to have the help.

PR #437, opened 2015-09-28
Update jscs to version 2.0

jbeezley:


aashish24:
changes looks good to me but garant seems to have some issues.

jbeezley:
Downloads from Midas fail occasionally, I haven't figured out why yet.

aashish24:
Okay so not specific to this branch. LGTM :+1:

PR #438, opened 2015-09-28
Reenable coverage testing

jbeezley:
This replaces the endpoints that handled coverage results from the old python server with middleware attached to the new node express server.  Phantomjs coverage is [works](http://my.cdash.org/viewCoverage.php?buildid=831259) from travis.  At some point, I'll probably enable coveralls support as well, but that is low priority.


aashish24:
Thanks.. LGTM, we need to increase our coverage (at least 50% in next few months) :+1: It might be a good excercise for a hackathon.

aashish24:
:+1:  LGTM

PR #439, opened 2015-09-29
Switch to data.kitware.com for data storage

jbeezley:
This replaces the data store at midas3.kitware.com, which has been causing frequent test failures.


danlamanna:
LGTM.

Issue #435, opened 2015-09-10
JSON reader renders polygons as lines


danlamanna:
Seen [here](https://github.com/OpenGeoscience/geojs/blob/master/src/core/jsonReader.js#L186)

@aashish24

aashish24:
@dcjohnston care to fix this bug?

dcjohnston:
Sure thing, probably won't be able to touch until this tomorrow at the earliest, most likely Monday.

aashish24:
Thanks, that sounds fine to me. Also specially since this is sort of related to the awesome work you have done recently.

dcjohnston:
Submitted PR.

PR #447, opened 2015-10-09
Iss435: Add polygon & multipolygon support for geoJsonReader.

dcjohnston:
I've updated the geojson reader example to include both a polygon of arizona and a multipolygon of california. Both are being rendered properly on the map using the 'polygon' feature class. As far as style goes, I tried to keep my changes minimal, with most changes occuring in _addFeature and read methods of the jsonReader class.


aashish24:
@dcjohnston this is really cool. I am thinking we should start manual documentation on some of these. We have raedthedocs but I don't think its that user friendly. Should we start with Github wiki for now? What you think @jbeezley

sources.json
```diff
@@ -33,6 +33,7 @@
         "geomFeature.js",
         "graphFeature.js",
         "contourFeature.js",
+	"choroplethFeature.js",
```
jbeezley: This is already included.

sources.json
```diff
@@ -46,7 +47,8 @@
         "init.js",
         "wigglemaps.js",
         "leaflet.js",
-        "clustering.js"
+        "clustering.js",
+	"scales.js"
```
jbeezley: Could you satiate my OCD and fix the whitespace here?

jbeezley:
There are some whitespace only changes in `jsonReader.js` that I would prefer reverted as well as some legitimate jscs/jshint errors.  Other than those small things, this is great.

jbeezley:
Go ahead and merge it.

aashish24:
I want that number plate :+1:

danlamanna:
This will fix #435

Issue #448, opened 2015-10-09
Improve documentation (Tutorials, API documentation etc.)


aashish24:
We have gallery which is great, but we are missing tutorials for example. @nicopresto @jbeezley @dcjohnston what you think?

jbeezley:
There are two big things that should happen for documentation.

* Tutorials: I had started several jsbin examples (like [this one](http://jsbin.com/tikota/23/edit?js,output)) that were intended to fill that role.  We should embed jsbin examples like this with some prose in the style of the [romanesco examples](http://romanesco.readthedocs.org/en/master/examples.html) in the RTD docs.
* [API documentation](http://opengeoscience.github.io/geojs/apidocs/):  For this we need two things:
  1. Fill in missing/incomplete docstrings.  I have been doing this as I go along.  The new lines in my refactoring branch are probably half docstrings.
  2. Fix the jsdoc template to work better with our code structure.  For example, show the classes with their namespaces `geo.core.pointFeature` rather than just `pointFeature` to distinguish the derived classes.

aashish24:
> Tutorials: I had started several jsbin examples (like this one) that were intended to fill that role. We should embed jsbin examples like this with some prose in the style of the romanesco examples in the RTD docs.

@jbeezley are you suggesting we create RTD manually (vs parsing the source code like we do right now?). 

> API documentation: For this we need two things:
> Fill in missing/incomplete docstrings. I have been doing this as I go along. The new lines in my refactoring branch are probably half docstrings.
> Fix the jsdoc template to work better with our code structure. For example, show the classes with their namespaces geo.core.pointFeature rather than just pointFeature to distinguish the derived classes.

I have slightly different view. Talking with some other folks including @danlamanna, I am thinking that we should follow D3 like API documentation which I don't think comes from the source code. Does it?

jbeezley:
> are you suggesting we create RTD manually (vs parsing the source code like we do right now?).

We do create RTD manually.  Only the API docs come from the source.

> I have slightly different view. Talking with some other folks including @danlamanna, I am thinking that we should follow D3 like API documentation which I don't think comes from the source code. Does it?

[...]

PR #451, opened 2015-10-21
Add the ability to control the display order of layers

jbeezley:
The does the simplest possible thing (changing the `z-index` property).  Leaflet has a similar interface for layer order.


src/core/layer.js
```diff
@@ -42,13 +44,36 @@ geo.layer = function (arg) {
...
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Get or set the z-index of the layer.  The z-index controls the display
+   * order of the layers in much the same way as the CSS z-index property.
+   *
+   * @param {number} [zIndex] The new z-index
+   * @returns {number|this}
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this.zIndex = function (zIndex) {
```
aashish24: @jbeezley this is great. One question thought; How would a user know what z-index to use? Should we need to have a method that returns that (and knows all the z-indices used?). Or instead of having them set the z-index directly, we let the user set the order that will use the z-index. @mgrauer what would make sense from a GUI point of view?

jbeezley: There are many ways you can do it.  I went with this because it is what leaflet does.  It would be pretty easy to add methods like `moveToTop` on the layer.  We would want a method that sets the order explicitly to go on the map.  That too would be easy if you want me to add it.

mgrauer:
From the Minerva perspective, I like the idea of an api like

   * moveToTop
   * moveToBottom
   * upOneLayer
   * downOneLayer
   * hide (may not be necessary)
   * show (may not be necessary)

jbeezley:
One thing I didn't do here, which we may want, is to change the order of the layers as it returned by `map.layers()` or `map.children()`.  Doing so would change the semantics of a `map.layerOrder` method that refers to the layers by their index in those arrays.

jbeezley:
> From the Minerva perspective, I like the idea of an api like
>
> moveToTop
moveToBottom
upOneLayer
downOneLayer

I can do that.

> hide (may not be necessary)
show (may not be necessary)

[`layer.visible`](https://github.com/OpenGeoscience/geojs/blob/a13acf523125d916c2e83f2770cef37c7581849b/src/core/layer.js#L141) should work.

src/core/layer.js
aashish24: > I went with this because it is what leaflet does. It would be pretty easy to add methods like moveToTop on the layer. We would want a method that sets the order explicitly to go on the map. That too would be easy if you want me to add it.
 
+1. It would be nice if you can add these as part of this PR (or next, no strong preference). Having z index is nice :+1: too.

aashish24:
> upOneLayer
> downOneLayer

could be  moveUp(number) moveDown(number) for a generic version

mgrauer:
> could be moveUp(number) moveDown(number) for a generic version

fine by me

jeffbaumes:
May be overkill but d3's [sort](https://github.com/mbostock/d3/wiki/Selections#sort) method could be nice on the container if there are lots of layers.

aashish24:
> May be overkill but d3's sort method could be nice on the container if there are lots of layers.

+1. We could take a custom operator / function to sort layers..

aashish24:
> +1. We could take a custom operator / function to sort layers..

Not suggesting we do in this PR but if not, we can create an issue for it.

jbeezley:
I'll do the `moveUp`/`moveDown` here.  Applying a sort operator or index permutation would go on the map class anyway, so we can do it in a separate PR.

jbeezley:
PTAL

src/core/layer.js
```diff
@@ -42,13 +44,123 @@ geo.layer = function (arg) {
...
+
+    // set the default
+    if (n === undefined) {
+      n = 1;
+    }
+
+    // set the sort direction that controls if we are moving up
+    // or down the z-index
+    sign = 1;
+    if (n < 0) {
```
aashish24: clever

src/core/layer.js
```diff
@@ -467,6 +562,19 @@ geo.layer = function (arg) {
     return m_height;
   };
 
+  if (arg.zIndex === undefined) {
+    arg.zIndex = m_map.children().length;
```
aashish24: Nice

aashish24:
LGTM :tada: One request would be to update the RTD on how to use this feature but other than that

SHIP IT!

jbeezley:
The methods will be documented in the api docs.  We don't really have a place for this kind of thing on RTD at the moment.  This would be a good thing to put in a tutorial for #448.

aashish24:
> The methods will be documented in the api docs. We don't really have a place for this kind of thing on RTD at the moment. This would be a good thing to put in a tutorial for #448.

Sounds good @jbeezley. Should I start putting some basic things in our current RTD for it?

jbeezley:
:+1:

PR #455, opened 2015-10-23
Add the ability to "pause" the map interactor

jbeezley:
This is useful so that applications can temporarily disable the interactor while displaying a custom UI element.

### DIFF
--- a/src/core/mapInteractor.js
+++ b/src/core/mapInteractor.js
@@ -25,7 +25,8 @@ geo.mapInteractor = function (args) {
       m_wait = false,
       m_disableThrottle = true,
       m_selectionLayer = null,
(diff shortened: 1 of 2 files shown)


jeffbaumes:
Will this still allow tiles to be loaded and rendered during the pause? I think it should.

jbeezley:
Yes, the only thing pausing does is ignore native mouse/keyboard events.

jeffbaumes:
Great.

aashish24:
@jbeezley this is great stuff! I was just thinking if _ignoreEvent or _skipEvent would have been better. But other than that :+1: LGTM 

https://dl.dropboxusercontent.com/u/72484/crossguard-lightsaber-lgtm.gif

PR #456, opened 2015-10-23
Click event improvements

jbeezley:
This is motivated by a conversation with @mbertrand about the click events.  With this PR, click events will always prevent other synthetic events (pan, zoom, brush, etc.) from occurring from the same native event.  Exactly how a click is detected is configurable via the options:
 * `options.click.enabled`: to enable globally
 * `options.click.buttons`: to enable per mouse button
 * `options.click.duration`: trigger a click only when the time between `mousedown` -> `mouseup` is less than this many milliseconds.  This check is disabled when the value is zero (the default).
 * `options.click.cancelOnMove`: a click will not be triggered if there is a `mousemove` between the `mousedown` and `mouseup`

This will almost certainly conflict trivially with #455.  I'll rebase as necessary.

### DIFF
--- a/src/core/mapInteractor.js
+++ b/src/core/mapInteractor.js
@@ -26,7 +26,8 @@ geo.mapInteractor = function (args) {
       m_disableThrottle = true,
       m_selectionLayer = null,
       m_selectionPlane = null,
-      m_paused = false;
+      m_paused = false,
+      m_clickMaybe = false;
 
   // Helper method to decide if the current button/modifiers match a set of
   // conditions.
@@ -97,6 +98,12 @@ geo.mapInteractor = function (args) {
       spring: {
         enabled: false,
         springConstant: 0.00005
+      },
+      click: {
+        enabled: true,
+        buttons: {left: true, right: true, middle: true},
+        duration: 0,
+        cancelOnMove: true
       }
     },
     m_options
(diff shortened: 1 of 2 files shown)


src/core/mapInteractor.js
```diff
@@ -779,10 +826,34 @@ geo.mapInteractor = function (args) {
...
+
     m_this._getMouseButton(evt);
     m_this._getMouseModifiers(evt);
 
-    // fire a click event here
+    // cancel any ongoing pan action
+    m_this.cancel('pan');
+
+    // unbind temporary handlers on document
+    $(document).off('.geojs');
```
aashish24: @jbeezley why this is required?

aashish24:
@jbeezley @mbertrand  I liked the idea +1. It would be nice if we can add some documentation for in in the RTD.

src/core/mapInteractor.js
jbeezley: `mousedown` adds event handlers to `document` to track the mouse even if it moves outside the map.  This is only needed during an ongoing pan/zoom so the handler is added temporary.  This just ensures that the handlers get unbound when stopping the pan action that had been initiated.

aashish24: Ah, I see, thanks.

aashish24:
@jbeezley didn't run the code, but in general it LGTM :+1:

aashish24:
LGTM :+1:

PR #457, opened 2015-10-26
Fix position of ui layers

jbeezley:
Fixes an [issue](https://github.com/Kitware/minerva/pull/166#issuecomment-150671496) with the ui layer when there is another layer on top.


jbeezley:
Included in #458.

PR #458, opened 2015-10-27
Fix jasmine tests

jbeezley:
This fixes an issue with the jasmine test runner that caused ctest to pass (almost) all asynchronous tests.

Depends on #457 which contains a fix to a bug exposed by this change.


danlamanna:
[LGTM](http://lgtm.in/i/xgu2)

PR #460, opened 2015-10-28
Remove remaining references to vgl submodule

jbeezley:


aashish24:
@jbeezley does vgl.js will be minified (or not) based on how we are building geojs (its been a while I tried to debug).

aashish24:
Had a related question but other than that :+1:

jbeezley:
It's minified in the minified library, but just concatenated in the other.

aashish24:
Great..thanks!

PR #463, opened 2015-11-03
Added option to support subdomains in tileUrl

aashish24:


aashish24:
@jbeezley I need to add a test. Should I add a regression test (image) or some sort of unit test?

src/core/osmLayer.js
```diff
@@ -43,6 +43,7 @@ geo.osmLayer = function (arg) {
     m_numberOfCachedTiles = 0,
     m_tileCacheSize = 100,
     m_baseUrl = 'http://tile.openstreetmap.org/',
+    m_subdomains = 'abc',
```
jbeezley: This really isn't general enough.  See [here](http://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Tile_servers)

src/core/osmLayer.js
```diff
@@ -108,7 +117,8 @@ geo.osmLayer = function (arg) {
   ////////////////////////////////////////////////////////////////////////////
   m_tileUrlFromTemplate = function (base) {
     return function (zoom, x, y) {
-      return base.replace('<zoom>', zoom)
+      return base.replace('{s}', m_this._getSubdomain(x, y))
```
jbeezley: I would either do all `<...>` or `{...}` templates rather than mixing the two.

src/core/osmLayer.js
aashish24: @jbeezley yes, for otile, we have 1,2,3,4, and for some others a,b,c (I know that) sicne we used default as tile.open***.org, I set the default to abc. Are you suggesting we leave it empty?

src/core/osmLayer.js
aashish24: @jbeezley good point, will change it. I have seen more {} vs <>. Do you have any preference for one or the another?

jbeezley:
Preferably unittests.  In other places, I have used a [mocked map](https://github.com/OpenGeoscience/geojs/blob/master/testing/test-cases/phantomjs-tests/mapInteractor.js#L25) to avoid generating an actual map object and to track the internal calls made.  It would probably be good to expose something like this more generally for testing.

src/core/osmLayer.js
jbeezley: I didn't notice it was configurable via the options.  In that case, can you document the option in the class docstring?

src/core/osmLayer.js
jbeezley: It seems that `{}` is used elsewhere.  I would say change it to that.

aashish24:
> Preferably unittests. In other places, I have used a mocked map to avoid generating an actual map object and to track the internal calls made. It would probably be good to expose something like this more generally for testing.

Thanks. I will see if I can find a test that is similar. Are we using mocked map for unit testing?  I agree it would be nice to have it somewhere else for general purpose testing.

jbeezley:
Yes, in the test I linked above. https://github.com/OpenGeoscience/geojs/blob/master/testing/test-cases/phantomjs-tests/mapInteractor.js#L25

I am doing some more elaborate mocking in some of the tests in the refactor branch as well.

aashish24:
> Yes, in the test I linked above. https://github.com/OpenGeoscience/geojs/blob/master/testing/test-cases/phantomjs-tests/mapInteractor.js#L25

okay, thanks. Should I move mocked map to a common location if that's okay in my branch?

aashish24:
@jbeezley I also changed to {z} vs {zoom} since {z} seems to be conventional.

src/core/osmLayer.js
aashish24: > I didn't notice it was configurable via the options. In that case, can you document the option in the class docstring?

Right, we have missing documentation. I was wondering about that and thought that you and @manthey may have added it in the refactored code. I will add the docs as well.

jbeezley:
It could be worthwhile to move that code.

In other news, it looks like there is still a race condition in the phantomjs test startup.  Blanket is starting the test runner before jasmine is done loading.  One of these days I want to move to a standard test runner like jsunit.

src/core/osmLayer.js
jbeezley: This entire file is gone in the refactored branch.

aashish24: > This entire file is gone in the refactored branch.

Right, should I skip testing then and create a issue instead? Also, someone should merge master in the refactored code and make sure all the new features are not lost.

aashish24:
@jbeezley I am thinking to add test in the newer code, for now, I will mark it done and create a issue for 

* Testing
* Porting to new code -- @manthey do you want to take care of this?

aashish24:
@jbeezley?

jbeezley:
I suppose this is okay, but we will need to be more picky about PR's once the refactor lands.

aashish24:
I agree :+1: .  I am hoping that the refactored code is going to land soon.

PR #466, opened 2015-11-19
Add `--allow-root` to bower instantiation

jbeezley:
This is necessary to install geojs as the root user.  It is the equivalent of `--unsafe-perm` for npm.

@mgrauer @kotfic


mgrauer:
LGTM.

PR #467, opened 2015-11-20
Discrete zoom and mouse event throttling

jbeezley:
This adds new utility methods for throttling and debouncing functions.  These methods are used in the map interactor to throttle all mouse events to improve rendering performance.  In addition, zoom events are debounced to fix discrete zooming on touch devices, which fire small continuous scroll events rather than large discrete events like a traditional mouse wheel.

The interaction with discrete zooming isn't great for a couple of reasons.

1.  There is a significant delay that occurs while the map is rendering
2.  There is no indication that the zoom is doing anything at intermediate values

It may help to implement a "snap to discrete values" feature, but that is out of scope of this PR.  You can try the new behavior at the following link.

http://jsbin.com/feduco/3

### DIFF
--- a/src/core/mapInteractor.js
+++ b/src/core/mapInteractor.js
@@ -22,11 +22,9 @@ geo.mapInteractor = function (args) {
       m_mouse,
       m_keyboard,
       m_state,
+      m_queue,
+      m_throttle,
       $node,
-      m_wheelQueue = { x: 0, y: 0 },
-      m_throttleTime = 10,
-      m_wait = false,
-      m_disableThrottle = true,
       m_selectionLayer = null,
       m_selectionPlane = null,
       m_paused = false,
@@ -52,33 +50,11 @@ geo.mapInteractor = function (args) {
     return Math.sqrt(x * x + y * y);
   }
 
-  // For throttling mouse events this is a function that
-  // returns true if no actions are in progress and starts
-  // a timer to block events for the next `m_throttleTime` ms.
-  // If it returns false, the caller should ignore the
-  // event.
-  function doRespond() {
-    if (m_disableThrottle) {
-      return true;
-    }
-    if (m_wait) {
-      return false;
-    }
-    m_wait = true;
-    window.setTimeout(function () {
-      m_wait = false;
-      m_wheelQueue = {
-        x: 0,
-        y: 0
-      };
-    }, m_throttleTime);
-    return true;
-  }
-
   // copy the options object with defaults
   m_options = $.extend(
     true,
     {
+      throttle: 30,
       panMoveButton: 'left',
       panMoveModifiers: {},
       zoomMoveButton: 'right',
@@ -114,6 +90,10 @@ geo.mapInteractor = function (args) {
 
   // options supported:
   // {
+  //   // throttle mouse events to at most this many milliseconds each (default 200)
+  //   // Note: this value is immutable for a given instantiation of this class.
+  //   throttle: number
+  //
   //   // button that must be pressed to initiate a pan on mousedown
   //   panMoveButton: 'right' | 'left' | 'middle'
   //
@@ -178,6 +158,10 @@ geo.mapInteractor = function (args) {
   //     cancelOnMove: true // cancels click if the mouse is moved before release
   //   }
   // }
+
+  // set the value now to avoid rebinding all kinds of handlers whenever
+  // settings change.
+  m_throttle = m_options.throttle;
 
   // A bunch of type definitions for api documentation:
   /**
@@ -330,6 +314,16 @@ geo.mapInteractor = function (args) {
   //  }
   m_state = {};
 
+  /**
+   * Store queued map navigation commands (due to throttling) here
+   * {
+   *   kind: 'move' | 'wheel',  // what kind of mouse action triggered this
+   *   method: function () {},  // the throttled method
+   *   scroll: {x: ..., y: ...} // accumulated scroll wheel deltas
+   * }
+   */
+  m_queue = {};
+
   ////////////////////////////////////////////////////////////////////////////
   /**
    * Connects events to a map.  If the map is not set, then this does nothing.
@@ -349,10 +343,15 @@ geo.mapInteractor = function (args) {
 
 
     // add event handlers
+    if (m_options.throttle > 0) {
+      $node.on('wheel.geojs', m_this._handleMouseWheel);
+    } else {
+      $node.on('wheel.geojs', m_this._handleMouseWheel);
+    }
+
     $node.on('mousemove.geojs', m_this._handleMouseMove);
     $node.on('mousedown.geojs', m_this._handleMouseDown);
     $node.on('mouseup.geojs', m_this._handleMouseUp);
-    $node.on('wheel.geojs', m_this._handleMouseWheel);
     // Disable dragging images and such
     $node.on('dragstart', function () { return false; });
     if (m_options.panMoveButton === 'right' ||
@@ -552,6 +551,8 @@ geo.mapInteractor = function (args) {
       out = m_state.action === action;
     }
     if (out) {
+      // cancel any queued interaction events
+      m_queue = {};
       m_state = {};
     }
     return out;
@@ -569,10 +570,8 @@ geo.mapInteractor = function (args) {
       return;
     }
 
-    if (m_state.action === 'momentum') {
-      // cancel momentum on click
-      m_state = {};
-    }
+    // cancel momentum on click
+    m_this.cancel('momentum');
 
     m_this._getMousePosition(evt);
     m_this._getMouseButton(evt);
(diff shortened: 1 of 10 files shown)


aashish24:
@jbeezley this is great. One thing I noticed that (using my mouse only) is that with my mouse wheel if I scroll it once, nothing happens which is unlike other mapping libraries. Is that what you were referring to?

src/core/mapInteractor.js
```diff
@@ -344,12 +339,20 @@ geo.mapInteractor = function (args) {
     // store the connected element
     $node = $(m_options.map.node());
 
+    // set methods related to asyncronous event handling
+    m_this._handleMouseWheel = throttled_wheel();
+    m_callZoom = debounced_zoom();
 
     // add event handlers
+    if (m_options.throttle > 0) {
+      $node.on('wheel.geojs', m_this._handleMouseWheel);
```
aashish24: @jbeezley both if and else are doing the same thing.. isn't it?

src/core/mapInteractor.js
```diff
@@ -111,6 +88,14 @@ geo.mapInteractor = function (args) {
 
   // options supported:
   // {
+  //   // throttle mouse events to at most this many milliseconds each (default 200)
```
aashish24: it says default is 200 but above I see is 30?

aashish24:
@jbeezley I tried it on mac as well using trackpad. Something is not quite working for me (jsbin demo). May be I am missing something but the user-experience seems bit unusual. @manthey @mgrauer can you try jsbin demo as well?

src/core/mapInteractor.js
jbeezley: Yes, good point.

jbeezley:
>  One thing I noticed that (using my mouse only) is that with my mouse wheel if I scroll it once, nothing happens which is unlike other mapping libraries. Is that what you were referring to?

I need to adjust the scaling or perhaps use a smarter method to make one wheel "click" correspond to a +/- 1 zoom change.  One issue is that a single "click" doesn't always return the same value for all devices/settings.  I may resort to truncating all single mouse wheel events greater than some threshold to +/- 1.

> Nit-picking but it seems that the comment style here different than used else where (see for m_queue)

The original comments on this file are not parseable by jsdoc.  I have started to make new comments so they will appear in the api docs.  Much of this whole block of old style comments needs to be modified to work correctly.

> I tried it on mac as well using trackpad. Something is not quite working for me (jsbin demo). May be I am missing something but the user-experience seems bit unusual.

Yes, it is, bordering on unusable really.  The issue is a lack of feedback combined with the delay when rendering combined with the fact that apple trackpad-like devices have a built-in "momentum"-like behavior (they continue sending scroll events after you stop scrolling).

This is the current logic used for discrete zooming:

1. Get wheel event
2. Add the magnitude to a static value (accumulator)
3. If the accumulator is > 1 in magnitude, alter the zoom level by `Math.round(accumulator)` and reset the accumulator to the remainder
4. Wait for another wheel event for 400 ms, if none occurs in that time reset the accumulator to 0

An alternative approach may be better:

1. Get wheel event
2. Add the magnitude to a static value (accumulator)
3. Wait 400ms for another wheel event
4. If no other wheel event occurs, zoom the map by +/- 1 depending on the sign of the accumulator

This would mean **any** debounced wheel event would trigger either a +1 or -1 zoom level change.  On a trackpad-like device, you would need to make successive "swipes" to make multiple zoom level changes.  The possible down side is that actual wheel events would be ignored if they occur too fast.

I would be interested in anybody's ideas or opinions about this.

aashish24:
> I need to adjust the scaling or perhaps use a smarter method to make one wheel "click" correspond to a +/- 1 zoom change. One issue is that a single "click" doesn't always return the same value for all devices/settings. I may resort to truncating all single mouse wheel events greater than some threshold to +/- 1.

Sure. 

> An alternative approach may be better:
> 
> Get wheel event
> Add the magnitude to a static value (accumulator)
> Wait 400ms for another wheel event
> If no other wheel event occurs, zoom the map by +/- 1 depending on the sign of the accumulator

Let me think more about it. In the mean time, do you have newer jsbin?

jbeezley:
I didn't change anything that would effect the behavior.

aashish24:
> I didn't change anything that would effect the behavior.

I see. I just saw the commits now. thanks.

mgrauer:
I played around with the jsbin and with a geojs in Minerva, I'm reporting on opinions.

Mouse wheel and Mac "magic trackpad" scrolling both feel "choppier" in the jsbin than in Minerva, not sure how much of that is due to full screen map vs smaller panel map.  The choppier trackpad is a better experience in the jsbin than in Minerva because it slows down the zooming to reasonable levels.  The choppier mouse wheel is a worse experience in the jsbin than Minerva, but only by a little bit, i.e. this change is a net benefit.

My magic trackpad "momentum" scrolling can overzoom either jsbin or Minerva versions, if I give it a good slide.

I'd be interested to try the suggested alternative approach if it is easy to setup a demo.  I'm worried about the user experience if I would have to scroll many separate times to zoom in or out, but would like to feel it.

From the Minerva perspective, we can always add zoom +/- buttons, if that helps.

aashish24:
> Get wheel event
> Add the magnitude to a static value (accumulator)
> Wait 400ms for another wheel event

I would think that 400 ms is too long as in general the default is 200ms but I wouldn't know until I look into this some more. 

> If no other wheel event occurs, zoom the map by +/- 1 depending on the sign of the accumulator

And if does then I am assuming it will be > +/-1?

> The possible down side is that actual wheel events would be ignored if they occur too fast.

I am not 100% sure if I follow here but  I guess it relates to my question above.

jbeezley:
> I would think that 400 ms is too long as in general the default is 200ms but I wouldn't know until I look into this some more.

I don't think this particular value makes much difference, it only determines when the remainder is thrown away.

I just looked at what [Leaflet](http://leafletjs.com/) does.  It seems to be very similar to what I have done here.  The difference is that the leaflet renderer is far more responsive than ours.  By this I mean it starts the redraw almost immediately.  GeoJS seems to get "choppy" and the rendering freezes.  I did a profile and it looks like the method [here](https://github.com/OpenGeoscience/geojs/blob/master/src/gl/planeFeature.js#L114) is blocking everything.  In the screenshot below, it blocks the main thread for over 3/4 of a second:

screen shot 2015-11-23 at 3 53 15 pm

aashish24:
> I don't think this particular value makes much difference, it only determines when the remainder is thrown away.

thanks, I agree it won't then. 

> 
> I just looked at what Leaflet does. It seems to be very similar to what I have done here. The difference is that the leaflet renderer is far more responsive than ours. By this I mean it starts the redraw almost immediately. GeoJS seems to get "choppy" and the rendering freezes. I did a profile and it looks like the method here is blocking everything. In the screenshot below, it blocks the main thread for over 3/4 of a second:

that would be the very first time image get's loaded but it shouldn't be almost a second, that seems too large. Can you verify if that's always the case?

jbeezley:
It shouldn't matter if it is the first time or not, there is no excuse for blocking the browser for that long for any reason.  Especially when all that is happening is rendering a handful of small pngs.

Drilling down into it, the issue is compiling the shaders.  It spends almost 100% of the time [here](https://github.com/OpenGeoscience/vgl/blob/283eed5272f2fa311014a64416bf412cbcc0d068/src/shader.js#L91). It is compiling 4 shaders for every image it loads.  Is this really necessary?  Can't all of the tiles use the same shader program?

aashish24:
> It shouldn't matter if it is the first time or not, there is no excuse for blocking the browser for that long for any reason. Especially when all that is happening is rendering a handful of small pngs.

I agree. what I was wondering is that we should have seen this in current master as well (sounds like it should be happening but may be not realized because of the way our interactor works in master). 

> Drilling down into it, the issue is compiling the shaders. It spends almost 100% of the time here. It is compiling 4 shaders for every image it loads. Is this really necessary? Can't all of the tiles use the same shader program?

Good point. They should be using the same material (shader + other attributes) and just need to bind their specific texture. I will push a branch today.

jbeezley:
You can observe the behavior in http://opengeoscience.github.io/geojs/examples/osm/ as well.  It begins to block the browser half a second or so after you stop navigating with the mouse.  At that point, it begins to load the new textures and everything is unresponsive.  It is more noticeable here because you don't get any visual response at all until after render is complete.

Sharing a shader program would help immensely.  Compiling the shaders is the only operation that even shows up on the profiler under the mouse event handler.

screen shot 2015-11-24 at 8 47 03 am

aashish24:
thanks @jbeezley this is helpful. The culprit is planeFeature which is overdue for optimization. I will push a branch until then let's wait on merging this one.

jbeezley:
This is rebased to the current master now.  The behavior feels better with touch events, but still not great.  I think it would be useful to add custom handling when a "normal" mouse wheel is detected to generate discrete zoom events with forced +/- 1 magnitude rather than applying a scaling factor.  How exactly to detect a normal mouse wheel isn't obvious though.

manthey:
This is working just fine in Chrome, FF, and IE with my mouse wheel and my touch pad.

aashish24:
+1 LGTM

PR #475, opened 2015-12-18
Bump vgl to benefit from caching shaders

### LINKED ISSUE #471: Share shaders among tiles in a tile layer
Compiling the shaders per tile causes a large performance hit.  See the discussion in this thread: https://github.com/OpenGeoscience/geojs/pull/467#issuecomment-159061631.

jbeezley:
Fixes #471

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,7 @@
     "proj4": "=2.3.3",
     "bootstrap": "~3.3.0",
     "bootswatch": "~3.3",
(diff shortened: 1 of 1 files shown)


jbeezley:
Unfortunately, the method of caching the shaders will have to be fixed because shader programs are context specific.  The new version of vgl creates a single shader cache and doesn't work when there is more than one GL context.

Issue #474, opened 2015-12-18
Reduce memory allocations


jbeezley:
A big performance bottleneck in the current code are a result of many large garbage collections occurring during mouse events.  In the current code base, I see garbage collections that take 20-50ms occurring in about 1/2 of the map navigation invocations.  This accounts for about half of the time spent in the event handlers when no new tiles are rendered.  We should work to avoid allocation of temporary objects inside of event handlers and reduce the size of those objects that are.

Specific ideas to consider:

* Use prototypical inheritance via more modern conventions like `_.extend` rather than inside a constructor
* Use a lighter weight eventing system via some standard library
* Use native promises (with es5-promise as a shim)
* Don't generate plane features from tiles instead use a simpler class that isn't connected to the scene tree
* Keep references to "scratch spaces" on objects that are used inside methods rather than privately scoped objects that get reallocated on each call

aashish24:
> Don't generate plane features from tiles instead use a simpler class that isn't connected to the scene tree

@jbeezley how does this affect the memory allocations? Could you please elaborate?

jbeezley:
Here are some articles talking about minimizing garbage collection:

https://www.scirra.com/blog/76/how-to-write-low-garbage-real-time-javascript
http://www.html5rocks.com/en/tutorials/speed/static-mem-pools/
http://buildnewgames.com/garbage-collector-friendly-code/
http://blog.artillery.com/2012/10/browser-garbage-collection-and-framerate.html

manthey:
This has been significantly improved via PR #528.  There is more that could be done, so I'll leave it up to @jbeezley if this is enough to close this issue or not.

jbeezley:
We have a lot more that needs to be done in this direction, but I agree #528 took care of the low hanging fruit.  I'll go ahead an close this now and open a new more specific issues as we identify them.

PR #478, opened 2015-12-21
Mark tiles as fetched.

manthey:
When tiles have been fetched, do less work by not recomputing bounds for each tile.  This results in fewer GC cycles.

This helps issue #474, though it isn't any of the advised changes.

### DIFF
--- a/src/core/tileLayer.js
+++ b/src/core/tileLayer.js
@@ -900,18 +900,10 @@
       this._tileTree = {};
 
       tiles.forEach(function (tile) {
-        tile.then(function () {
-          if (tile !== this.cache.get(tile.toString())) {
-            /* If the tile has fallen out of the cache, don't draw it -- it is
-             * untracked.  This may be an indication that a larger cache should
-             * have been used. */
-            return;
-          }
-          /* Check if a tile is still desired.  Don't draw it if it isn't. */
-          var mapZoom = map.zoom(),
-              zoom = this._options.tileRounding(mapZoom),
-              bounds = this._getViewBounds();
-          if (this._canPurge(tile, bounds, zoom)) {
+        if (tile.fetched()) {
+          /* if we have already fetched the tile, don't recompute the map zoom
+           *  and view bounds, as we can use what we currently have. */
+          if (this._canPurge(tile, view, zoom)) {
             this.remove(tile);
             return;
           }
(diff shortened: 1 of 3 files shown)


src/core/tileLayer.js
```diff
@@ -920,9 +912,31 @@
 
           // mark the tile as covered
           this._setTileTree(tile);
-        }.bind(this));
+        } else {
+          tile.then(function () {
+            if (tile !== this.cache.get(tile.toString())) {
+              /* If the tile has fallen out of the cache, don't draw it -- it
```
aashish24: @manthey so if the tile is not found in the cache, hence the condition is true, it will return?

aashish24:
Mostly looks reasonable, just had a question.

src/core/tileLayer.js
manthey: Yes.  This can happen, for instance, when you quickly pan a large distance.  We fetch an image at the beginning of the pan, and by the time we fully get the image, the map has been panned to that the tile is no longer being displayed.  We no longer want to keep the tile or image in memory, so we want to discard it.

aashish24: okay, makes sense. Do you have any statistics on how much it improved the performance?  but overall LGTM :+1:

manthey: No firm numbers, but some anecdotal results.

Using the vgl renderer, we render no more often then the browser animation frame rate.  Before this change, when panning slowly and consistently, it seemed like we missed nearly half of those frames on my desk top machine, largely due to time spend in GC.  After this change, I only see a missed frame during a slow pan when we add six or more tiles concurrently.

aashish24:
+1 LGTM :+1:

aashish24:
@manthey I relaxed the requirement temporary. Let me know if the tests passes on your machine.

aashish24:
@jbeezley looks like garant is stuck. Can you restart the bot or whatever needs to be re-started?

jbeezley:
I don't know what is going on.  The build slave is up, but it isn't communicating with github.  I started a manual build http://my.cdash.org/buildSummary.php?buildid=875239.  Then I got another failure:
```
   Error when uploading file: /home/buildbot/buildbot/geojs-garant-linux-release_selenium/build/Testing/20151221-0500/Test.xml
   Error message was: Operation too slow. Less than 1 bytes/sec transferred the last 120 seconds
   Problems when submitting via HTTP
Error in read script: /home/buildbot/buildbot/geojs-garant-linux-release_selenium/common.ctest
```
I don't know if there is a problem with Kitware's network or maybe just the switch my computers are attached to, but network errors have been occurring a lot lately.  Anyway all the tests passed on this branch on the last run.

aashish24:
weird @manthey go ahead and merge it if there are no more changes.

PR #482, opened 2015-12-23
Use native arrays rather than vec3 and mat4.

manthey:
vec3 and mat4 create Float32Array data types by default.  These are lower precision than native arrays, and should only be used if the loss of precision is acceptable or the values must be passed to a WebGL context.

Rather than use vec3.fromValues, in most cases a plain array can be used.  I've added two new utility functions to create vec3 and mat4 compatible arrays that are just native arrays.

If the speed/memory benefit of an explicit array type is beneficial, these could use Float64Array instead (we would have to test if this is actually faster).

For the full benefit of this, the vgl branch of view-array-type is required.  Without that, it will still have some benefit.

I've tested this (with the vgl branch) up to zoom levels 30, and the results look smooth and flawless.

### DIFF
--- a/src/core/camera.js
+++ b/src/core/camera.js
@@ -54,13 +54,13 @@
      * The view matrix
      * @protected
      */
-    this._view = mat4.create();
+    this._view = geo.util.mat4AsArray();
 
     /**
      * The projection matrix
      * @protected
      */
-    this._proj = mat4.create();
+    this._proj = geo.util.mat4AsArray();
 
     /**
      * The projection type (one of `this.constructor.projection`)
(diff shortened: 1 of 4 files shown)


aashish24:
> I've tested this (with the vgl branch) up to zoom levels 30, and the results look smooth and flawless.

:tada: how did you get tiles upto zoom level 30? I am assuming that this is home cooked repetitive tiles. Changes looks great :+1: +1

manthey:
I'm using Brian's test tile server tweaked to provide tiles up to level 50.  The tiles are great for testing, though repetitive.  It would be cool if we had a fast fractal tile server to show this off...

aashish24:
> I'm using Brian's test tile server tweaked to provide tiles up to level 50. The tiles are great for testing, though repetitive. It would be cool if we had a fast fractal tile server to show this off...

fractal tile server? 

then changes looks good to me :+1:

PR #484, opened 2015-12-28
Increase the default tile cache size.

manthey:
Since we can now zoom to level 42, the tile cache could need to be larger.  For a 4k display where we include the lower tiles, we could have as many as 409 tiles on the display.  I've rounded up, as caching the tiles beyond that has benefits, too.

The number of tiles per level is ((ceil(width / tile_size) + 1) * (ceil(hieght / tile_size) + 1)).  For a 4k display that is actually 4096 x 2304 pixels, this works out to 17 * 10 + 9 * 6 + 5 * 4 + 3 * 3 + 2 * 2 * (max_level - 3) tiles.

We could make the default tile cache take into account the window size and max level and thus make it as small as possible while still large enough.  It would require potentially expanding or contracting it upon window resize or max level change.  If anyone thinks that is a useful feature, create an issue and we can implement it.

### DIFF
--- a/src/core/tileLayer.js
+++ b/src/core/tileLayer.js
@@ -136,7 +136,7 @@
     options = $.extend(true, {}, this.constructor.defaults, options || {});
     if (!options.cacheSize) {
       // this size should be sufficient for a 4k display
-      options.cacheSize = options.keepLower ? 400 : 200;
+      options.cacheSize = options.keepLower ? 600 : 200;
     }
     if ($.type(options.subdomains) === 'string') {
       options.subdomains = options.subdomains.split('');


aashish24:
+1


> We could make the default tile cache take into account the window size and max level and thus make it as small as possible while still large enough. It would require potentially expanding or contracting it upon window resize or max level change. If anyone thinks that is a useful feature, create an issue and we can implement it.

Can we check instead the max memory available and make this decision based on that?

aashish24:
the changes LGTM :+1:

manthey:
I don't know of any general way to check the max available memory from javascript.  You can get some information from Chrome via `performance.memory` values, though those don't seem to reflect how many images have been loaded or change the way I would expect -- Chrome uses a different amount of memory for a tab that is freshly loaded and one that has been reloaded while flushing the cache.

Issue #481, opened 2015-12-23
Handle high zoom levels better


manthey:
At high zoom levels, there are several issues:

At zoom levels 17 or 18, the html and d3 renderers fail to show tiles because the css offsets are too large for browsers to handle.  To fix this, we need to adjust the camera and tile offsets.  This needs to be done keeping in mind that we want to rotate and tilt camera views in the future.

When using the standard map projections, at zoom levels about 18 or so, the vgl renderer starts misaligning tiles.  This is due to a loss in precision from the proj4 computations.  Either we need to increase the precision used in proj4 or provide an alternative for common projections.

At zoom level 25 or so, the vgl renderer starts having issues with tile alignment even when using pixel coordinates rather than a map projection.

When panning at zoom levels of 19 or higher, the pan becomes rough.

manthey:
The loss of precision in the vgl renderer is caused by using a `vec3` to store the tile origins.  `vec3` uses a `GLMAT_ARRAY_TYPE`, which is 32-bit floating point, and thus loses precision of the origin.  I'll have a simple change to vgl that addresses this.  It doesn't fix the problems with the html and d3 renderers or with panning.

aashish24:
> The loss of precision in the vgl renderer is caused by using a vec3 to store the tile origins. vec3 uses a GLMAT_ARRAY_TYPE, which is 32-bit floating point, and thus loses precision of the origin. I'll have a simple change to vgl that addresses this. It doesn't fix the problems with the html and d3 renderers or with panning.

:snowman: I saw the change.. cool!

manthey:
With PR #482 and the appropriate vgl updates, the vgl renderer looks flawless up to zoom level 42 in Chrome.  At levels above that I start to see precision issues.

jbeezley:
Well, 2^-42 degrees latitude is 25 nanometers.

aashish24:
@manthey @jbeezley should we close this now?

manthey:
This requires the changes in VGL to be complete.

manthey:
Resolved via PR #497, #482, and others.

PR #485, opened 2015-12-28
Allow zooming in further on HTML and d3 rendered tiles.

manthey:
With this change, all renderers can zoom in to level 40.  Above level 40, there are some panning and fill artifacts in various renderers.

Along with PR #484, this finishes addressing issue #481.


aashish24:
> Above level 40, there are some panning and fill artifacts in various renderers.

@manthey what do you think causing this now?

src/d3/tileLayer.js
```diff
@@ -46,14 +48,45 @@ geo.d3.tileLayer = function () {
...
+    }
+    var x = parseInt(view.left), y = parseInt(view.top);
+    var tileCache = m_this.cache._cache;
+    $.each(canvas.selectAll('.geo-tile-layer')[0], function (idx, el) {
+      var layer = parseInt($(el).attr('data-tile-layer')),
+          scale = Math.pow(2, level - layer);
       el = m_this._getSubLayer(layer);
-      var scale = Math.pow(2, level - layer);
       el.attr('transform', 'matrix(' + [scale, 0, 0, scale, 0, 0].join() + ')');
+      var layerx = parseInt(x / Math.pow(2, level - layer)),
```
aashish24: @manthey some comments might be useful to describe the math

aashish24:
At a high level the changes looks reasonable except may be missing few comments!

manthey:
Fundamentally, the precision of doubles.  There are 52 bits of mantissa in a double.  A tile is 2^8 pixels across.  This means that without a lot of work, the best we could hope for is to zoom in to level 44.  Anything higher, and we'd expect strange floating point rounding errors to affect things.  The loss of a few bits of precision is to be expected in going through various projections and transforms.  I think getting level 40 is a pretty good outcome.

aashish24:
> I think getting level 40 is a pretty good outcome.

I agree. For geospatial data, typically its stops at 19 or 20. For biomedical data, I think 40 should be a very good number.

manthey:
@aashish24 I've added some comments and increased the code test coverage.

Level 40 zoom is pretty absurd for any real-world image.  With 256 pixel tiles, we can handle a 79,000 yotta-pixel image.

aashish24:
thanks for accommodating my request. I didn't verify the math exactly but otherwise code changes looks good to me :+1: +1

aashish24:
> Level 40 zoom is pretty absurd for any real-world image. With 256 pixel tiles, we can handle a 79,000 yotta-pixel image.

:fireworks:

Issue #472, opened 2015-12-18
Implement a tile fetch queue


jbeezley:
Currently tile ajax requests are occurring on demand.  The browser allows only a small number of requests to occur simultaneously and internally keeps a queue of outstanding requests.  When a large number of tiles are requested during user interaction, tiles that are no longer relevant for the current view bounds block newer requests from being processed.  We should implement a custom queue that checks if the tile is still in bounds before issuing a request.  If it is not, it can be safely ignored.  Further optimization could possibly adjust the order of the queue so that more important tiles are prioritized.

aashish24:
Charles and I discussed this and I believe he has some code. I request him to bring this code to geojs but I believe he has been busy or forgot about it. @manthey and I will look at this issue.

manthey:
I have an implementation of this that I will submit shortly.  I still need to add some tests and check some of the code.

aashish24:
great @manthey did you borrow code from slide atlas or this is a new implementation?

manthey:
It is a new implementation that uses javascript promises (à la jquery).

aashish24:
> It is a new implementation that uses javascript promises (à la jquery).

I see. Is there a reason, we should not look into what SlideAtlas has?

manthey:
I did look.  I decided it would be faster to write code in the same style as what else we have in geojs than it would be to adapt the code in SlideAtlas.  Specifically, we use promises to handle a tile when the image is requested.  This allows individual components to not need to know about how things are being processed.  For instance, the imageTile object doesn't need to know anything about the fetch queue.  This will be a good thing when we handle vector tiles, as that class won't need to care about the queue either.

I've written a generic Deferred object fetch queue, so it could be reused for non-tiles if appropriate, too.

aashish24:
> I did look. I decided it would be faster to write code in the same style as what else we have in geojs than it would be to adapt the code in SlideAtlas. Specifically, we use promises to handle a tile when the image is requested. This allows individual components to not need to know about how things are being processed. For instance, the imageTile object doesn't need to know anything about the fetch queue. This will be a good thing when we handle vector tiles, as that class won't need to care about the queue either.

Sure, faster is better :+1: but we should keep looking into slide-atlas code to bring it in geojs if possible and reasonable. Thanks for the update.

manthey:
Added in PR #486.

PR #486, opened 2015-12-31
Implemented a tile fetch queue.

manthey:
This is actually a general deferred object fetch queue.  It expects that no more than some number of Deferred objects should be processed at once.  For images in modern browsers, there is standard limit of six images per domain at a time.  This is the default for the fetch queue.  If more items are requested, they are further deferred and are loaded in a priority order of last-requested-first.  Items can be cancelled as well if they are no longer needed.

Also fixed the metrics that determine the load order of tiles.  There was an issue that tiles were passed and the function didn't know how to access tile.index.  This metric has been extended to have a concept of a 'bottomLevel', where anything lower than that level is lower priority.  This allows highly zoomed areas to show useful tiles first, rather than showing the absolute lowest tiles first.

This resolves issue #472.

### DIFF
--- a/src/core/fetchQueue.js
+++ b/src/core/fetchQueue.js
@@ -1,0 +1,211 @@
+(function () {
+  'use strict';
+
+  //////////////////////////////////////////////////////////////////////////////
+  /**
+   * This class implements a queue for Deferred objects.  Whenever one of the
+   * objects in the queue completes (resolved or rejected), another item in the
+   * queue is processed.  The number of concurrently processing items can be
+   * adjusted.  At this time (2015-12-29) most major browsers support 6
+   * concurrent requests from any given server, so, when using the queue for
+   * tile images, thie number of concurrent requests should be 6 * (number of
+   * subdomains serving tiles).
+   *
+   * @class
+   *
+   * @param {Object?} [options] A configuration object for the queue
+   * @param {Number} [options.size=6] The maximum number of concurrent deferred
+   *    objects.
+   * @param {Number} [options.track=600] The number of objects that are tracked
+   *    that trigger checking if any of them have been abandoned.  The fetch
+   *    queue can grow to the greater of this size and the number of items that
+   *    are still needed.  Setting this to a low number will increase
+   *    processing time, to a high number can increase memory.  Ideally, it
+   *    should reflect the number of items that are kept in memory elsewhere.
+   *    If needed is null, this is ignored.
+   * @param {function} [options.needed=null] If set, this function is passed a
+   *    Deferred object and must return a truthy value if the object is still
+   *    needed.
+   */
+  //////////////////////////////////////////////////////////////////////////////
+  geo.fetchQueue = function (options) {
+    if (!(this instanceof geo.fetchQueue)) {
+      return new geo.fetchQueue(options);
+    }
+    options = options || {};
+    this._size = options.size || 6;
+    this._track = options.track || 600;
+    this._needed = options.needed || null;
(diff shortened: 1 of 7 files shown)


aashish24:
@manthey this is great addition :boom: I am doing the review now. Did you notice any performance improvement?

manthey:
Yes, especially when moving the map rapidly.  The tiles you are looking at appear before the tiles that are hidden.

jbeezley:
The fetch seems to work well, but I'm seeing issues with the tile layer.  It is trying to load invalid tiles now, for example:  `http://otile1.mqcdn.com/tiles/1.0.0/map/4/-2/9.png`.  The same issue occurs with and without tile wrapping.  It is a bit concerning that none of the tile layer tests detect this.  We should probably add some tests that cover this scenario.

A probably unrelated issue that this exposes is that CORS errors do not seem to be triggering error handlers on the tile promises.  When using tiles served locally, I just get a normal 404 and everything behaves as I would expect, but when getting tiles from a remote CDN.  The invalid urls raise CORS errors (at least on Chrome 47.0.2526.106) in the console

```
Image from origin 'http://otile1.mqcdn.com' has been blocked from loading by Cross-Origin Resource Sharing policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8082' is therefore not allowed access. The response had HTTP status code 404.
otile1.mqcdn.com/tiles/1.0.0/map/0/-1/0.png:1 GET http://otile1.mqcdn.com/tiles/1.0.0/map/0/-1/0.png 
```

When using the GL renderer an exception is raised when trying to load the texture [here](https://github.com/OpenGeoscience/vgl/blob/master/src/texture.js#L136).  After that the map will not redraw, but tile fetches continue to show up in the network view.

In Firefox (43.0.2), the 404's coming from the CDN are not triggering CORS errors.

aashish24:
> http://otile1.mqcdn.com/tiles/1.0.0/map/4/-2/9.png

Just wanted to understand. There should not be a negative. How did we end up with a  negative index?

aashish24:
@manthey I am going line by line so please allow me till this afternoon to review the branch

manthey:
I had a typo.  In order to load tiles in the desired order, it is better to sort them before creating them, so that they get created in the order we want to load them.  When we wrap, we have the screen index (which can be negative) and the source index (which can't).  I had a typo which made the source unavailable.

src/core/fetchQueue.js
```diff
@@ -0,0 +1,211 @@
...
+   *    needed.
+   */
+  //////////////////////////////////////////////////////////////////////////////
+  geo.fetchQueue = function (options) {
+    if (!(this instanceof geo.fetchQueue)) {
+      return new geo.fetchQueue(options);
+    }
+    options = options || {};
+    this._size = options.size || 6;
+    this._track = options.track || 600;
```
aashish24: @manthey curious why did you pick up 600?

manthey: As the comments say, picking a high number wastes memory and picking a low number wastes cpu.  The number should match the size of the tile cache, which we default to either 600 or 200 depending on whether we keep lower tiles or not (see comments in the tileLayer for why).  As we default to keeping lower tiles, that means the most common default will be 600.

aashish24: thanks @manthey the code looks reasonable to me :+1: but I didn't run it.

aashish24:
@manthey looks like we need to update the branch with master.

PR #487, opened 2016-01-04
Fix an issue at high zooms with d3 and html renderers.

manthey:
At high zooms, the calculated offset failed to use the tile offset, which meant that the tile offset would still suffer from a loss of precision in CSS or SVG coordinates.  By taking into account the tile offset, this is better.


aashish24:
LGTM :+1: may be better if @jbeezley have a look at it as well.

jbeezley:
LGTM

PR #488, opened 2016-01-04
Improve the speed of the tile layer.

manthey:
When animating transitions where the center and zoom is changing, only generate one pan action per time step (we had been generating two).

Don't trigger geo events on each tile.

Remove some code that does nothing.  The `_deferredPurge` feature was never invoked, nor was an already cached tile ever purged immediately.

Cache the tileOffset calculations.

When using the vgl renderer, skip some of the code that is needed by the other renderers.

Reduce the number of times _update is called on the tile layer.  The layer and featureLayer inits were both binding geo.event.*, resulting in two calls to _update each time an event was triggered.  Add a flag to the layer init function to avoid binding events when we know we are going to bind them in the subclass.  We ALWAYS fire a pan event after a zoom event when we generate the events internally.  Therefore, on the tileLayer, ignore zoom events for _update.  It just wastes time.

### DIFF
--- a/src/core/featureLayer.js
+++ b/src/core/featureLayer.js
@@ -97,12 +97,12 @@ geo.featureLayer = function (arg) {
     }
 
     /// Call super class init
-    s_init.call(m_this);
+    s_init.call(m_this, true);
 
     /// Bind events to handlers
     m_this.geoOn(geo.event.resize, function (event) {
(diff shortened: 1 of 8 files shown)


jbeezley:
This is a nice improvement.  LGTM.

Issue #496, opened 2016-01-05
Verify non-square tiles working on tile-layer


aashish24:

manthey:
This works perfectly.  Our unit tests showed it to be working, too.

When using non power-of-two tiles, the VGL renderer is substantially slower (which is expected).

I'm closing this issue, as everything is working.

aashish24:
Awesome :tada: @manthey

PR #499, opened 2016-01-07
Better support non-square tiles in the tile example.

manthey:
Allow specifying the tile width and height in query parameters (this is not exposed in the controls).  When using non-square tiles, better compute the maximum bounds.

This was done to help verify #496.


jbeezley:
LGTM once master is merged in.

Issue #312, opened 2015-02-12
Warn user when when GL is not available


jbeezley:

jbeezley:
Even better, fallback to a different renderer when possible.

aashish24:
+1

manthey:
Someone mentioned that there was some problem with detecting WebGL reliably.  Does anyone remember the details?  This stackoverflow issue shows how others have done it: http://stackoverflow.com/questions/11871077/proper-way-to-detect-webgl-support

Should there be a function in vgl? Or should this be solely in geojs?

jbeezley:
We could probably copy the [Modernizr code](https://github.com/Modernizr/Modernizr/blob/master/feature-detects/webgl/extensions.js).  I could go either way with where to put it.  I suppose in VGL would make the most sense because it is only place where webgl API is actually touched.   I can see an API inside geojs that allows the user to probe which renderers are available in the current environment which would be more general than querying for WebGL (i.e. we could allow for loading geojs without vgl or d3).

manthey:
Added in PR #502.

PR #502, opened 2016-01-08
Check if a renderer is available.

manthey:
If not, optionally use a fallback.

This is largely intended to fallback when webGL is unavailable, but it also has the feature that d3 is no longer a hard requirement.

Add a simple phantom test for the jquery-plugin.

Also, added a reload function for the jquery-plugin, mainly to facilitate testing.  This could also help lazy-loading in some instances.

Addresses issue #312.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ script:
     - npm install
     - npm run build
     - ./node_modules/.bin/grunt docs
+    - ./node_modules/.bin/grunt copy:jqueryui
     - ./node_modules/.bin/grunt serve-test &
     - sleep 1
     - mkdir _build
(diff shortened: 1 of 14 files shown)


aashish24:
@manthey changes looks good to me but few things: 

1. We should update the document (we should start practicing this) if we have a place for it in the current document. 
2. Let's try to add these issues at the start of the sprint so that we can better track our milestones. 
3. I liked your suggestion of passing features list for the next round but I think I would say it's a low priority for now.

manthey:
@aashish24 I've update the documentation a little.

aashish24:
thansk @manthey, I appreciate it. Other than dashboard are not done yet, it LGTM :+1:

PR #505, opened 2016-01-11
Fix some tile related bugs.

manthey:
There was the possibility of getting a tile 'stuck' on the map.  This was caused by removing tiles from the tileCache when we weren't truly done with some tiles and not calling the _remove function on those tiles.  This was fixed by calling the remove function when removing tiles from the cache.

There was a problem where we could discard a tile we wanted, only to re-add it almost immediately.  This occurred when there was a comparatively large amount of tile churn compared to the tileCache size.  As new tiles were added to the tile cache, old tiles were removed from the cache.  If a new tile was the same as an old tile but was of low priority, all of the high priority new tiles would be added, pushing the old reference to low priority out of the cache.  Now, tiles are all added to the cache *before* the cache is purged of excess tiles.

Fixed a tile load order issue that could occur under high tile churn.  If a tile was already in the fetch queue, it wasn't properly added to the front of the fetch queue.

Made the tileCache automatically grow if it is smaller than the minimum size we need to hold all rendered tiles.

Add more tests to the fetchQueue, tileCache, and tileLayer.

### DIFF
--- a/src/core/fetchQueue.js
+++ b/src/core/fetchQueue.js
@@ -135,6 +135,15 @@
     };
 
     /**
+     * Get the position of a deferred object in the queue.
+     * @param {Deferred} defer Deferred object to get the position of.
+     * @returns {number} -1 if not in the queue, or the position in the queue.
+     */
+    this.get = function (defer) {
+      return $.inArray(defer, this._queue);
+    };
+
+    /**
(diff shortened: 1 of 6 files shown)


jbeezley:
Very nice!  LGTM.

PR #507, opened 2016-01-14
Don't show bad images on vgl planeFeatures.

manthey:
Only show tiles that have images in the vgl planeFeature.  We were showing tiles that had completed (including those with errors), which resulted in dead image images.  This is general to planeFeature and not limited to tiles.


jbeezley:
LGTM

PR #510, opened 2016-01-18
Use port 30100 for testing instead of 50100

### LINKED ISSUE #509: Don't use an ephemeral port for testing
We use port 50100 for testing.  This is in the ephemeral port range, and therefore can occasionally have conflicts.  We should use a port that isn't in that range (perhaps 30100 instead).

jbeezley:
Fixes #509


manthey:
LGTMWTAGAH

jbeezley:
MNTTIH

PR #515, opened 2016-01-21
Build geo.js with webpack

### LINKED ISSUE #511: Use commonjs for dependency management

jbeezley:
Fixes #511 

I'm pushing this now just as a heads up to anyone following along.  This PR will make every source file in GeoJS as node-style (sometimes called commonjs2) module with explicit declarations of both internal and external dependencies.  Once that is done, about half of the [Gruntfile](https://github.com/OpenGeoscience/geojs/blob/master/Gruntfile.js) will be replaced by a [webpack config](http://webpack.github.io/docs/configuration.html).  Future PR's will remove most of the rest of the Gruntfile that builds the examples and tests in favor of a primarily webpack-based build.

To facilitate these changes, some of the code is being reorganized.  The `init.js` files will be moved to `index.js`.  Several random utility methods that are either defined globally or directly on the `geo` namespace will be moved into `util/index.js`.  The new library will be almost completely backward compatible (except for global methods like `inherit`) by the creation of a top-level `index.js` that will reconstruct the original `geo` namespace from the reorganized modules.  Webpack will compile this entrypoint into a [UMD](https://github.com/umdjs/umd) module that will appear identical to users of the old library.

I know @manthey has a PR coming soon that will have conflicts with this.  Most of these changes are mechanical, so I can rebase when it is ready.


aashish24:
Yay :smile:

jbeezley:
There are still a few outstanding issues, but I'm ready for people to start trying this out.  The build process for the library no longer requires grunt, instead you just do a `npm run build`.  This will produce the usual `/dist/built/geo.*.js` files that should be mostly backward compatible.  The biggest change from the prior release for users of the library is moving the global `inherit` method to `geo.inherit`.  I think most of the other changes should be transparent to the vast majority of use cases.

I made a couple of changes to the built library that I'm willing to change back if anyone feels strongly about it.
* The jquery plugin is no longer included.  Having it enable itself according to the existence of jquery-ui (which itself loads asynchronously) is not easy to do when making a UMD module; it is probably possible, but I think it will cause more problems than it is worth.  I think it is best to move it to a new package on npm (e.g. geojs-jquery-plugin).
* `proj4`, `gl-matrix`, and `pnltri` are now included in `geo.min.js`.  I have yet to see an occasion where there is a reason to keep them separate.  `d3` is still in `geo.ext.min.js` and can be loaded via an external script mostly because it is technically optional, but it maybe to just include it as well.  The only other library in the external bundle is `jquery`, which will almost always be present at the global scope.
* I'm hiding several examples from the main index (as I've threatened to do for a while now), but they will still be built and hosted.

Finally, the biggest change here is in the testing.  I removed the old custom jasmine runners in favor of [karma](https://karma-runner.github.io/0.13/index.html).  In addition, I migrated all of the selenium jasmine tests into phantomjs because as it turned out none of them really needed webgl.  On travis, the tests are run via `npm run ci`, which runs the code with coverage instrumentation and submits the results to [codecov](https://codecov.io/github/OpenGeoscience/geojs?branch=commonjs). For local development and debugging, it is probably easiest to run `npm run start` which will start up the karma test server.  You can then load [http://localhost:9876/debug.html](http://localhost:9876/debug.html) in any browser to run the tests.

There are a few unresolved issues:
* `mapInteractor` interaction tests are disabled because they don't consistently pass in the test environment.  I may need to mock `setTimeout` to force the tests to run synchronously.
* Several `osmLayer` tests are disabled because they aren't working.  I still have to track this down.  It is probably related to the vgl mocking not working correctly.
* The performance collection and reporting doesn't work.  There are two problems here:
  * This line `var c = CanvasRenderingContext2D.prototype.canvas` raises an error `TypeError: 'canvas' getter called on an object that does not implement interface CanvasRenderingContext2D.` on firefox and `Uncaught TypeError: Illegal invocation` on chrome. I have no idea why it worked before but not now.
  * There endpoint that collects the performance data isn't activated.  This is pretty easy to do in karma, I just haven't done it because of the error instrumenting the global objects.
* `ctest` will no longer run the phantomjs tests nor report them to cdash.  I can either be lazy and create a test command that just runs `npm run test`, or I can use cmake to generate multiple karma configs that will execute every test file individually as it was before.  I haven't decided if it is worth it or not.

Finally, I want to eventually build the examples with webpack as well (and get rid of grunt completely), but that can probably wait for a future PR.

src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  quadFeature: require('./core/quadFeature'),
+  renderer: require('./core/renderer'),
+  sceneObject: require('./core/sceneObject'),
+  tile: require('./core/tile'),
+  tileCache: require('./core/tileCache'),
+  tileLayer: require('./core/tileLayer'),
+  timestamp: require('./core/timestamp'),
+  transform: require('./core/transform'),
+  vectorFeature: require('./core/vectorFeature'),
+  version: require('./core/version'),
```
jeffbaumes: Is there a reason the above aren't in a `core/index.js` and core can just be included in one line here? For example it seems like you do that with `util`.

jbeezley: That would make the most sense, but for backward compatibility the core modules are not named `geo.core.*`, but `geo.*`.  I would need to do the python equivalent of `from .core import *`.  I *could* change it to something like this though:

```javascript
module.exports = $.extend({},
  require('core'),
  {
    gl: require('./gl'),
    // etc
  }
}
```

It would probably be best just to stick with convention and move all of the files from core up a directory.  (Ditto with the utility methods like `geo.registerFeature`.)  I may just do that unless someone has concerns.

jeffbaumes: Moving up a directory sounds reasonable to me. @ronichoudhury convinced me recently that having code at the root level is ok for a JS library.

jbeezley:
This is now ready for review.  These are the latest changes:
* Moved all files from `src/core/` to `src/` and reorganized some of the utility methods.
* Removed the jquery ui plugin.  (I'm in the process of creating a new repo for it.)
* Updated the documentation

There are two unsolved issues which I am leaving for later.
* One minor test is currently skipped because it frequently fails on travis
* The performance instrumentation on canvas is disabled because it isn't working with webpack.  (We weren't actually collecting this data in the CI process anyway.)

aashish24:
Great progress @jbeezley 

> Removed the jquery ui plugin. (I'm in the process of creating a new repo for it.)

curious about the reasoning behind it.

aashish24:
@jbeezley let's plan for code review on Tuesday of this week. I have few questions but this is looking great!

aashish24:
LGTM +2 thanks for the explanation @jbeezley

aashish24:
Looking great :clap:

manthey:
LGTM.

Issue #442, opened 2015-10-01
Add ability to rotate the map (orthogonal to view direction) 


aashish24:

manthey:
@aashish24, @jbeezley: When we have a rotated camera (and, in the future, a tilted camera), what would you expect camera().bounds to return?

Since it is an object with left, top, right, bottom, I imagine that it would be the 2-D bounding box in the internal map coordinate system that covers the visible area.  This has the virtue that we would then fetch tiles that cover the visible area, even if we get too many of them.

Alternately, it could be the four corners of the screen at altitude 0 as a set of four points.  For compatibility, it could have left, top, right, bottom as the 2-D bounding box and the four corners under other properties (ul, ur, ll, lr)? (upperLeft, upperRight, lowerLeft, lowerRight)? (topLeft, topRight, bottomLeft, bottomRight)?

Or it could be eight points defining the frustum of the visible space...

Opinions?

jbeezley:
My original intent was to have the `left`, `right`, `bottom`, and `top` mean the maximum extent of the visible bounds in world coordinates.  My plan was always to just load everything inside these bounds and optimize later.

What these values mean in 2.5 or 3D, I have no idea.  Maybe now is the correct time to start worrying about it.

aashish24:
> Since it is an object with left, top, right, bottom, I imagine that it would be the 2-D bounding box in the internal map coordinate system that covers the visible area. This has the virtue that we would then fetch tiles that cover the visible area, even if we get too many of them.

It would be axis (or world) align bounding box if I understand what you are saying here I believe that's what @jbeezley is saying as well.

aashish24:
> Or it could be eight points defining the frustum of the visible space...

I would construct a 2d box that covers the viewable region and then I will move the planes such that they fit the geometry contained in that viewable volume (if we have any 3D). We shoudl have another api to get the frustrum.

manthey:
I have other interface questions.  I assume that if you zoom or pan, the rotation should stay the same.

What should you see if we are clamping the zoom?  Allow the same zoom levels as the unrotated map?  Or allow zooming out until opposite corners are visible on the display?

What should you see if we are clamping the X or Y bounds?

If a program sets `camera.bounds` (frequently with the output from `boundsFromZoomAndCenter`), there will need to be an option about whether to undo the rotation or not.

aashish24:
> I have other interface questions. I assume that if you zoom or pan, the rotation should stay the same.

agree

[...]

PR #516, opened 2016-01-22
Map rotation

manthey:
Maps now have a rotation property (default=0).  The tiles that are fetched are the covering rectangle of tiles.  This could be optimized so that only tiles that have a portion on the display would be fetched (though calculating that would be more costly).  Rotations work on all renderers.

Added rotation interactions.  By default ctrl+left mouse button will rotation around the center of the map by the angle that the mouse cursor moves around the center of the map.  Ctrl+wheel will rotate in 5 degree increments (strictly, one zoom level wheel movement is the same amount of wheel movement that rotates 6 degrees).

Adjust bounding box to handle clamping when rotated.  This will always let you see the entire map or image, but isn't as intuitive as it could be from a UI standpoint.  Perhaps this should be refactored in a later PR.

Fixed issues with map.bounds and map.zoomAndCenterFromBounds not returning data in the specified gcs (but always in the map gcs).

Tiles in the gl renderer will be pixel-perfect at integer zoom levels for rotations that are multiples of 90 degrees.

Removed the minX, minY, maxX, maxY parameters from the tileLayer and osmLayer classes, as they were not used.

Stopped examples from trying to load a favicon (this can be changed if we ever have a favicon, I was just tired of the 404 errors on it).

Stop allowing text selections in the examples.  When dragging maps, selecting text looks bad.

Added a "Spin to Budapest" option to the transitions example.

Updated the tiles example to allow setting the maxBounds and unitsPerPixel in the query arguments.  Added allowRotation to both the controls and the query arguments.  Changed how debug borders are drawn in the null renderer on the tiles example.

One of the added tests compares the null renderer with the d3 renderer.  Due to some wackiness in the phantom svg renderer, this only works at 0 and 180 degree rotations.

Fixed some css issues if the browser has odd line-height and font-size behavior (this affected the null renderer's tile sizes in phantom).

Removed vestigial references to map.toggle(), as it couldn't have ever worked, and any caller with enough information to use it could have toggle a layer's visibility on their own.

For transitions, deeply copy the options so that we aren't altering a data structure the caller may still be using.

Renamed defaultOpts to opts in a lot of the transition code, as it references the specific options and not the defaults.

Actually attach the geo.event.cancelAnimation and geo.event.cancelNavigation functionality in transitions.

Added tests that cover most of map.js (besides also covering most of the new code).  Added a selenium test for rotation that needs a new base-line image.

### DIFF
--- a/src/core/camera.js
+++ b/src/core/camera.js
@@ -7,11 +7,11 @@
    * camera provides a methods for converting between a map's coordinate system
    * to display pixel coordinates.
    *
-   * For the moment, all camera trasforms are assumed to be expressible as
+   * For the moment, all camera transforms are assumed to be expressible as
    * 4x4 matrices.  More general cameras may follow that break this assumption.
    *
    * The interface for the camera is relatively stable for "map-like" views,
-   * i.e. when the camera is pointing in the direction [0, 0, -1], and placed
+   * e.g. when the camera is pointing in the direction [0, 0, -1], and placed
    * above the z=0 plane.  More general view changes and events have not yet
    * been defined.
    *
@@ -27,11 +27,13 @@
    *   * {@link geo.event.camera.viewport} when the viewport changes
    *
    * By convention, protected methods do not update the internal matrix state,
-   * public methods do.  For now, there are two primary methods that are
-   * inteded to be used by external classes to mutate the internal state:
+   * public methods do.  There are a few primary methods that are intended to
+   * be used by external classes to mutate the internal state:
    *
    *   * bounds: Set the visible bounds (for initialization and zooming)
    *   * pan: Translate the camera in x/y by an offset (for panning)
+   *   * viewFromCenterSizeRotation: set the camera view based on a center
+   *        point, boundary size, and rotation angle.
    *
    * @class
    * @extends geo.object
@@ -521,22 +523,21 @@
      * @returns {object} bounds object
      */
     this._getBounds = function () {
-      var pt, bds = {};
+      var ul, ur, ll, lr, bds = {};
 
-
-      // get lower bounds
-      pt = this.displayToWorld({
-        x: 0, y: this._viewport.height
+      // get corners
+      ul = this.displayToWorld({x: 0, y: 0});
+      ur = this.displayToWorld({x: this._viewport.width, y: 0});
+      ll = this.displayToWorld({x: 0, y: this._viewport.height});
+      lr = this.displayToWorld({
+        x: this._viewport.width,
+        y: this._viewport.height
       });
-      bds.left = pt.x;
-      bds.bottom = pt.y;
 
-      // get upper bounds
-      pt = this.displayToWorld({
-        x: this._viewport.width, y: 0
-      });
-      bds.right = pt.x;
-      bds.top = pt.y;
+      bds.left = Math.min(ul.x, ur.x, ll.x, lr.x);
+      bds.bottom = Math.min(ul.y, ur.y, ll.y, lr.y);
+      bds.right = Math.max(ul.x, ur.x, ll.x, lr.x);
+      bds.top = Math.max(ul.y, ur.y, ll.y, lr.y);
 
       return bds;
     };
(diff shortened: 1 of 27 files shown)


manthey:
This resolves issue #442.

aashish24:
:fireworks: Will have a look at it later in detail

jbeezley:
This is very nice!  I uploaded the baseline and fixed the selenium test.

I did notice the widget example isn't working, but I think that is a regression in master and unrelated to these changes.  If you're seeing the same, LGTM and I'll add an issue to fix the example.

aashish24:
> This resolves issue #442.

@manthey @jbeezley I will post my review today.

src/core/camera.js
```diff
@@ -624,6 +665,9 @@
      * @param {number} zoom The zoom scale to apply
      */
     this.zoom = function (zoom) {
+      if (zoom === 1) {
+        return;
```
aashish24: @manthey why return here for zoom === 1 ?

src/core/map.js
```diff
@@ -64,6 +69,12 @@ geo.map = function (arg) {
     return new geo.map(arg);
   }
   arg = arg || {};
+
+  if (arg.node === undefined || arg.node === null) {
+    console.warn('map creation requires a node');
```
aashish24: @manthey why warning and not an error?

src/core/camera.js
```diff
@@ -633,6 +677,29 @@
...
+      }
+      axis = axis || [0, 0, -1];
+      if (!center) {
+        center = [0, 0, 0];
+      } else if (center.x !== undefined) {
+        center = [center.x || 0, center.y || 0, center.z || 0];
+      }
+      var invcenter = [-center[0], -center[1], -center[2]];
+      mat4.translate(this._view, this._view, center);
+      mat4.rotate(this._view, this._view, rotation, axis);
```
aashish24: classic stuff!

src/core/map.js
```diff
@@ -92,6 +104,9 @@ geo.map = function (arg) {
       m_queuedTransition = null,
       m_clock = null,
       m_discreteZoom = arg.discreteZoom ? true : false,
+      m_allowRotation = (typeof arg.allowRotation === 'function' ?
```
aashish24: Nice !

src/core/map.js
```diff
@@ -373,6 +415,53 @@ geo.map = function (arg) {
...
+
+    m_this.geoTrigger(geo.event.rotate, evt);
+
+    if (origin && origin.geo && origin.map) {
+      var shifted = m_this.gcsToDisplay(origin.geo);
+      m_this.pan({x: origin.map.x - shifted.x, y: origin.map.y - shifted.y});
+    } else {
+      m_this.pan({x: 0, y: 0});
+    }
+    /* Changing the rotation can change our minimum zoom */
```
aashish24: @manthey why would the min zoom will change because of rotation?

src/core/camera.js
manthey: It seems silly to do a bunch of needless multiplications when we know we don't have to.  This code isn't particularly expensive, and multiplying by 1 should never result in a precision change, but as a general habit, the less operations we do, the faster things run and the less chance for loss of precision.

src/core/map.js
manthey: In most of geojs we issue warnings and not errors.  We can change it if you like (in which case, perhaps we should revisit the other warnings, too).

src/core/map.js
manthey: Say you have a window that is 512 x 512 and are clamping the zoom.  Without rotation, you can zoom out to level = 1.  You rotate 45 degrees.  Now, the width and height of the rotated map at level 1 is sqrt(2) * 512, so you need to be able to zoom out to level = 0.5 to see the entire map.

aashish24: Ah I see, I forgot clamping.

aashish24:
LGTM :+1:

PR #522, opened 2016-01-28
Switching from jshint/jscs to ESLint

jbeezley:
ESLint make validating commonjs semantics much more pleasant so I'm making this transition now help help with the on going work in #515.  The rules included are relatively minimal for the moment so that I don't have to go through and make a bunch of manual code changes.  The majority of the code changes in this PR were done automatically using `eslint --fix` which can correct simple errors such as indentation and `'` vs. `"` issues.  I decided to get all of those changes out of the way now so we can just set the rules and move on.


aashish24:
Nice !

src/gl/vglRenderer.js
```diff
@@ -249,7 +249,7 @@ geo.registerRenderer('vgl', geo.gl.vglRenderer);
   geo.gl.vglRenderer.supported = function () {
     if (checkedWebGL === undefined) {
       /* This is extracted from what Modernizr uses. */
-      var canvas, ctx, exts;
+      var canvas, ctx, exts; // eslint-disable-line no-unused-vars
```
aashish24: I am wondering about "eslint-disable-line no-unused-vars" here.

aashish24:
Looks good to me :+1:

src/gl/vglRenderer.js
jbeezley: It's a special case.  The variables are used in the try/catch block below.  It's possible we could get rid of them, but I didn't think it worth my time figuring out.

aashish24: Thanks! I was wondering about that.  LGTM :+1:

jbeezley:
@manthey I manually fixed some of the weird issues that you brought up.

manthey:
LGTMWTIH

PR #524, opened 2016-02-02
Iss311 add more marker styles

dcjohnston:
New styles are 'point', 'wedge', and 'bar'. These styles are set using the style properties 'originStyle' and 'endStyle'. I've also made it so that for a set of vectors in a Vector feature, there can only be two marker styles, but the marker DOM elements are no longer related to the number of data points in the vector feature, which could have caused issues if a user wanted origin and end markers but only one datapoint.

I have also added an example called 'vectors' demonstrating these features.


dcjohnston:
Refactored so that each data point can get its own style in the vector feature (fixes failing selenium test). I also fixed an issue where two vector features on a single layer would create a bug because they were sharing the same set of markers. I did not use group elements on a per data point basis for performance reasons, but each vector feature gets one group element within defs where we nest the marker elements for those data points. I added a jasmine suite as well.

aashish24:
Awesome @dcjohnston 

> Refactored so that each data point can get its own style in the vector feature (fixes failing selenium test)

I am wondering if thats necessary since it will lead to performance issue specially for webgl implementation.

src/core/vectorFeature.js
```diff
@@ -74,9 +74,8 @@ geo.vectorFeature = function (arg) {
           strokeColor: 'black',
           strokeWidth: 2.0,
           strokeOpacity: 1.0,
-        // TODO: define styles for the end markers
-        // originStyle: 'none',
-        // endStyle: 'arrow',
+          originStyle: 'none',
```
aashish24: Is none documented somewhere?

dcjohnston:
I wasn't sure, but it seems like since the functions defining marker styles take, as arguments, a datum and an index for that datum in the collection for the vector Feature, i.e. tailStyle = f(d, i), each datum has to have it's own markers for head and tail (assuming neither is null).

aashish24:
@dcjohnston I am wondering if you can put together a even nicer example for showcasing vector feature. We can talk more about that offline. Also a GL implementation would be awesome.

src/core/vectorFeature.js
dcjohnston: I didn't see it, but basically 'none' does not create a DOM marker element.

aashish24:
@dcjohnston will review it soon. Any suggestion for better vector feature example? I am thinking hurricane wind velocity dataset for example: http://www.caseydietrich.com/FigureGen/20091112/01_Katrina_Ele-Wind_SELA_0100.png

aashish24:
@dcjohnston could you please update the branch?

dcjohnston:
@aashish24 Merged master into the PR. I like the hurricane data example. Is that data readily available?

aashish24:
> @aashish24 Merged master into the PR. I like the hurricane data example. Is that data readily available?

yes it is but I don;t thin it has wind velocity data.

src/core/vectorFeature.js
aashish24: sounds good @dcjohnston

src/d3/vectorFeature.js
```diff
@@ -37,8 +37,8 @@ geo.d3.vectorFeature = function (arg) {
    * @param {number} i The marker index
    */
   ////////////////////////////////////////////////////////////////////////////
-  function markerID(d, i) {
-    return m_this._d3id() + '_marker_' + i;
+  function markerID(d, i, position) {
```
aashish24: missing doc for position @dcjohnston

dcjohnston: Added that to the docs

aashish24:
@jbeezley since you are more familiar with d3 code, can you review this branch?

jbeezley:
It looks very nice.  The only major issue that I see is that the marker definitions are added on every render, but never removed.  You can see this by the list of DOM nodes matching `svg > defs > g` while interacting.

As @aashish24 mentioned, I would like to see a better example with actual data of some sort.  I would be okay with removing this example for now and adding the creation of a new example as a separate issue.  Soon, I'm going to start culling the examples that currently exist so that the better ones stand out.

src/d3/vectorFeature.js
```diff
@@ -48,32 +49,100 @@ geo.d3.vectorFeature = function (arg) {
...
+   * @param {function} originStyle The marker style for the vector head
+   * @param {function} endStyle The marker style for the vector tail
    */
   ////////////////////////////////////////////////////////////////////////////
-  function updateMarkers(data, stroke, opacity) {
+  function updateMarkers(data, stroke, opacity, originStyle, endStyle) {
+
+    var markerConfigs = {
+      'arrow': {
+        attrs: {'class': 'geo-vector', 'viewBox': '0 0 10 10', 'refX': '1', 'refY': '5', 'markerHeight': '5', 'markerWidth': '5', 'orient': 'auto'},
```
jbeezley: Could you add an additional class to these (something like `geo-vector-arrow`) to distinguish them via selectors?

testing/test-cases/phantomjs-tests/d3VectorFeature.js
```diff
@@ -0,0 +1,75 @@
...
+      };
+    })
+    .style({
+      originStyle: 'none',
+      endStyle: 'arrow'
+    })
+    .draw();
+
+    vectorLines = d3.select("#map svg").selectAll('line');
+    expect(vectorLines.size()).toBe(3);
```
jbeezley: It would be nice to make assertions here on the `marker-start` and `marker-end` attributes to ensure that the style is correct.

Also, it would be good to add an assertion on changing the style of an existing arrow so that we can get test coverage on [this line](https://github.com/OpenGeoscience/geojs/pull/524/files#diff-f2a15b34a2f5c5d3010b558bee85a73dL187).

src/d3/vectorFeature.js
dcjohnston: All set

dcjohnston: These issues are fixed. I'll add tests to confirm. It was just a minor css issue that caused the d3 selection to miss the already existing markers.

testing/test-cases/phantomjs-tests/d3VectorFeature.js
dcjohnston: I added a test that checks the style specific class (geo-marker-arrow, geo-marker-bar, etc.) for each marker.

jbeezley:
That better, but I would like to see is a test that verifies that the marker referenced by both the `marker-start` and `marker-end` attributes exist and have the correct style class.  Something like this for example:
```javascript
var marker = d3.select("#map svg").selectAll('line').attr('marker-start').match(/url\((.*)\)/);
expect(marker.length).toBe(2);
expect(d3.select(marker[1]).classed('geo-vector-arrow')).toBe(true);
```

aashish24:
@dcjohnston can you look into what @jbeezley suggested?

dcjohnston:
@aashish24 should be good. the new tests includes a sanity check on the marker id's (checking proper hear or tail prefix), and then checks each referenced marker to ensure that the correct class is present.

jbeezley:
Great thanks!  LGTM.

aashish24:
thanks @dcjohnston, appreciate it. I updated the branch with current master.

aashish24:
@dcjohnston its hard to see which test is failing. Do you mind having a closer look?

jbeezley:
It is [here](http://my.cdash.org/viewTest.php?onlyfailed&buildid=919587); I'm not sure what went wrong.  I triggered a rebuild.

aashish24:
thanks @jbeezley I am going to merge this one since everyone gave it LGTM

PR #527, opened 2016-02-04
UI Layer is implicitly moved to top whenever a new layer is added.

dcjohnston:
However, the user can still override this behavior by explicitly calling move up/down on a single layer after it has been added to the map. If this is not the desired behavior I can add a few more checks for those methods, but right now, whenever a new layer is added, the map's createLayer method will check for a UI layer and move to top. Added a test inside the layerReordering suite.

### DIFF
--- a/src/core/map.js
+++ b/src/core/map.js
@@ -509,8 +509,12 @@ geo.map = function (arg) {
       layerName, m_this, arg);
 
     if (newLayer) {
-
       m_this.addChild(newLayer);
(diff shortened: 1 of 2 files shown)


aashish24:
Make sense to me LGTM :+1: unless someone else has any objections to it. I cannot imaging a UI layer needs to be below the data layer.

jbeezley:
LGTM.  Thanks!

Issue #517, opened 2016-01-24
Share shader programs between tiles


jbeezley:
The [shaderProgram.link](https://github.com/OpenGeoscience/vgl/blob/master/src/shaderProgram.js#L235) method is causing performance problems.  See [this comment](https://github.com/OpenGeoscience/geojs/issues/471#issuecomment-169413038) for details.

manthey:
I intend to resolve this as part of issue #228, which is next on my list of tasks.

manthey:
Resolved in PR #528.

Issue #473, opened 2015-12-18
Optimize `tileLayer._update`


jbeezley:
To reduce the code complexity during development of the recent refactor, the `tileLayer._update` method is over-aggressive in recomputing the visible tile set and rerendering.  It may help to throttle the rate at which this method is called similar to what is done with mouse events in #467, but most likely it will be necessary to keep a cache of tiles in for a larger viewport and only recompute the new tileset under certain conditions (pan or zoom by a certain amount, etc.).

jbeezley:
When doing this we should keep track of the time spent in the update and ensure that no single call takes longer than 10ms.  Here are some nice resources about animation frame budget and debugging performance problems:

http://jankfree.org/
https://addyosmani.com/blog/performance-optimisation-with-timeline-profiles/

manthey:
This has been vastly improved in PR #528.

Measuring and recording times will be addressed by issue #334, so I'm closing this issue.

Issue #374, opened 2015-04-15
Add support for non-rectangular tiles


aashish24:

jbeezley:
Are you thinking hexagonal tiles?

ronichoudhury:
+:cat: for hex tiles

manthey:
If by this we just mean quads, it will be covered along with issue #228.

aashish24:
@manthey I was thinking quads, this was something needed for polar projection tiles.

manthey:
Arbitrary quadrangles are supported via PR #528.  If we ever want non-quadrangle tiles, we should make a new feature request.

Issue #228, opened 2014-10-24
Update plane feature


aashish24:
I want to call it rectFeature and provide the ability to have more than one...

manthey:
It seems like we should move in another direction and allow quadrilateral features that don't have to be rectangles, but would be easier to instantiate as rectangles.  This would make doing interesting map projections easier.  For instance, if you used a sinusoidal projection, square map tiles could be distorted into arbitrary quadrilaterals, and, at high enough zoom levels, the results would look okay.

aashish24:
> It seems like we should move in another direction and allow quadrilateral features that don't have to be rectangles, but would be easier to instantiate as rectangles. This would make doing interesting map projections easier. For instance, if you used a sinusoidal projection, square map tiles could be distorted into arbitrary quadrilaterals, and, at high enough zoom levels, the results would look okay.

+1. Also, we should have this feature create multiple instance of plane or rectangle using single geometry vs individual. Current planeFeature behaves differently than point and line.

aashish24:
@manthey based on our discussions I think we need to do the following: 

- [ ] Rename planeFeature to quadFeature
- [ ] Using geometry instancing to save memory and for performance
- [ ] Fix API so that we can pass coordinates for multiple quads

Let me know if you want to take of this.

manthey:
Currently, in vgl, there are two create*Plane functions, one for untextured (solid color) and one for textured planes.  Neither of these currently support multiple planes.  In geojs, when we create multiple points, we don't use the `vgl.utils.createPoints` function.  Rather, the bulk of the points feature is in geojs, including the shaders.

I can either add this as a `vgl.utils.createQuads` function, or with the bulk of the code in geojs `geo.gl.quadFeature`.  Either way, there will be a `quadFeature` in geojs and new shaders will be needed.  As geojs gets much more thorough testing, my inclination is to put the bulk (possibly all) of the code in geojs.

Opinions?

jbeezley:
I think that we need to define the boundary between vgl and geojs.  I believe the original intent was to use vgl as an abstraction layer around GL, but increasingly the GL code has bled into geojs.  Without clear API with documentation and testing in vgl, it is difficult know where that line should exist.

We should either commit to using a true wrapper around GL or get rid of the idea completely.  If we do want to continue with an abstraction, my preference would be to migrate to an established package like [stackgl](http://stack.gl/packages/) to do it.

aashish24:
[...]

PR #528, opened 2016-02-04
Quad feature

manthey:
Adding a quad feature which allows multiple quads each with a different image.

Moved getBuffer from various gl features to geo.util.getGeomBuffer so as to avoid duplicating code.

Updates to the latest vgl as it needs to be able to set the origin of the modelViewOrigin uniform.

If multiple quads use the same texture, that texture is only created once.  This doesn't get leveraged with an osm layer using a single image, as the tile layer creates a separate Image element per tile, but does get the benefit with general quads.

Preview images are supported for tiles, but only if those preview images are loaded -- if they are still loading, they won't be used.

Optionally cache information about quads to reduce computation.  Cache vgl textures inside images to reduce recomputing them.

Support solid-colored quads, including previewColor.

Switched the tile layer to use the quad feature.

If the tiles haven't changed, don't requeue them.  Before, on all updates, we would calculate which tiles are needed and then make sure they were queued.

Added some support for altering tile projections.  This is not really sufficient, as reprojecting tiles cause the tiles to have a non-uniform density, which means our calculation of the appropriate tiles to show at different zoom levels is incorrect.  If a projection cannot wrap and wrapping is turned on, we may try to ask for a functionally infinite set of tiles.

Do a little less work while checking the tile queue.  This greatly reduces the cpu load when no new tiles are needed or old tiles go away.

The mockVGLRenderer functions now count how often GL functions are invoked.  This is used in some of the tests.

Note that with the change for the tile layer from gl.planeFeature to gl.quadFeature, the gl.planeFeature source module is no longer covered.

Fixed a bug that could occur if a tile layer didn't include tiles for zoom level 0.


manthey:
This addresses issues #517 (use a single shader program), issue #228 (update planeFeature -- though, strictly, planeFeature hasn't been altered), issue #374 (non-square tiles -- additional work will be needed to truly generalize this).

This helps issue #473 (optimize tileLayer._update -- it does much less work if the tile set isn't changing) and issue #474 (reduce memory allocations).

jbeezley:
I added the qlQuad screenshot.  Is the new [parallelProjection failure](http://my.cdash.org/testDetails.php?test=23169316&build=908156) expected?

manthey:
No.  I was expecting everything to pass.

manthey:
When I load the parallelProjection test locally, it looks fine.

aashish24:
I like the new test images :smile: lily is my favorite. Will review it but it won't be quick since it has reasonable changes

manthey:
Here are some timing results, using the transitions example, starting with the browser having cached the tile images, but geojs not having cached the tiles except those visible in the default view, and running through the five transitions in the order shown in the example a total of three times.  In all cases I had the developer console open on the browser, which probably impacted the results some.  The frame times are the amount of time spent in `requestAnimationFrame`.

| Test                 | Total frames | Average frame time | Frame time stddev | Slow frames | Percent slow frames | Worst frame | < 0.5s | < 1s | < 2s | < 4s | < 8s | <16s | <32s | <64s | <128s | <256s | <512s | <1024s | <2048s | 
|----------------------|--------------|--------------------|-------------------|-------------|---------------------|-------------|--------|------|------|------|------|------|------|------|-------|-------|-------|--------|--------| 
| desktop-chrome-plane | 1712         | 12.69              | 9.98              | 421         | 24.59               | 84.9        | 0      | 0    | 0    | 3    | 733  | 594  | 289  | 85   | 8     | 0     | 0     | 0      | 0      | 
| desktop-chrome-quad  | 1823         | 3.31               | 2.52              | 8           | 0.44                | 20.74       | 3      | 58   | 269  | 1140 | 234  | 113  | 6    | 0    | 0     | 0     | 0     | 0      | 0      | 
| desktop-ie-quad      | 825          | 4.78               | 2.74              | 7           | 0.85                | 22.77       | 0      | 0    | 16   | 465  | 240  | 97   | 7    | 0    | 0     | 0     | 0     | 0      | 0      | 
| destop-firefox-plane | 571          | 55.25              | 120.35            | 317         | 55.52               | 1316.73     | 0      | 0    | 0    | 1    | 45   | 237  | 155  | 54   | 12    | 33    | 25    | 8      | 1      | 
| destop-firefox-quad  | 1804         | 7.5                | 3.81              | 69          | 3.82                | 41.04       | 0      | 2    | 7    | 237  | 841  | 658  | 58   | 1    | 0     | 0     | 0     | 0      | 0      | 
| laptop-chrome-plane  | 1068         | 25.55              | 25.58             | 643         | 60.21               | 157.15      | 0      | 0    | 2    | 4    | 123  | 350  | 380  | 119  | 79    | 11    | 0     | 0      | 0      | 
| laptop-chrome-quad   | 1631         | 3.89               | 5.65              | 54          | 3.31                | 62.62       | 6      | 226  | 488  | 460  | 311  | 88   | 35   | 17   | 0     | 0     | 0     | 0      | 0      | 
| laptop-firefox-plane | 233          | 130.04             | 211.72            | 223         | 95.71               | 1126.31     | 0      | 0    | 0    | 1    | 2    | 9    | 83   | 57   | 29    | 17    | 18    | 15     | 2      | 
| laptop-firefox-quad  | 1067         | 21.92              | 7.02              | 972         | 91.1                | 67.21       | 3      | 7    | 1    | 0    | 1    | 115  | 892  | 46   | 2     | 0     | 0     | 0      | 0      | 

The tests were run on my desktop and my laptop.  The desktop has an AMD FirePro W5000.  

Short summary: the quads are *vastly* faster than the planes.  Except for Firefox on my laptop, the speed is very good, with very few dropped frames.

IE reported very few slow frames, but only animated half as many frames as expected.  This means that it spent a lot of time *outside* of `requestAnimationFrame` and failed to serve frames as often as expected (it may also have limited the frame rate to 30fps rather than 60fps used in Chrome and Firefox).

aashish24:
> Short summary: the quads are vastly faster than the planes. Except for Firefox on my laptop, the speed is very good, with very few dropped frames.

This is awesome @manthey How did you capture these times?

src/core/map.js
```diff
@@ -616,13 +617,13 @@ geo.map = function (arg) {
...
     m_height = h;
 
-    m_this.camera().viewport = {width: w, height: h};
-
     reset_minimum_zoom();
     var newZoom = fix_zoom(m_zoom);
     if (newZoom !== m_zoom) {
       m_this.zoom(newZoom);
     }
+    m_this.camera().viewport = {width: w, height: h};
```
aashish24: @manthey why was this necessary?

src/core/quadFeature.js
```diff
@@ -0,0 +1,443 @@
...
+ *   (data).  The position is an Object which specifies the corners of the
+ *   quad: ll, lr, ur, ul.  At least two opposite corners must be specified.
+ *   The corners do not have to physically correspond to the order specified,
+ *   but rather correspond to that part of an image (if there is one).  If a
+ *   corner is unspecified, it will use the x coordinate from one adjacent
+ *   corner, the y coordinate from the other adjacent corner, and the average
+ *   z value of those two corners.  For instance, if ul is unspecified, it is
+ *   {x: ll.x, y: ur.y}.  Note that each quad is rendered as a pair of
+ *   triangles: (ll, lr, ul) and (ur, ul, lr).  Nothing special is done for
+ *   quads that are no convex or quads that have substantially different
```
aashish24: quads that are no convex --> I haven't see "no convex" usage in long time but it is valid. Just wondering if you have used it purposely.

src/core/quadFeature.js
```diff
@@ -0,0 +1,443 @@
...
+      quadList.forEach(function (quad, idx) {
+        for (i = 0; i < order1.length; i += 1) {
+          poly1[i].x = quad.pos[order1[i] * 3];
+          poly1[i].y = quad.pos[order1[i] * 3 + 1];
+          poly1[i].z = quad.pos[order1[i] * 3 + 2];
+          poly2[i].x = quad.pos[order2[i] * 3];
+          poly2[i].y = quad.pos[order2[i] * 3 + 1];
+          poly2[i].z = quad.pos[order2[i] * 3 + 2];
+        }
+        if (geo.util.pointInPolygon(coordinate, poly1) ||
```
aashish24: Nice!

src/core/tileLayer.js
```diff
@@ -302,6 +303,8 @@
       var o = this._origin(level);
       var map = this.map();
       point = this.displayToLevel(map.gcsToDisplay(point, null), level);
+      if (isNaN(point.x)) { point.x = 0; }
```
aashish24: @manthey  why this check? are we expecting nans?

src/gl/quadFeature.js
```diff
@@ -0,0 +1,385 @@
...
+      }
+    });
+  };
+
+  /**
+   * Render all of the color quads using a single mapper.
+   *
+   * @param renderState: the render state used for the render.
+   */
+  this._render_color_quads = function (renderState) {
```
aashish24: @manthey on the style here, why not _renderColorQuads?

aashish24:
@manthey looks great! I only have few questions. I didn't review the test code line by line but did look into the API and GL implementation and it looks great !

aashish24:
@manthey I locally merged your branch into master did npm install and grunt and got other examples running but the quad one is failing: 

http://localhost:8082/data/grid.jpg Failed to load resource: the server responded with a status of 404 (Not Found)

aashish24:
Also, in wms example,  I don't see wms layer unless I zoom in or pan. Not sure if this branch caused this may be this was the case before. (http://localhost:8082/examples/wms/)

manthey:
I captured the times by instrumenting the `requestAnimationFrame` function.  When we run transitions, we typically have two `requestAnimationFrame` calls per frame, one for rendering and one for computing where the next frame will be located.  The times are the sum of those two calls (strictly, `requestAnimationFrame` is passed the starting time of the frame callback, so the time is actually the delta between `window.performance.now()` and the callback time value.

src/core/quadFeature.js
manthey: It should read "not convex".

manthey:
Regarding http://localhost:8082/data/grid.jpg -- I think this file is only pulled down from Girder if you also build the selenium tests.

manthey:
I see the issue with the WMS example, too.  I'll look into it.

manthey:
@aashish24 The issue with the WMS example is that we *change* the url after the wms tile layer is created, and I was failing to clear a cache.  It is good that it showed up this error, but from an efficiency standpoint, the WMS example could just start with the correct url, so it doesn't bother to load a handful of different tiles before loading the tiles we want.

jbeezley:
I don't know if @aashish24 wants to chime in again, but this LGTM.

aashish24:
LGTM Ship it!

Issue #519, opened 2016-01-28
Map widgets example is not working


jbeezley:
The deployed version of the [map widgets example](http://opengeoscience.github.io/geojs/examples/widgets/) is no longer working.  I'm almost certain it was working at least on Chrome at the time of the 0.6.0 release.  I'm now using Chrome 47.0.2526.111.  I tried in current versions of every other major browser, and it is not working there either.

manthey:
It works in Chrome 48.0.2564.82 m (64-bit), but not in Chrome 48.0.2564.97 m (64-bit).

jbeezley:
All changes are breaking changes.

manthey:
Fixed in PR #530.

PR #530, opened 2016-02-11
Fix the widgets example.

manthey:
If the widget uses a position with x and y parameters, these are expected to be in the map interface gcs (latitude and longitude).  Rather, one should use top, left, right, and bottom for fixed screen locations.

Also, I noticed of Firefox that the examples had a needless vertical scrollbar, which I have hidden.

This fixes issue #519.


jbeezley:
Well that's an easy fix.  LGTM.

Issue #394, opened 2015-05-26
Z values don't plot properly on gl features


manthey:
If you have a non-zero z value, it doesn't plot where you would expect on the map projection.  There is something wrong with either the model view matrix or the projection matrix.  It seems to also engage the clipping in some instances where it shouldn't.

aashish24:
@manthey were you using point feature or line? Just wanted to make sure that I can replicate this and add a test for it once I have a fix. It may be that it exists for all the features.

manthey:
I encountered it in the polygon feature.  I also happens in the contour
feature using the same matrices.

I assume that there is an issue with one of the matrix calculations, but
didn't hunt it down.

On Mon, Jun 22, 2015 at 9:43 AM, Aashish Chaudhary <notifications@github.com
> wrote:

> @manthey <https://github.com/manthey> were you using point feature or
> line? Just wanted to make sure that I can replicate this and add a test for
> it once I have a fix. It may be that it exists for all the features.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OpenGeoscience/geojs/issues/394#issuecomment-114108670>
> .
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

manthey:
I deliberately ignore the z value in the contour feature, so if you fix
this, we should change that, too.

On Mon, Jun 22, 2015 at 9:52 AM, David Manthey <david.manthey@kitware.com>
wrote:

> I encountered it in the polygon feature.  I also happens in the contour
> feature using the same matrices.
>
> I assume that there is an issue with one of the matrix calculations, but
> didn't hunt it down.
>
> On Mon, Jun 22, 2015 at 9:43 AM, Aashish Chaudhary <
> notifications@github.com> wrote:
>
>> @manthey <https://github.com/manthey> were you using point feature or
>> line? Just wanted to make sure that I can replicate this and add a test for
>> it once I have a fix. It may be that it exists for all the features.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/OpenGeoscience/geojs/issues/394#issuecomment-114108670>
>> .
>>
>
>
>
> --
> David Manthey
> R&D Engineer
> Kitware Inc.
> (518) 881-4439
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

aashish24:
@zachmullen this would be good bug to work on.

[...]

Issue #329, opened 2015-02-25
Add tests for bundled examples


jbeezley:

manthey:
This has been begun in PR #658.

The remaining examples should be straightforward to add.

manthey:
All of the infrastructure for this is in place, with both image and javascript tests available.  Many, but not all examples have tests as of #866.

PR #531, opened 2016-02-15
Add reprojection example

manthey:
The projections are taken from spatialreference.org, with some that look best listed at the top.  The capital cities of the world are taken from geonames.org.

There is a performance issue when a select box exists with a few thousand items, so, by default, only a few dozen projections are shown.  The example can show all of the known projections by commenting out a line.

Fixed a bug in d3 points where the stroke would not appear.  This depends on the order of the style keys in the style dictionary -- if stroke came after strokeColor, the stroke might not appear.

Fixed an issue in vgl points where if a transform added a z-value, they might not be rendered.  This could be changed if we ever address issue #394.

Allow changing a widget's geographic position.

Added the ability to set the maximum bounds on a map after it is created.

When we exited the vglRenderer, we did not destroy the canvas object.  This means that if you create and destroy vgl layers repeatedly, eventually you will exhaust a browser's available gl contexts (depending on the hardware this could be as few as four gl layers or as many as 16).  Additionally, more cleanup is done to ensure that memory can be released in a timely fashion.

Also, in testing this by toggling between transforms, I've added a guard against invalid projections, so that a error is sent to the console and the state can be recovered.  Before, if a bad projection was set, it was often impossible to reset it to a valid value.

When the tile layer is a minimum level that isn't 0 and we want to keep lower levels and we aren't wrapping, then we should keep the entire lower level.  This has two benefits: reprojections can look decent just by setting a minimum level even if their bounds are odd, and padding with a non-zero minimum level still has something to show.  This has one drawback: if you have a very high minimum level (such as 5 or above), too many tiles need to be kept in memory.


aashish24:
this is awesome @manthey!

src/gl/pointFeature.js
```diff
@@ -261,10 +261,19 @@ geo.gl.pointFeature = function (arg) {
...
       position[i3 + 1] = posVal.y;
       position[i3 + 2] = posVal.z || 0;
+      nonzeroZ = nonzeroZ || position[i3 + 2];
     }
     position = geo.transform.transformCoordinates(
                   m_this.gcs(), m_this.layer().map().gcs(),
                   position, 3);
+    /* Some transforms modify the z-coordinate.  If we started with all zero z
+     * coordinates, don't modify them.  This could be changed if the
+     * z-coordinate space of the gl cube is scaled appropriately. */
```
jbeezley: Does proj4 set z-coordinates or is that something we are doing internally?  I didn't think proj4 was even aware of a third dimension.

manthey: Yes, proj4 sometimes produces non-zero z-coordinates.  This happens, for instance, in some of the projections using datums other than WGS84/NAD83, where there is a different ellipsoid.

jbeezley:
This LGTM, but I want to reprioritize #329 once it goes in.  Having more complicated examples like this is awesome, but it makes testing them more important.

PR #535, opened 2016-02-17
Add canvas renderer

law12019:
Still needs testing and opacity. Also, there is an issue with rendering order.  I believe quad features need to be sorted so low res are rendered before high res tiles.

### DIFF
--- a/src/canvas/canvasRenderer.js
+++ b/src/canvas/canvasRenderer.js
@@ -113,8 +113,9 @@ geo.canvas.canvasRenderer = function (arg) {
         m_this.context2d.clearRect(0, 0, viewport.width, viewport.height);
 
         var features = layer.features();
-        // loop ?
-        features[0]._renderImageQuads(m_this.context2d, map);
+        for (var i = 0; i < features.length; i += 1) {
+          features[i]._renderOnCanvas(m_this.context2d, map);
+        }
       });
     }
     return m_this;
--- a/src/canvas/canvasRenderer.js
+++ b/src/canvas/canvasRenderer.js
@@ -1,0 +1,168 @@
+//////////////////////////////////////////////////////////////////////////////
+/**
+ * Create a new instance of class canvasRenderer
+ *
+ * @class
+ * @extends geo.renderer
+ * @param canvas
+ * @returns {geo.gl.canvasRenderer}
+ */
+//////////////////////////////////////////////////////////////////////////////
+geo.canvas.canvasRenderer = function (arg) {
+  'use strict';
+
+  if (!(this instanceof geo.canvas.canvasRenderer)) {
+    return new geo.canvas.canvasRenderer(arg);
+  }
+  arg = arg || {};
+  geo.renderer.call(this, arg);
+
+
+  var m_this = this,
+      m_width = 0,
+      m_height = 0,
+      m_renderAnimFrameRef = null,
+      s_init = this._init,
+      s_exit = this._exit;
+
+  // ccl: is there a batter way of gettign the layer in render?
+  m_this.ccl_layer = arg.layer;
+
+  /// TODO: Move this API to the base class
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Return width of the renderer
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this.width = function () {
+    return m_width;
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Return height of the renderer
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this.height = function () {
+    return m_height;
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Get API used by the renderer
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this.api = function () {
+    return 'canvas';
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Initialize
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this._init = function () {
+    if (m_this.initialized()) {
+      return m_this;
+    }
+
+    s_init.call(m_this);
+
+    var canvas = $(document.createElement('canvas'));
+    m_this.context2d = canvas[0].getContext("2d");
+
+    canvas.attr('class', 'canvas-canvas');
+    $(m_this.layer().node().get(0)).append(canvas);
+
+    m_this.canvas(canvas);
+    /* Initialize the size of the renderer */
+    var map = m_this.layer().map(),
+        mapSize = map.size();
+    m_this._resize(0, 0, mapSize.width, mapSize.height);
+
+    return m_this;
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Handle resize event
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this._resize = function (x, y, w, h) {
+    m_width = w;
+    m_height = h;
+    m_this.canvas().attr('width', w);
+    m_this.canvas().attr('height', h);
+    m_this._render();
+
+    return m_this;
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Render
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this._render = function () {
+    if (m_renderAnimFrameRef === null) {
+      m_renderAnimFrameRef = window.requestAnimationFrame(function () {
+        m_renderAnimFrameRef = null;
+        m_this.context2d.fillStyle = "#33FFFF"
+        m_this.context2d.fillRect(300,200, 450,175);
+
+        var layer = m_this.layer(),
+            map = layer.map(),
+            camera = map.camera(),
+            viewport = camera._viewport,
+            rotation = map.rotation() || 0;
+        // Clear the canvas.
+        m_this.context2d.setTransform(1, 0, 0, 1, 0, 0);
+        m_this.context2d.clearRect(0,0,viewport.width,viewport.height);
+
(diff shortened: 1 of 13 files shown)


law12019:
@manthey PTAL

src/canvas/tileLayer.js
```diff
@@ -0,0 +1,87 @@
+geo.canvas.tileLayer = function () {
```
aashish24: @law12019 why do we need tileLayer override in canvas?

src/canvas/canvasRenderer.js
```diff
@@ -0,0 +1,166 @@
+//////////////////////////////////////////////////////////////////////////////
+/**
+ * Create a new instance of class canvasRenderer
+ *
+ * @class
+ * @extends geo.renderer
+ * @param canvas
+ * @returns {geo.gl.canvasRenderer}
```
aashish24: Is this  a typo: @returns {geo.gl.canvasRenderer} should be
@returns {geo.canvasRenderer}

src/canvas/canvasRenderer.js
```diff
@@ -0,0 +1,166 @@
...
+
+  var m_this = this,
+      m_width = 0,
+      m_height = 0,
+      m_renderAnimFrameRef = null,
+      s_init = this._init,
+      s_exit = this._exit;
+
+  // ccl: is there a batter way of gettign the layer in render?
+  m_this.ccl_layer = arg.layer;
```
aashish24: Do we need this? looks like ccl_layer is not used in this code.

src/canvas/quadFeature.js
```diff
@@ -0,0 +1,123 @@
...
+      return;
+    }
+
+    $.each(m_quads.imgQuads, function (idx, quad) {
+      var w = quad.image.width,
+          h = quad.image.height;
+      // Canvas transform is affine, so quad has to be a parallelogram
+      // Also, canvas has no way to render z.
+
+      // Tiles should be rendered from low res to high res.  Except for
```
aashish24: I guess we don't need to refer to tile here.

aashish24:
@law12019 this is a great addition! I have few comments (nothing major). Please have a look. I will checkout the branch and will run locally

aashish24:
@law12019 switching to canvas almost worked in the tile example. One issue I encounter was related to the fact the tiles didn't get updated. For me at-least on Chrome, it stuck with a low res time, so looks like a missing update or similar call.

src/canvas/tileLayer.js
manthey: @aashish24 This isn't an override, this is additive.  The way we do class inheritance in geojs doesn't really support multiple inheritance.  As such, for all of the tileLayers (d3, gl, and now canvas), we modify the instance of the tileLayer after it has been created.  This code is just doing the same thing as the d3 and gl renderers.  Ideally, we'd be using a more common javascript inheritance mechanism and it would be cleaner.

src/canvas/quadFeature.js
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Build this feature
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this._build = function () {
+    if (!m_this.position()) {
+      return;
+    }
+    m_quads = this._generateQuads();
+  };
```
manthey: In the tileLayer code, we set a z value based on the level.  We can use this z value to sort the tiles here so that they get rendered low to high:
```
    if (m_quads.imgQuads) {
        m_quads.imgQuads.sort(function (a, b) {
          return a.pos[2] - b.pos[2];
        });
      }
```

src/canvas/canvasRenderer.js
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * Render
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this._render = function () {
+    if (m_renderAnimFrameRef === null) {
+      m_renderAnimFrameRef = window.requestAnimationFrame(function () {
+        m_renderAnimFrameRef = null;
+        m_this.context2d.fillStyle = '#33FFFF';
+        m_this.context2d.fillRect(300, 200, 450, 175);
```
manthey: This looks like some leftover debug code.

src/canvas/canvasRenderer.js
```diff
@@ -0,0 +1,166 @@
...
+            map = layer.map(),
+            camera = map.camera(),
+            viewport = camera._viewport;
+        // Clear the canvas.
+        m_this.context2d.setTransform(1, 0, 0, 1, 0, 0);
+        m_this.context2d.clearRect(0, 0, viewport.width, viewport.height);
+
+        var features = layer.features();
+        // loop ?
+        features[0]._renderImageQuads(m_this.context2d, map);
```
manthey: Eventually, we'd like a renderer to support different layer types with different features.  It might be better to change this to something like:
```
for (var i = 0; i < features.length; i += 1) {
  feature[i]._renderOnCanvas(m_this.context2d, map);
}
```
and place the call to `_renderImageQuads` in the quadFeature class.

src/canvas/quadFeature.js
```diff
@@ -0,0 +1,123 @@
...
+                             (p0.x - p2.x) / w, (p0.y - p2.y) / h,
+                             p2.x, p2.y);
+
+      context2d.drawImage(quad.image, 0, 0);
+      /*
+      if (quad.opacity !== opacity) {
+        opacity = quad.opacity;
+      }
+      */
+    });
```
manthey: You should be able to just use `context2d.globalAlpha = quad.opacity;` to set the opacity (restoring the globalOpacity to the original value at the end of the function).

src/canvas/quadFeature.js
```diff
@@ -0,0 +1,123 @@
...
+    if (!m_quads.imgQuads.length) {
+      return;
+    }
+
+    $.each(m_quads.imgQuads, function (idx, quad) {
+      var w = quad.image.width,
+          h = quad.image.height;
+      // Canvas transform is affine, so quad has to be a parallelogram
+      // Also, canvas has no way to render z.
+
```
manthey: If we really want to support arbitrary quads, it can be done in canvas2d (see http://jsfiddle.net/gamealchemist/zch3gdrx/ as an example).  I don't think it is necessary at this time.

src/canvas/quadFeature.js
```diff
@@ -0,0 +1,123 @@
...
+   *
+   * @param renderState: the render state used for the render.
+   */
+  this._renderImageQuads = function (context2d, map) {
+    if (!m_quads.imgQuads.length) {
+      return;
+    }
+
+    $.each(m_quads.imgQuads, function (idx, quad) {
+      var w = quad.image.width,
```
manthey: Please add a guard before this of the form
```
if (!quad.image) {
  return;
}
```
of something similar.  When a quad is waiting for the image to load, the image attribute may not be present.

src/canvas/tileLayer.js
aashish24: Okay, let's talk about it at some point. I will look at this code in more detail.

src/canvas/canvasRenderer.js
law12019: I added this before I knew how to get the layer from the rendering.  It is not used, and has been removed.

src/canvas/quadFeature.js
law12019: I thought that would be  a solution.  I was afraid a sort would be expensive.

src/canvas/canvasRenderer.js
law12019: Fixed.

src/canvas/canvasRenderer.js
law12019: Removed

src/canvas/quadFeature.js
manthey: In `_build`.  It is only called when the quads change (added or removed), where as _renderImageQuads will get called on every render.

src/canvas/canvasRenderer.js
aashish24: thanks @law12019

src/canvas/canvasRenderer.js
aashish24: @law12019 yes, we need to make this change specially since we are planning to use the canvas renderer for another feature.

src/canvas/tileLayer.js
aashish24: @manthey and I talked and once we switch to quadFeature for d3, this code may no longer be required.

manthey:
I just ran a speed comparison of our four renderers on my desktop machine and on my laptop, both using the latest Chrome browser.

On my desktop, the vgl (webgl) and canvas are equally fast, and miss no frames during animation.  d3 (svg) only draws 71% of the frames, and null (html) draws 54% of the frames.

On my laptop, canvas is the winner, drawing nearly every frame (99.8%), vgl draws 90% of the frames, d3 24% (the worst), and html 31%.

aashish24:
> On my laptop, canvas is the winner, drawing nearly every frame (99.8%), vgl draws 90% of the frames, d3 24% (the worst), and html 31%.

this is very interesting. I guess drawing an image using canvas is fast and it makes sense to me. I am wondering about other features.

manthey:
I've pushed a few other changes I recommended for making the canvas renderer more general.  Aside from adding some tests, I think this is done.  We should create two issues to finish generalize the quad feature when we do: one to add color quads and one to handle non-parallelograms.  Neither of those is important for current work.

aashish24:
thanks @manthey I will have a look.

aashish24:
Commit 43ff236a2a0f50618a805df087cbc161ae6f9538 added most critical stuff we need right now @dcjohnston

manthey:
I've added basic testing, so this is fully ready for review.

law12019:
LGTM

PR #536, opened 2016-02-18
Version 0.8.0

jbeezley:


docs/conf.py
```diff
@@ -73,7 +73,7 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.5.0'
+version = '0.8.0'
```
aashish24: Looks like we missed it for 0.7?

package.json
```diff
@@ -49,5 +49,6 @@
     "gis",
     "webgl",
     "svg"
-  ]
+  ],
+  "main": "geo.js"
```
aashish24: curious why this is added

docs/conf.py
jbeezley: Yes, our RTD site is badly neglected.

package.json
jbeezley: It allows someone to use geojs with webpack or browserify more easily without having to provide an explicit [resolver](https://webpack.github.io/docs/configuration.html#resolve).

aashish24: ah, I see. thanks this helped.

aashish24:
LGTM +2

PR #538, opened 2016-02-18
Use relative urls to data files in examples for github.io

jbeezley:
On github.io, our `/dist` directory isn't mounted at `/`, but rather at `/geojs/`.  To make the code interoperable between our development server and the deployment, we need to use relative paths.


aashish24:
LGTM :+1:

PR #543, opened 2016-03-03
Submit notes when we run a build.

manthey:
This submits some performance tracking data for the osmLayer as a build note.  This note appears in cdash alongside of the build.

This also exposes a test in the transitions demo (add ?test=true[&renderer=(vgl|canvas|d3|html)] to the url).  This currently only logs test results to the console.  My intent is to show them in a dialog, but I haven't written that yet.

Change the jasmine test time from 5 to 15 seconds.  It would probably be better to make this configurable per test.

Changed the travis build to use -j 3 to parallelize some of the tests.


jbeezley:
It's possible (but not terribly convenient) to override the jasmine timeout.  See the [clustering tests](https://github.com/OpenGeoscience/geojs/blob/master/testing/test-cases/phantomjs-tests/clustering.js#L111-L117).

manthey:
Actually, that won't solve the problem.  In the test runner, there is a
hard limit for the entire jasmine run from open to done currently set at 5
seconds in master.  Although you can relax the time of an individual test,
the hard limit will still kill the whole test after 5 seconds.

On Thu, Mar 3, 2016 at 1:15 PM, Jonathan Beezley <notifications@github.com>
wrote:

> It's possible (but not terribly convenient) to override the jasmine
> timeout. See the clustering tests
> <https://github.com/OpenGeoscience/geojs/blob/master/testing/test-cases/phantomjs-tests/clustering.js#L111-L117>
> .
>
> —
> <https://github.com/OpenGeoscience/geojs/pull/543#issuecomment-191892719>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

jbeezley:
Is that the ctest timeout?  A hacky solution would be to just set the `TIMEOUT` property on the relevant test.  I wouldn't work too hard on it because we'll probably be moving to karma to run the tests anyway.

manthey:
No, it is the hard timeout in the [jasmine runner](https://github.com/OpenGeoscience/geojs/blob/master/testing/test-runners/run-jasmine.js#L23).

jbeezley:
I see.  Just leave it at 15 seconds then, we don't have that many async tests anyway.

jbeezley:
LGTM

PR #546, opened 2016-03-09
Don't load all tiles when keepLower is false

jbeezley:
@manthey I'm not entirely sure why the `level === minLevel` block exists.  It causes problems when trying to use WMS tiles.  We set `keepLower=false` and as a result all tiles in a vertical strip are loaded.  In our [example](http://opengeoscience.github.io/geojs/examples/wms/), the browser slows to a crawl as you zoom in.  If you have any better solutions, let me know.


manthey:
A fine change.  It exists in case the tile levels do not start at zero.  In those instances, we want to load the complete lowest level, whatever it is.

manthey:
LGTM.

manthey:
It is used in the reprojection example.

Issue #537, opened 2016-02-18
Map center resets after pan momentum


jbeezley:
To reproduce:
* drag the map quickly to generate momentum
* mouseup and keep the mouse point still
* mousewheel to generate a zoom

The map center snaps back to where it was before the panning.  If you move the mouse pointer after the pan, but before the zoom, then the behavior is correct.  This is possibly related to #260.

manthey:
Fixed in PR #547.

PR #547, opened 2016-03-10
Fix a few issues with transitions.

manthey:
There were several bugs with momentum, transitions, and spring-back.

If momentum was occurring, a mouse wheel event would take place in the wrong location.  For instance, pan a map such that it has momentum and use the mouse wheel to zoom.  The map zooms, but is pans back suddenly to the start of the momentum location.  Now, when a new mouse action occurs, momentum is canceled.

If a transition was occurring, it would "fight" with mouse interaction.  Now, when a mouse action occurs, a transition is canceled and the transitioncancel event is triggered.

Springback used hard-coded boundaries in a fixed coordinate system.  Now it uses maxBounds in the map coordinate system.

Added a spring to bounds option to the tiles example.

Moved the mockAnimationFrame functions to the testUtils.js file so it can be used in multiple tests.

Added a mockDate routine to facilitate testing momentum and springback.

This fixes issue #537.

### DIFF
--- a/examples/tiles/index.jade
+++ b/examples/tiles/index.jade
@@ -15,6 +15,9 @@ block append mainContent
     .form-group(title="Clamp vertical panning to the extent of the tiles.")
       label(for="map-clampBoundsY") Clamp Y Bounds
       input#map-clampBoundsY.mapparam(param-name="clampBoundsY", type="checkbox", placeholder="true", checked="checked")
+    .form-group(title="If clamping is turned off, spring back when the tiles are panned too far.")
+      label(for="mapinteractor-spring") Spring to Bounds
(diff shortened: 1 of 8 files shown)


brianhelba:
@manthey Can we demo this to @law12019 ? I'd be interested to get his feedback on the behavior.

jbeezley:
Yeah, that momentum/spring code really should go somewhere else.  I think it could be handled pretty easily using `map.transition` rather than producing a completely independent code path.  Perhaps a topic for a future refactor.

In any case, this change LGTM.

aashish24:
> Yeah, that momentum/spring code really should go somewhere else.

+1

aashish24:
@manthey  LGTM :+1:

PR #549, opened 2016-03-16
Momentum on zoom and fix a momentum bug.

manthey:
Momentum on zoom only applies when zoom is changed via a continuous action (the right mouse button, for instance).

Fixed a weird momentum "bounce".  Before, if you panned quickly, then stopped moving the mouse without releasing the mouse button, then release the mouse button a long time later, momentum is still applied.


jbeezley:
LGTM

Issue #542, opened 2016-03-02
Improve zoom feedback


manthey:
Animate zoom transitions by default.

manthey:
Resolved in PR #550.

PR #550, opened 2016-03-17
Added zoom animation as an option.

manthey:
Added zoom animation as an option. The speed and easing can be adjusted.

When a second transition is queued, we have a one animation frame delay when neither the current nor the new animation was being applied.  This delay has been removed.

Added some code to ensure pixel-perfect alignment in certain instances where it seemed like it didn't happen.

This resolves issue #542.

### DIFF
--- a/src/core/map.js
+++ b/src/core/map.js
@@ -338,14 +338,21 @@ geo.map = function (arg) {
   /**
    * Get/Set zoom level of the map
    *
+   * @param {number} val if undefined, return the current zoom level.
+   *    Otherwise, the new zoom level to set.
+   * @param {object} origin if present, an object with 'geo' containing the
+   *    gcs coordinates where the action started and 'map' containing the
+   *    display coordinates of the same location before the zoom is applied.
+   * @param {boolean} ignoreDiscreteZoom if true, ignore the discreteZoom
+   *    option when determining the new view.
    * @returns {Number|geo.map}
    */
   ////////////////////////////////////////////////////////////////////////////
   this.zoom = function (val, origin, ignoreDiscreteZoom) {
-    var evt, bounds;
     if (val === undefined) {
       return m_zoom;
     }
+    var evt, bounds;
 
     /* The ignoreDiscreteZoom flag is intended to allow non-integer zoom values
      * during animation. */
@@ -356,7 +363,8 @@ geo.map = function (arg) {
 
     m_zoom = val;
 
-    bounds = m_this.boundsFromZoomAndCenter(val, m_center, m_rotation, null);
+    bounds = m_this.boundsFromZoomAndCenter(val, m_center, m_rotation, null,
+                                            ignoreDiscreteZoom);
     m_this.modified();
 
     camera_bounds(bounds, m_rotation);
@@ -369,9 +377,10 @@ geo.map = function (arg) {
 
     if (origin && origin.geo && origin.map) {
       var shifted = m_this.gcsToDisplay(origin.geo);
-      m_this.pan({x: origin.map.x - shifted.x, y: origin.map.y - shifted.y});
+      m_this.pan({x: origin.map.x - shifted.x, y: origin.map.y - shifted.y},
+                 ignoreDiscreteZoom);
     } else {
-      m_this.pan({x: 0, y: 0});
+      m_this.pan({x: 0, y: 0}, ignoreDiscreteZoom);
     }
     return m_this;
   };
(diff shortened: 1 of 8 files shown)


src/core/map.js
```diff
@@ -447,7 +458,7 @@ geo.map = function (arg) {
     m_rotation = rotation;
 
     var bounds = m_this.boundsFromZoomAndCenter(
-        m_zoom, m_center, m_rotation, null);
+        m_zoom, m_center, m_rotation, null, ignoreRotationFunc);
```
aashish24: why we need to pass ignoreRotationFunc now?

aashish24:
@manthey mostly looks good to me. Can you update this branch with current master?

aashish24:
@manthey I believe https://github.com/OpenGeoscience/geojs/pull/549 is merged into this branch? If that's right, should we close #549?

manthey:
@aashish24 I was trying to keep conceptual ideas in different PRs.  #549 could be reviewed and approved first, then this PR.

src/core/map.js
manthey: During a transition, if we in discreteZoom mode and are animating both zoom and rotation, if we don't pass this flag we may end up discretizing the zoom level during the transition.

manthey:
@aashish24 I rebased on master.

src/core/map.js
```diff
@@ -1194,11 +1226,14 @@ geo.map = function (arg) {
    * Cancel any existing transition.  The transition will send a cancel event
    * at the next animation frame, but no further activity occurs.
    *
+   * @param {string} source option cause of the cancel.
```
jbeezley: Could you describe this argument a bit more.  It seems you are expecting a string like `<method name>.<action>`.  Can this method be called without a `source` argument?

manthey: Yes, `source` is optional  The description should have read 'optional' (not option).  It falls back to the a previous value (if it was set), and that falls back to an empty string.

I've used it as `<method>.<action>`, but that isn't strictly required.  I'll add to the comment that it is preferred.

jbeezley: Sounds good.  The [jsdoc syntax](http://usejsdoc.org/tags-param.html) for optional arguments is typically `@param {string} [source]`.

jbeezley:
Testing this out it works well with an ordinary mouse wheel, but it is still feels awkward with touch scrolling to me.  My [Magic Mouse](http://www.apple.com/shop/product/MLA02LL/A/magic-mouse-2) tries to simulate scrolling animation by firing many small scroll events.  These events are internally smoothed out over a period ~1 second which gives the same animated feel that we are emulating in our source.  Using this with my mouse gives the distinct feeling that the mouse is fighting with map's internal animation.

I had attempted to throttle the wheel events to prevent some of this, but I don't think it is sufficient.  For one, the fact that the wheel events continue to fire for up to 1 second after your finger stops scrolling tends to make you overshoot.  In mapbox, there is a [block of code](https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/handler/scroll_zoom.js#L82-L112) that does more intelligent throttling depending on whether it detects a mousewheel or trackpad that we may have to emulate to unify the experience between devices.

There is also a race condition that I can pretty easily trigger with my mouse that causes the map animation to stop between zoom levels even when discrete zoom is enabled.  It also seems to occasionally remember a queued transition after the animation is stopped.  The next wheel event will start the queued transition before starting the new one.  This means that the zoom will sometimes start animating in the wrong direction briefly before responding to the intended interaction.

aashish24:
> Testing this out it works well with an ordinary mouse wheel, but it is still feels awkward with touch scrolling to me. My Magic Mouse tries to simulate scrolling animation by firing many small scroll events. These events are internally smoothed out over a period ~1 second which gives the same animated feel that we are emulating in our source. Using this with my mouse gives the distinct feeling that the mouse is fighting with map's internal animation.

@manthey I am assuming that you are planning to address these? Please let me know.

manthey:
Yes.  I need to get a Magic Mouse or borrow a Mac laptop to see what events
are being fired and why it behaves oddly there.

On Fri, Mar 25, 2016 at 11:56 AM, Aashish Chaudhary <
notifications@github.com> wrote:

> Testing this out it works well with an ordinary mouse wheel, but it is
> still feels awkward with touch scrolling to me. My Magic Mouse tries to
> simulate scrolling animation by firing many small scroll events. These
> events are internally smoothed out over a period ~1 second which gives the
> same animated feel that we are emulating in our source. Using this with my
> mouse gives the distinct feeling that the mouse is fighting with map's
> internal animation.
>
> @manthey <https://github.com/manthey> I am assuming that you are planning
> to address these? Please let me know.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/OpenGeoscience/geojs/pull/550#issuecomment-201342298>
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

manthey:
@jbeezley The Mac gives zero-distance wheel events, which I wasn't expecting.  Let me know if there are still issues.

jbeezley:
Great, looks like that fixed it.  LGTM.

aashish24:
Yay :+1:

PR #551, opened 2016-03-18
Clean up attribution on map exit.

manthey:
When map.exit() is called, it didn't delete the attribution div.


jbeezley:
Good catch.  I wonder if it would be better to just do `m_this.node().clear()` to remove everything inside the map's main element.

manthey:
@jbeezley Now with a test, too.

jbeezley:
Perfect.   LGTM.

aashish24:
> @jbeezley Now with a test, too.

:grinning:

PR #556, opened 2016-04-12
Combine GeoJSON geometries into a single feature

jbeezley:
This should make the geojson reader faster and more useable.  It will create a maximum of 3 feature objects (`point`, `line`, `polygon`) for each GeoJSON file loaded.  In addition, it will now support any valid GeoJSON including rarely seen types such as `GeometryCollection`.


aashish24:
This is working nice @jbeezley LGTM +2

PR #561, opened 2016-04-20
Expose methods to define new proj4 definitions

jbeezley:
This exposes `proj4.defs` as `geo.transform.defs` and also adds a method to lookup proj4 definitions from [epsg.io](http://epsg.io/).  This can be used as follows:

```javascript
geo.transform.lookup('EPSG:26986').then(function () {
    // Now the projection can be used
    geo.transform.transformCoordinates('EPSG:26986', 'EPSG:4326', {x: 0, y: 0});
});
```

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -1,3 +1,5 @@
+var proj4 = require('proj4');
+
 //////////////////////////////////////////////////////////////////////////////
 /**
  * This purpose of this class is to provide a generic interface for computing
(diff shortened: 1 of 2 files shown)


aashish24:
this is nice @jbeezley. @mgrauer I know you don't need it but could you try it with your previous commit?

mgrauer:
I'll check it out, possibly tomorrow afternoon.

aashish24:
LGTM :+1:

Issue #544, opened 2016-03-04
Don't render area outside of bounds


manthey:
On tiles images, the size doesn't always align to tile boundaries.

manthey:
Added in PR #606.

PR #568, opened 2016-04-27
Allow overriding the tilesAtZoom function.

manthey:
Prior to this, it was assumed that there was 2^level tiles in each direction at each level.  For non-square tile source (such as slide images), this is not the case.  Unless (until?) the tilesLayer knows how many pixels of tiles are available for the maximum level, the `tilesAtZoom` function needs to change.  This allows it to be specified as an option.

This is related to issue #544, but does not complete that.


jbeezley:
LGTM

PR #570, opened 2016-05-05
Speed up transforms by caching the transform class instances.

manthey:
For simple transforms, calculate them with less object creations.


src/transform.js
```diff
@@ -24,10 +24,30 @@ var proj4 = require('proj4');
...
 //////////////////////////////////////////////////////////////////////////////
 
+var transformCache = {};
+/* Up to maxTransformCacheSize squared might be cached.  When the maximum cache
+ * size is reached, the cache is completely emptied.  Since we probably won't
+ * be rapidly switching between a large number of transforms, this is adequate
+ * simple behavior. */
+var maxTransformCacheSize = 10;
+
 var transform = function (options) {
```
aashish24: @manthey does it make sense to have a generic cache object?

manthey: Do you mean a generic cache for things besides transform?  Or something else?

In general, we use only two or four transforms, so caching those in a module-level object is fine.  Invalidating the cache when it reaches a certain size (rather than doing it via LRU) is quick and simple, and, since we won't generally have lots of transforms active at on time, it should be performant, too.

jbeezley:
LGTM

src/transform.js
aashish24: > Do you mean a generic cache for things besides transform

Yes generic cache object since we have been caching multiple things now and I feel like we could use something like that (not for this PR though).

PR #573, opened 2016-05-10
Version 0.9.1

jbeezley:


jbeezley:
This is ready to go.  I updated the documentation and exposed jquery via the more common variable `jQuery` (where bootstrap looks for it) to fix the examples.

manthey:
LGTM.

Should we load the minified geo.min.js rather than geo.js in the examples, as, since it is webpacked, the source isn't all that readable anyway?

aashish24:
LGTM :+1:

aashish24:
I prefer geo.js as it is easy to debug when I use example for testing something. I heard the same from others too but I am open either way.

jbeezley:
That's only for the main page to load jquery for bootstrap.  The other pages use the non-minified bundle.

PR #574, opened 2016-05-11
Improve performance canvas heatmap

aashish24:


aashish24:
Fixes #574 with many minor tweaks and one major tweak to get the max performance for canvas heatmap. Few other changes (binning, gaussian circle etc.) will be upcoming in separate PRs.

manthey:
During the development of this code, I did some timing tests on my desktop:

Rendering time for the heatmap in Chrome:

| # of points | Before | This PR |
| -----------:| ------:| -------:|
| 1,000 | 73 ms | 27 ms |
| 10,000 | 425 ms | 68 ms |
| 100,000 | 4373 ms | 577 ms |
| 1,000,000 | 44 s | 5.7 s |

Firefox is slower.

Of this, around 75 or 80% of the time is spent in `drawImage`.

This excludes the benefits of debouncing the rendering.

manthey:
@aashish24 Do you have any machines that exercise the big-endian code?

aashish24:
> @aashish24 Do you have any machines that exercise the big-endian code?

No I don't.

src/canvas/heatmapFeature.js
```diff
@@ -116,19 +124,55 @@ var canvas_heatmapFeature = function (arg) {
...
+      m_typedBufferData = new Uint32Array(m_typedBuffer);
+    }
+
+    m_typedBufferData[1] = 0x0a0b0c0d;
+
+    isLittleEndian = true;
+    if (m_typedBuffer[4] === 0x0a &&
+        m_typedBuffer[5] === 0x0b &&
+        m_typedBuffer[6] === 0x0c &&
+        m_typedBuffer[7] === 0x0d) {
```
manthey: When I try this out in the Chrome console, it doesn't work.  `m_typedBuffer` references an `ArrayBuffer`, which doesn't give me index reference access to the array.  I think this comparison needs to be on the `m_typedClampedBuffer`, not on `m_typedBuffer`.  What is here works, but only because we've only tried in on little-endian machines.

aashish24: I see. let me have a look at it.

aashish24: @manthey yes, it should be on clamped buffer, thanks for report this. I pushed a fix.

manthey:
@aashish24 In thinking about this further, won't this mess up the gradient on a big-endian machine because it will switch the color channels?  Specifically, in `_colorize`, the data from `gradient` is going to be in the same order as the data we want in the canvas `putImageData`, but we are reversing it.

The original (non-typed) behavior was only slower than the current code because it called `m_this.style('opacity')` for every pixel.  I just tested this with that original code without multiplying the by opacity style, and the timing is statistically identical whether we use typed arrays or the simple copy that was done originally.

aashish24:
> @aashish24 In thinking about this further, won't this mess up the gradient on a big-endian machine because it will switch the color channels? Specifically, in _colorize, the data from gradient is going to be in the same order as the data we want in the canvas putImageData, but we are reversing it.

I don't think so? Since gradient index look will give us the right value but we need to write that value that we get from gradient to the right offset. 

Too bad that jsperf is down or else we could have tested it with some simple code (performance). The original author claimed that in the general it performed faster.

aashish24:
it was reported that firefox folks has seen quite a bit of performance improvement (but I cannot confirm now since jsperf is under renovation): https://jsperf.com/canvas-pixel-manipulation/6

manthey:
@aashish24 I made a simplification that is slightly faster and no longer needs separate code paths for different endians.

src/canvas/heatmapFeature.js
```diff
@@ -124,54 +124,23 @@ var canvas_heatmapFeature = function (arg) {
...
   ////////////////////////////////////////////////////////////////////////////
-  this._colorize = function (context2d, width, height, imageData, gradient) {
-    var isLittleEndian = true, i, j, index, pixels;
-
-    // Determine whether Uint32 is little- or big-endian.
-    if (!m_typedBuffer || (m_typedBuffer.length !== imageData.data.length)) {
-      m_typedBuffer = new ArrayBuffer(imageData.data.length);
+  this._colorize = function (pixels, gradient) {
+    var grad = new Uint32Array(gradient.buffer);
+    var i, j, k, pixlen = pixels.length;
```
aashish24: can we remove the other var (to be consistent). Also I was told that fewer vars declarations is better.

manthey: I'll make it so there is just one statement.

I'd be shocked if there is a measurable performance difference between single and multiple var statements.

jbeezley: The one `var` rule comes from jslint.  I think the justification for it is preventing C programmers from misunderstanding how scoping rules work in javascript.  My preference is just to keep all of the variable declarations at the top of the function, but I don't think enforcing a one `var` rule is worthwhile.

aashish24:
> @aashish24 I made a simplification that is slightly faster and no longer needs separate code paths for different endians.

less is better. I will have a look. How did you capture the performance?

manthey:
I am using my modified example where I can load an arbitrary number of points.  I loaded 100,000 points.  I then used chrome's console to render the same frames in a predictable way: `$.each([40,30,20,10,0], function (idx, angle) {window.setTimeout(function(){heatmap.layer().map().rotation(angle * Math.PI / 180);}, 5000 + 1000 * idx); });` where I've exposed the heatmap feature as a global variable.  After starting that in the console, I used Chrome's Timeline function, and saw the time consumed for each frame (~500 ms per frame).

Note that Chrome is slower if the computer is busy doing something else, so make sure this is the only cpu-intensive process for consistent results.

src/canvas/heatmapFeature.js
aashish24: Roger that. I think most of our code is one var and declarations are on top of the function. As far as we are consistent, I am good.

manthey:
@aashish24 This now has complete test coverage.  At this point, I think you should do review and decide if it is ready for merge.

aashish24:
LGTM :+1:

Issue #575, opened 2016-05-11
SVG transforms seem to be wonky on Chrome


ronichoudhury:
If you use the d3 renderer to plot some points on a map, then you zoom out and then pan, the dots will slide from their correct positions. It almost feels like a transform is being applied twice to the dots. If you don't zoom first, the problem doesn't appear. And this happens only on Chrome, so far as I can tell.

ronichoudhury:
So, this seems to be affecting just the null renderer for the base layer, combined with the d3 renderer in a feature layer. When I instantiate my map with those settings, the bug appears both in Chrome and Firefox. :relieved:

I was getting the bug on Chrome because I don't have WebGL enabled, which causes the base layer to fall back on the null renderer, which in turn exhibits the problem.

manthey:
@ronichoudhury If you are still set up to test this, can you try the canvas and d3 renderers for the osm layer (there are four renderers: `null`, `'canvas'`, `'d3'`, and `'vgl'`).

ronichoudhury:
`'canvas'`, `'d3'`, and `'vgl'` all show the correct behavior. I confirmed again while I was at it that `null` does not.

manthey:
Thank you.  I'll hunt down the bug.

ronichoudhury:
Thanks!

manthey:
This is only a problem with the `null` renderer's transform.  I've confirmed by having multiple osm layers overlapping -- the null renderer's layer migrates around compared to the other layers.

manthey:
Fixed in PR #576.

PR #576, opened 2016-05-13
Fix an alignment issue with the null renderer.

manthey:
At non-integer zoom levels, a pan would result in the tiles being offset from where they should have been.

I'm not sure how to test this short of using image comparison.  We could make a test with tile layers using different renderers of different opacities, so that they should line up after panning or other actions.

Fixes issue #575.


ronichoudhury:
The problem seems to have gone away. LGTM!

ronichoudhury:
Will you perform a release based on this bugfix? That would solve the problem in Candela very nicely.

PR #577, opened 2016-05-13
Add controls to the heatmap example.

manthey:
Move the example data set to data.kitware.com.

Allow selecting the cities and earthquake data sets.

Fix an issue with updating the radius or blurRadius on the fly (the circle wasn't being recomputed).

Fix an issue with changing the color gradient on the fly (the gradient wasn't being recomputed).

PR #578, opened 2016-05-16
Add an option to have approximate Gaussian points instead of solid points with a blur.

manthey:
This currently depends on the changes in PR #577.  It could be separated if necessary.


src/canvas/heatmapFeature.js
```diff
@@ -96,27 +96,58 @@ var canvas_heatmapFeature = function (arg) {
...
+        /* This approximates a gaussian distribution by using a 10-step
+         * piecewise linear radial gradient.  Strictly, it should not stop at
+         * the radius, but should be attenuated further.  The scale has been
+         * selected such that the values at the radius are around 1/256th of
+         * the maximum, and therefore would not be visible using an 8-bit alpha
+         * channel for the summation.  The values for opacity were generated by
+         * the python expression:
+         *   from scipy.stats import norm
+         *   for r in [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]:
+         *     opacity = norm.pdf(r, scale=0.3) / norm.pdf(0, scale=0.3)
```
aashish24: @manthey why apply scale to it?

manthey: @aashish24, so that it has a sensible radius and fall-off.  By using a scale of 0.3, the intensity at radius=1 is approximately 1/256th of the central intensity.  This means that all of the pixels in the rendered circle have some impact on the display.  If we use a a larger number (such as the default), the circle will still have significant magnitude at the edges of our rendered area.  If we use a smaller number, then we won't use the whole area of our circle.

Simplistically, instead of using a scale, we could have scaled the radius.  The mathematics are equivalent (we could have used `norm.pdf(r/0.3) / norm.pdf(0)`).

aashish24: Thanks for the detailed explanation and comment.

aashish24:
Lgtm

PR #579, opened 2016-05-16
Add an option to bin the data before generating a heatmap

manthey:
This is faster for large data sets, as fewer images are drawn to the canvas.  It is slower for small data sets, as the binning takes some time.  The default setting 'auto' will only bin if there are fewer bins than data points.  The bin size may be auto-calculated (1/8 of the point radius) or set explicitly.  Setting it explicitly to larger values will be faster but more and more wrong.

This is currently based on other PRs.  It could be isolated if necessary.

### DIFF
--- a/src/canvas/heatmapFeature.js
+++ b/src/canvas/heatmapFeature.js
@@ -181,8 +181,140 @@ var canvas_heatmapFeature = function (arg) {
 
   ////////////////////////////////////////////////////////////////////////////
   /**
-   * Render each data point on canvas
+   * Render individual data points on the canvas.
    * @protected
+   * @param {object} context2d the canvas context to draw in.
+   * @param {object} map the parent map object.
+   * @param {Array} data the main data array.
+   * @param {number} radius the sum of radius and blurRadius.
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this._renderPoints = function (context2d, map, data, radius) {
+    var position = m_this.gcsPosition(),
+        intensityFunc = m_this.intensity(),
+        minIntensity = m_this.minIntensity(),
+        rangeIntensity = (m_this.maxIntensity() - minIntensity) || 1,
+        idx, pos, intensity;
+
+    for (idx = data.length - 1; idx >= 0; idx -= 1) {
+      pos = map.worldToDisplay(position[idx]);
+      intensity = (intensityFunc(data[idx]) - minIntensity) / rangeIntensity;
+      if (intensity <= 0) {
+        continue;
+      }
+      // Small values are not visible because globalAlpha < .01
+      // cannot be read from imageData
+      context2d.globalAlpha = intensity < 0.01 ? 0.01 : (intensity > 1 ? 1 : intensity);
+      context2d.drawImage(m_this._circle, pos.x - radius, pos.y - radius);
+    }
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Render data points on the canvas by binning.
+   * @protected
+   * @param {object} context2d the canvas context to draw in.
+   * @param {object} map the parent map object.
+   * @param {Array} data the main data array.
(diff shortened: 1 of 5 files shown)


aashish24:
Thanks @manthey I will have a look at it.

src/heatmapFeature.js
```diff
@@ -26,6 +26,11 @@ var transform = require('./transform');
  *   be computed.
  * @param {number} [updateDelay=1000] Delay in milliseconds after a zoom,
  *   rotate, or pan event before recomputing the heatmap.
+ * @param {boolean|number|'auto'} [binned='auto'] If true or a number,
```
aashish24: @manthey could we make it (binned) false by default?

src/canvas/heatmapFeature.js
```diff
@@ -181,39 +181,179 @@ var canvas_heatmapFeature = function (arg) {
...
+   * Render data points on the canvas by binning.
+   * @protected
+   * @param {object} context2d the canvas context to draw in.
+   * @param {object} map the parent map object.
+   * @param {Array} data the main data array.
+   * @param {number} radius the sum of radius and blurRadius.
+   * @param {number} binned size of the bins in pixels.
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this._renderBinnedData = function (context2d, map, data, radius, binned) {
```
aashish24: nit-pick: binned seems confusing as as we used binner in the args which makes me think that it can take boolean values as well. Can we change it binSize or something similar please?

manthey: Sure.

src/heatmapFeature.js
manthey: Yes.

I thought auto would be a better default, since based on the rest of our default values, binning probably won't occur until you have ~200,000 data points (depending on screen size), at which point the speed improvement is useful.

aashish24: Sure. I can see both sides. What happens if a user has dynamic update and let's say their number of points changes from 100,000 to 200,000 in that case now they will see a different behavior. Binning introduces / increases the visualization error (lack of a better word), that's why I thought that it should be false by default but I am with auto as well.

manthey: I agree -- either would be fine as the default.  `auto` definitely introduces a discontinuity in the visualization as the number of points crosses the threshold (resizing the window could also do this).  On the other hand, if you have a million data points, binning will be much faster.  Your call on which should be the default.

aashish24: let's go with auto :+1: we can always change it later depending on user's feedback.

src/canvas/heatmapFeature.js
```diff
@@ -181,39 +181,179 @@ var canvas_heatmapFeature = function (arg) {
...
+    var position = m_this.gcsPosition(),
+        intensityFunc = m_this.intensity(),
+        minIntensity = m_this.minIntensity(),
+        rangeIntensity = (m_this.maxIntensity() - minIntensity) || 1,
+        viewport = map.camera()._viewport,
+        bins = [],
+        rw = Math.ceil(radius / binned),
+        maxx = Math.ceil(viewport.width / binned) + rw * 2 + 2,
```
aashish24: @manthey why +2 for maxx and maxy?

manthey: We use an offset based on the first valid data point to prevent the data from shifting from bin to bin (ensuring that each data point will always be the same bin as the same scale).  We need to ensure that we calculate any bin that might affect the screen.  We have to add the radius to each side (in this case `rw`, which is the number of bins needed to make up the radius), plus something in case the bins are offset.  This probably could be `+ 1` rather than `+ 2`, and then line 255 and 259 would have to only add 1 if the offset shifts a particular direction.  It was conceptually simpler to keep it symmetric and add 2 rather than 1 and ensure that offset always was shifting in the same direction.

aashish24:
LGTM :+1:

PR #580, opened 2016-05-27
Remove grunt

jbeezley:
This PR is mostly to remove grunt from the build process.  As an alternative, a script builds all of the jade templates for the examples into html.  Webpack is used to build all of the dependency into one main bundle that all of the examples use.

A few other changes are implemented here as well:

* Updated the npm dependencies as well necessary style changes due to ESLint
* Added several npm scripts
  * examples: build examples and start server on port 8082
  * start-test/stop-test: start and stop the express test server for selenium testing (uses [forever](https://github.com/foreverjs/forever) to daemonize the process)
  * docs: build the api documentation
* Added Style checking the example sources
* Added an [EditorConfig](http://editorconfig.org/) file
* Turned on `SELENIUM_TESTS` by default because it is needed for downloading data for the examples.  I would like to use an alternative mechanism for this, but I'll do that in a different PR.


examples/common/js/jsonlint.js
```diff
@@ -0,0 +1,418 @@
+/* Jison generated parser */
```
manthey: I don't think you meant to check in this file.

jbeezley: I did actually.  As it exists on npm, jsonlint [won't work](https://github.com/zaach/jsonlint/issues/57) with webpack.  The library is small, so I copied it here and removed the offending code.

manthey:
LGTM.

aashish24:
Farewell to Grunt :clap:

Issue #540, opened 2016-02-18
Implement a d3 rendered quad feature


jbeezley:
This should replace the `d3.planeFeature` in two places.
* The rectangle selection UI element in [mapInteractor](https://github.com/OpenGeoscience/geojs/blob/09484b857b866bea4acd28a8dc90cada177b29cf/src/core/mapInteractor.js#L659)
* The tiles in [d3.tileLayer](https://github.com/OpenGeoscience/geojs/blob/09484b857b866bea4acd28a8dc90cada177b29cf/src/d3/tileLayer.js#L13)

manthey:
Added in PR #581.

Issue #548, opened 2016-03-16
Selection with rotation is wrong


manthey:
If the map is rotated, the selection box is rotated (which it shouldn't be).

manthey:
Fixed in PR #582.

Issue #541, opened 2016-02-18
Remove the plane feature


jbeezley:
It is now superseded by the quad feature.  Depends on #540.

manthey:
Removed in PR #583.

PR #581, opened 2016-06-01
	Add a d3 quadFeature

manthey:
This will render similar to the gl quadFeature, with the exception that when an image quad is not a parallelogram, the image is still drawn as one.  The gl renderer skews an image using two different transforms making the quad contain the full image, but with a transform discontinuity along one of the diagonals when the quad is not a parallelogram.  The d3 render will skew the image with a single transform, and non-parallelograms will be render somewhat incorrectly.

Initially, this was implemented with polygons for both color quads and image quads, but polygons using images have to use `defs` to define an svg pattern, and were substantially slower than using images (one virtue of using polygons is that non-parallelogram quads were rendered with the correct area, though the image was still incorrectly transformed).

The d3 tile layer has been switched to use the quad feature.  This is slightly faster than using the plane feature (perhaps a 10% speed up), but still much slower than canvas or gl.

The quad tests were dropped when the webpack refactor was done.  They have been added back in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   },
   "scripts": {
     "build": "webpack --config webpack.config.js && webpack --config external.config.js",
+    "build-examples": "webpack --config webpack-examples.config.js",
     "lint": "eslint --cache .",
     "test": "karma start karma-cov.conf.js --single-run",
     "start": "karma start karma.conf.js",
@@ -78,7 +79,7 @@
     "examples": "webpack-dev-server --config webpack-examples.config.js --port 8082 --content-base dist/",
     "start-test": "node examples/build.js; forever start ./testing/test-runners/server.js",
     "stop-test": "forever stop ./testing/test-runners/server.js",
-    "docs": "jsdoc --pedantic -d dist/apidocs -r src"
+    "docs": "jsdoc --pedantic -d dist/apidocs -r src -c jsdoc.conf.json"
   },
   "keywords": [
     "map",
--- a/src/d3/d3Renderer.js
+++ b/src/d3/d3Renderer.js
@@ -478,7 +478,6 @@ var d3Renderer = function (arg) {
     m_features[arg.id] = {
       data: arg.data,
       index: arg.dataIndex,
-      defs: arg.defs,
       style: arg.style,
       attributes: arg.attributes,
       classes: arg.classes,
@@ -510,24 +509,12 @@ var d3Renderer = function (arg) {
         attributes = m_features[id].attributes,
         classes = m_features[id].classes,
         append = m_features[id].append,
-        defs = m_features[id].defs,
         selection = m_this.select(id, parentId).data(data, index);
     selection.enter().append(append);
     selection.exit().remove();
     setAttrs(selection, attributes);
     selection.attr('class', classes.concat([id]).join(' '));
     setStyles(selection, style);
-    if (defs) {
-      selection = m_defs.selectAll('.' + id).data(defs.data, index);
-      var entries = selection.enter().append(defs.append);
-      if (defs.enter) {
-        defs.enter(entries);
-      }
-      selection.exit().remove();
-      setAttrs(selection, defs.attributes);
-      selection.attr('class', (defs.classes || []).concat([id]).join(' '));
-      setStyles(selection, defs.style);
-    }
     return m_this;
   };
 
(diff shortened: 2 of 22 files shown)


manthey:
This resolves issue #540 and is needed by issues #541 and #548.

package.json
```diff
@@ -70,6 +70,7 @@
   },
   "scripts": {
     "build": "webpack --config webpack.config.js && webpack --config external.config.js",
+    "build-examples": "webpack --config webpack-examples.config.js",
```
aashish24: +1

tests/cases/osmLayer.js
```diff
@@ -1,6 +1,9 @@
 // Test geo.core.osmLayer
 var geo = require('../test-utils').geo;
 var $ = require('jquery');
+var mockAnimationFrame = require('../test-utils').mockAnimationFrame;
```
aashish24: Why we have to add mockAnimationFrame now?

src/d3/d3Renderer.js
```diff
@@ -207,35 +202,24 @@ var d3Renderer = function (arg) {
...
-        Math.pow(lowerRight.y - upperLeft.y, 2) +
-        Math.pow(lowerRight.x - upperLeft.x, 2)) / m_diagonal;
-      // calculate the translation
-      rotation = map.rotation();
-      rx = -m_width / 2;
-      ry = -m_height / 2;
-      dx = scale * rx + center.x;
-      dy = scale * ry + center.y;
-    }
+    scale = Math.sqrt(
```
aashish24: So now we compute scale everytime vs earlier it was stored / cached in the html?

src/tileLayer.js
```diff
@@ -728,8 +728,8 @@ module.exports = (function () {
       container.append(tile.image);
       container.css({
         position: 'absolute',
-        left: (bounds.left - parseInt(div.attr('offsetx') || 0, 10)) + 'px',
-        top: (bounds.top - parseInt(div.attr('offsety') || 0, 10)) + 'px'
+        left: (bounds.left - parseFloat(div.attr('offsetx') || 0)) + 'px',
```
aashish24: what is the reason behind switching from parseInt to parseFloat?

aashish24:
Not directly related to this branch: I am having issues with data: 

```
[ 63%] Downloading grid.jpg
-- Data is here: https://data.kitware.com/api/v1/file/560a89518d777f7bfaadd3dd/download
CMake Error at Girder_FetchScripts/fetch_26e86857de2e8c4c4fc808dfd6b12e95_grid.jpg.cmake:7 (message):
  Error downloading grid.jpg: "Unsupported protocol"
```

jbeezley:
You probably need to compile cmake with SSL support.

aashish24:
thanks @jbeezley didn't have this issue before. Is it because we switched to https on data recently?

aashish24:
that was indeed the problem, my cmake was built without openssl.

src/d3/d3Renderer.js
manthey: Before, we computed a scale for each layer of tiles and modified it upon movement.  Now, we only calculate one transform (rather than one per layer).  This is done initially and on movement (not every render).  So we actually calculate less transforms than we did.

src/tileLayer.js
manthey: In thinking about this, perhaps I should have forced the svg layer to use `int`s rather than let the html layer use `float`s.  Fundamentally, I was doing nothing to round the svg layer's transform to integers, and the test code that compares different tile renderings to see if they line up was failing.

I was considering this last night, and I should change this back to ints and make the svg layer match -- otherwise the html renderer shows small gaps between tiles when we don't show lower tile levels.

tests/cases/osmLayer.js
manthey: I moved the rendering into a requestAnimationFrame callback for d3 features. Before, as we added tiles, we would render them as they came in,  This means that when you panned and we loaded multiple tiles, the tiles would render many times in that animation frame, which caused a huge performance lag for that moment.  By using an animation frame, we render all of the new tiles once, and it limits the performance hit.  It still takes longer than I'd like to add many svg elements at once, but it is swifter than it was.

src/d3/d3Renderer.js
aashish24: okay, thanks.

tests/cases/osmLayer.js
aashish24: Makes sense to me. thanks!

aashish24:
@manthey LGTM :+1: @jbeezley do you bless this change too?

jbeezley:
Yup, LGTM.

PR #584, opened 2016-06-06
The performance value reported from Travis to cdash via notes wasn't calculating time properly.

manthey:
Our test environment does something peculiar to the time values, such that the time stamp of animation frames is not directly comparable to either the javascript date obtained by `new Date()` (it is off by a few seconds), nor with `window.performance.now()` (which reports values that are probably time-from-start in milliseconds rather than epoch milliseconds).

Also, we only reported timings for mocked vgl osm tiles.  This is now reporting timings for canvas and svg tiles, too.


aashish24:
@manthey I have some questions on this branch. I will stop by after lunch to discuss.

aashish24:
LGTM :+1:

Issue #470, opened 2015-12-17
GL features and layers don't redraw when calling `draw`


jbeezley:
You have to call draw on the map object to cause features to update on the screen.

Basic test:  http://jsbin.com/kutoku/3/edit?js,output

manthey:
Fixed by PR #585.

PR #585, opened 2016-06-06
GL and canvas features don't redraw when their draw methods are called.

manthey:
Rather than calling draw on a layer or feature, you have to call draw on the whole map.  This fixes the issue.  Resolves issue #470.

Fix the hurricanes example's line highlight.

Fix the picking example (though keep it disabled).


aashish24:
thanks @manthey, i will have a closer look for review.

aashish24:
@manthey LGTM 👍

PR #586, opened 2016-06-07
When running the examples, allow overriding the host and port.

manthey:
This allows changing the host and port via a command like 'HOST=0.0.0.0 PORT=8080 npm run examples'.


aashish24:
LGTM :+1:

Issue #588, opened 2016-06-08
Fix unreliable quadFeature test


jbeezley:
[This condition](https://github.com/OpenGeoscience/geojs/blob/613107888d7808ff569b526de4922378ff2c4778/tests/cases/quadFeature.js#L562-L563) frequently fails to resolve on travis.
```
JS 2.1.1 (Linux 0.0.0) geo.quadFeature geo.canvas.quadFeature wait for next render canvas C FAILED
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
PhantomJS 2.1.1 (Linux 0.0.0): Executed 372 of 603 (1 FAILED) (0 secs / 17.424 secs)
```

manthey:
Fixed via PR #589.

PR #589, opened 2016-06-08
The canvas.quadFeature test sometimes fails on travis.

manthey:
This was caused by the canvas renderer being called twice in the time interval for checking if it had been called.  By switching some of the tests on call counts from === to >=, this will still pass.  I've checked several of the tests, not just the one that failed most often.

This fixes issue #588.


jbeezley:
LGTM

PR #590, opened 2016-06-09
Create layers based on a list of features.

manthey:
Currently, if you want to use a feature, you need to know what renderers in works in and manually check if those renderers are available on the current platform.  If the renderer isn't available, we fall back to the null renderer, which is only useful for the tile layers.

With this change, you can ask for a new layer based on the features you want.  If there are any renderers available that support all of the listed features, that renderer will be used.  Furthermore, there is now an inherent preference in renderers in the order vgl -> canvas -> d3 -> null, so we'll fall back to a better renderer than the null renderer in many instances.

Features can have capabilities.  For instance, the quad feature is only fully implemented in the vgl renderer.  Image quads are mostly implemented in the d3 and canvas renderers (when the quad is a parallelogram).  Color quads are implemented in the d3 renderer.  If you ask for a layer that supports {features: ['quad']}, you could get any of the vgl, d3, or canvas renderers (based on the support of that renderer).  If you ask for {features: ['quad.clr']}, you will only get either the vgl or d3 renderer.  If you ask for {features: ['quad.img-full']}, only vgl will do.

Many examples have been updated to use this.  This means that some examples will now work on more platforms (since they would only work if you have vgl, but now will work in other cases).


aashish24:
🎉 will complete my review tomorrow but I liked what I reviewed so far

examples/choropleth/main.js
```diff
@@ -31,11 +31,13 @@ $(function () {
...
-  // Create a gl feature layer
-  var vglLayer = map.createLayer(
+  // Create a feature layer.  We could either ask for a layer via a specific
+  // render {renderer: 'vgl'} or for a layer that supports our feature
+  // {features: ['choropleth']}.
+  var choroplethLayer = map.createLayer(
     'feature',
     {
-      renderer: 'vgl'
+      features: ['choropleth']
```
aashish24: +1

examples/hurricanes/main.js
```diff
@@ -369,8 +369,8 @@ $(function () {
     }
   );
 
-  // Create a feature layer to draw on
-  layer = map.createLayer('feature');
+  // Create a feature layer to draw on.
+  layer = map.createLayer('feature', {features: ['line.multicolor']});
```
aashish24: what is line.multicolor refers to?

aashish24: I see now that you have added options at the time of registration. Interesting idea but now we have to make sure we are consistent about it and document it as well.

src/canvas/quadFeature.js
```diff
@@ -143,5 +143,5 @@ var canvas_quadFeature = function (arg) {
 inherit(canvas_quadFeature, quadFeature);
 
 // Now register it
-registerFeature('canvas', 'quad', canvas_quadFeature);
+registerFeature('canvas', 'quad', canvas_quadFeature, {clr: false, img: true, 'img-full': false});
```
aashish24: could we change clr to color please? clr sometime could be confused with clear.

examples/hurricanes/main.js
manthey: Our gl lines can vary color from point to point.  The d3 lines can't.

We probably should document the feature capabilities in the base feature.

src/canvas/quadFeature.js
aashish24: @manthey how do control and maintain this vocabulary? May we should make it part of public interface of baseclass (keys) and then the derived classes can just refer to it like this: 

registerFeature('canvas', 'quad', canvas_quadFeature, {quadFeature.color: false, quadFeature.img: true, quadFeature.img-full: false});

thoughts?

aashish24:
Mostly looks great :yum:  to me but I have one suggestion for discussions.

src/canvas/quadFeature.js
manthey: Sounds reasonable.  The base class can document what the keys mean, too.  I'll make those changes.

aashish24: great, thanks for the consideration @manthey

aashish24:
LGTM :tada:

aashish24:
@manthey should we go ahead and merge it?

jbeezley:
LGTM

PR #594, opened 2016-06-16
Improve polygons

manthey:
Add a polygon example.

Speed up gl polygons a large amount.  Before, the set of polygons in the example was taking between 240 and 350 ms to regenerate.  The variability was largely caused by garbage collection.  There were many slow points in our processing (earcut only accounts for 35 ms of this time).  Now, much less memory is allocated and deallocated, and some maps have been unrolled to plain functions.  The full generation time is around 60 to 75 ms, with much lower memory turn-over.  Further, if only the style changes, the geometry doesn't need to be recomputed, so just recalculating styles takes 11 ms or so.

Sped up the `pointInPolygon` function, dropping its time (using the example data) from 10 ms to 2 ms or so.

When mousing over a polygon, these changes mean that it feels 'real-time' instead of very laggy.

Polygons had to be defined as {outer: [(points)], inner: [[(points)], ...]}, with inner being optional.  Now, they can also be defined as [(points)], skipping the container structure, which is the same as {outer: [(points)]}.

Improved the documentation on polygonFeature.

Note that this is dependent on PR #590, as the example uses that.  It could be separated if necessary.


manthey:
Some performance comparisons (all times in milliseconds) using world-wide 10 arcminute bathymetric data with 5779 polygons with lots of holes that convert to 2,209,058 triangles (6,627,174 vertices) from a 51 Mb json file.

| Step | Before | After | Comments |
|-------|---------:|-------:|----------------|
| getCoordinates | 1305 | 160 | This is getting called twice based on how the example is run.  This is the slowest of the two times |
| createGlPolygons -- first run | 21405 | 16017 | Earcut takes around 14000 ms.  The rest is geometry and style calculations |
| createGlPolygons -- highlighting | 27280 | 5.39 | The before timing started out taking around 19500 ms, but gradually increased due to garbage collection |
| pointSearch | 101 | 2.26 | |

The speed improvements come from changing map functions to loops, only calculating triangles as needed, calculating bounding boxes of polygons, and taking advantage of uniform values.

aashish24:
> Some performance comparisons

this is cool. For the point search, the speed up is because of the use of bbox?

src/gl/polygonFeature.js
```diff
@@ -82,122 +85,193 @@ var gl_polygonFeature = function (arg) {
...
+      uniform = uniformPolyFunc(item, itemIndex);
+      if (uniform) {
+        if (fillColorVal === undefined) {
+          color = fillColorFunc(vertices[0], 0, item, itemIndex);
+        }
+        if (fillOpacityVal === undefined) {
+          opacity = fillOpacityFunc(vertices[0], 0, item, itemIndex);
+        }
+      }
+      if (uniform && onlyStyle && items[k].uniform && items[k].color && color.r === items[k].color.r && color.g === items[k].color.g && color.b === items[k].color.b && opacity === items[k].opacity) {
```
aashish24: nit pick: can we wrap this long line/

src/polygonFeature.js
```diff
@@ -8,6 +8,27 @@ var feature = require('./feature');
  *
  * @class geo.polygonFeature
  * @extends geo.feature
+ * @param {Object} arg Options object
+ * @param {Object|Function} [arg.position] Position of the data.  Default is
+ *   (data).  The position is an Object which specifies the location of the
+ *   data in geo-spatial context.
```
aashish24: minor suggestion: geo-spatial context is not a must.

src/polygonFeature.js
```diff
@@ -67,7 +92,9 @@ var polygonFeature = function (arg) {
   /**
    * Get the internal coordinates whenever the data changes.  For now, we do
    * the computation in world coordinates, but we will need to work in GCS
-   * for other projections.
+   * for other projections.  Also compute the extents of the outside of each
+   * polygon for faster chcking if points are in the polygon.
```
aashish24: @manthey you may want to run a spell checker on code comments ``chcking --> checking``

src/util/init.js
```diff
@@ -38,32 +38,40 @@
...
           inside = !inside;
         }
-      });
+      }
 
-      (inner || []).forEach(function (hole) {
-        inside = inside && !geo.util.pointInPolygon(point, hole);
-      });
+      if (inner && inside) {
+        (inner || []).forEach(function (hole) {
```
aashish24: since for is much faster than forEach why not use for here as well?

manthey:
It is also changing from the forEach to a loop, and not bothering to check
the holes if the point wasn't inside the outer polygon.

On Mon, Jun 20, 2016 at 10:45 AM, Aashish Chaudhary <
notifications@github.com> wrote:

> Some performance comparisons
>
> this is cool. For the point search, the speed up is because of the use of
> bbox?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/OpenGeoscience/geojs/pull/594#issuecomment-227163602>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AIX_R5djQMfTNe5QvO24yTDIHBnWCTT-ks5qNqeVgaJpZM4I3hEE>
> .
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

src/polygonFeature.js
manthey: We describe position elsewhere with this same text elsewhere, too.  I'll
change it.

On Sat, Jun 25, 2016 at 12:20 AM, Aashish Chaudhary <
notifications@github.com> wrote:

> In src/polygonFeature.js
> <https://github.com/OpenGeoscience/geojs/pull/594#discussion_r68483789>:
>
> > @@ -8,6 +8,27 @@ var feature = require('./feature');
> >   *
> >   * @class geo.polygonFeature
> >   * @extends geo.feature
> > + * @param {Object} arg Options object
> > + * @param {Object|Function} [arg.position] Position of the data.  Default is
> > + *   (data).  The position is an Object which specifies the location of the
> > + *   data in geo-spatial context.
>
> minor suggestion: geo-spatial context is not a must.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/OpenGeoscience/geojs/pull/594/files/67fa92de4928f8d849ff0abbbb2b8dabbe14eb1e#r68483789>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AIX_Rx0mnlk2llX-t1ZSXeDCfbAgp4Bxks5qPKyOgaJpZM4I3hEE>
> .
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

src/util/init.js
manthey: There usually aren't that many holes, so the speed increase in this change
is minor, and I didn't bother to make it.

On Sat, Jun 25, 2016 at 12:53 AM, Aashish Chaudhary <
notifications@github.com> wrote:

> In src/util/init.js
> <https://github.com/OpenGeoscience/geojs/pull/594#discussion_r68484109>:
>
> >
> > -      (inner || []).forEach(function (hole) {
> > -        inside = inside && !geo.util.pointInPolygon(point, hole);
> > -      });
> > +      if (inner && inside) {
> > +        (inner || []).forEach(function (hole) {
>
> since for is much faster than forEach why not use for here as well?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/OpenGeoscience/geojs/pull/594/files/67fa92de4928f8d849ff0abbbb2b8dabbe14eb1e#r68484109>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AIX_RzDWzZTRVoza9g0EWmiY-yWFp9Ugks5qPLRDgaJpZM4I3hEE>
> .
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

aashish24: thanks. I would have changed them all but I am fine with leaving it as it is.

aashish24:
LGTM :+1:

PR #595, opened 2016-06-28
Wms styling

dorukozturk:
Added an example to style wms raster tiles on the fly via the sld file. We are including visualization specs to the url as a parameter called "SLD_BODY". The UI listens for any change on the forms and updates the map by doing the following steps:

1) Delete the current layer
2) Regenerate the URL using the new parameters
3) Send a GET request with the new visualization parameters

@aashish24 @jbeezley @manthey


manthey:
I *think* all of our other example thumb.jpg files are 800x600 pixels in size, so for consistency, we probably want to make this one the same.

examples/sld/main.js
```diff
@@ -0,0 +1,874 @@
...
+        ],
+    7: ['#ffffcc', '#d9f0a3', '#addd8e', '#78c679', '#41ab5d',
+            '#238443', '#005a32'
+        ],
+    8: ['#ffffe5', '#f7fcb9', '#d9f0a3', '#addd8e', '#78c679',
+            '#41ab5d', '#238443', '#005a32'
+        ],
+    9: ['#ffffe5', '#f7fcb9', '#d9f0a3', '#addd8e', '#78c679',
+            '#41ab5d', '#238443', '#006837', '#004529'
+        ]
```
manthey: We are limiting the color count to 8.  Is there some virtue to keeping the entries for >= 9 in the `colorbrewer` object?

jbeezley: I would actually prefer installing this through npm anyway.

examples/sld/index.jade
```diff
@@ -0,0 +1,68 @@
...
+        option(value="Spectral") Spectral     
+        option(value="RdYlGn") RdYlGn     
+        option(value="Accent") Accent     
+        option(value="Dark2") Dark2     
+        option(value="Paired") Paired     
+        option(value="Pastel1") Pastel1     
+        option(value="Pastel2") Pastel2     
+        option(value="Set1") Set1     
+        option(value="Set2") Set2    
+        option(value="Set2") Set2
```
jbeezley: It would be better to generate these using template [iteration](http://jade-lang.com/reference/iteration/) on the colorbrewer object.

dorukozturk:
I think this is ready for a second review. @manthey @jbeezley

jbeezley:
It looks much better, thanks.  One thing I'm noticing is that the overlay is misaligned a bit.  It looks like it is shifted to the north a bit.  The tiles are being placed correctly, so it is either a problem with the bounding box computation (in EPSG:3857 coordinates) or the geoserver is misinterpreting them.  What's odd is that the other wms example is done the same way, but there doesn't seem to be a similar misalignment.

dorukozturk:
@jbeezley You are right. There was an error with the projection. I fixed it.

jbeezley:
[LGTM](http://lgtm.in/i/sKGC3urSp)

Issue #494, opened 2016-01-05
Outlines for the polygon feature


aashish24:

manthey:
Added in PR #597.

Issue #596, opened 2016-07-11
Line features in GL don't render correctly


manthey:
Wide lines at acute vertices appear to have the most trouble:

![image](https://cloud.githubusercontent.com/assets/8781639/16732788/d342abe0-474c-11e6-8c24-0c3f4d53c105.png)

manthey:
There are two separate bugs here.

One is that some angles were miscalculated.  A fix for that is part of PR #597.

The second is the inside of acute angles.  If a line segment is short enough, the interior vertex of the intersection is badly placed.

manthey:
Here is an image that shows a variety of line artifacts.  The blue lines are full-opacity lines with repeated junction points, while the grey lines are half-opacity lines without repeated junctions (i.e., if the grey line is [A, B, C, D], the blue line is [A, B, B, C, C, D]).

From left to right:
- Wide lines that overlap do not have constant opacity.  I don't see a way to fix this short of computing the bounding polygon for the line with width and then rendering it as a generic polygon.  We probably will just accept this as a limitation.
- Lines that exceed the miter limit don't render correctly.  We need more vertices to properly render a bevel join.  If we switch to generating the join normals during line vertex creation (as opposed to computing them in the vertex shader), we could generate an intermediate line segment or triangle when we exceed the miter limit.  Otherwise, we should at least detect this in the line vertex creation and pass a repeated vertex for the next or previous vertex.
- The center polyline is drawn correctly.
- This shows again the wide line overlap opacity issue.
- Lines that are wide, short, and joined at acute angles generate strange triangles.
 
![image](https://cloud.githubusercontent.com/assets/8781639/16958283/d7c94f0e-4dad-11e6-8097-dd6d4a8a2205.png)

manthey:
In addressing this issue, we have the opportunity to also features to our lines.

Features we can add:
- Line cap style (butt, round, square)
- Line join style (miter, bevel, round, miter-clip)
- Miter limit (currently this isn't configurable)
- Stroke offset.  This could be simple, such as drawing the entire line to the center, left, or right, or it could be a float value that offsets the line anywhere between those locations.
- Antialiasing (the can be configurable in its magnitude).

Features that aren't being considered right now:
- Dashing
- Textured lines (these would be straightforward to add in the 2-D case).

[...]

PR #597, opened 2016-07-11
Add stroking to polygons

manthey:
This is done by creating a sub-feature of lines for the polygons if the polygon feature has any value other than false for the stroke style (either true or a function).  Lines are created for each exterior and interior loop in the polygons, and are not set for interaction.  Lines are created with the same starting and ending point.

Strokes can have all of the properties that lines get -- strokeWidth, strokeColor, strokeOpacity, strokeStyle.  Lines are rendered with the same renderer as the polygons (and on the same level).

Polygons now have a flag to turn off filling the polygon (though the polygon is still created and still earcut, so this doesn't speed things up for polygons with fill = false).

It would be better to add a flag to the line feature to indicate a closed loop.

Changed the gl.object class to not re-subclass sceneObject, as all gl features are already subclasses of sceneObject.

This requires a vgl update to allow unbinding vertex data.  That change requires an update to the quad feature.

Fixed a typo in the lineFeature.

This resolves issue #494.


manthey:
This is only a WIP until a new VGL version is available.

jbeezley:
This looks good, I just need to fix a couple of the selenium tests before it goes in.

jbeezley:
I found out the changes here broke styling in the `jsonReader` class because the call signature of the `stroke` and `fill` styles changed to be per polygon rather than per vertex.  The change makes sense because you can't interpolate a boolean.

Anyway, I updated the `glMultiPolygon` test to add strokes and fixed a couple of other issues that were causing spurious test failures.  With these changes, it LGTM.

jbeezley:
One other thing I noticed that we should look into after this goes in is the quality of the stroking, which at the moment is pretty bad.  See the screenshot attached, which is at display resolution (not zoomed in).

1. Many line segments seem to be missing triangles.  This could be due to numerical precision.
2. It almost looks like the fill bleeds out beyond the stroke, but it could be some weird antialiasing effect.

screen shot 2016-07-16 at 1 24 37 pm

manthey:
I filed issue #596 about the way lines are rendered.  They are rendered after the fill, so should never underneath it.  You'll see the exact same lines (and same errors) with a fill opacity of 0.  It appears like the vertices of some of the triangles are wrong due to an error in how mitering is calculated.

manthey:
@jbeezley Can you check if the selenium tests need to be updated after the line fix?

jbeezley:
Done.  LGTM.

src/gl/lineFeature.js
```diff
@@ -96,6 +94,8 @@ var gl_lineFeature = function (arg) {
           '  else  anglePrev = atan(deltaPrev.y / aspect, deltaPrev.x);',
           '  if (deltaNext.xy == vec2(0.0, 0.0)) angleNext = anglePrev;',
           '  float angle = (anglePrev + angleNext) / 2.0;',
+          '  if (abs(anglePrev - angleNext) >= PI)',
+          '    angle += PI;',
```
aashish24: +1 :+1:

PR #598, opened 2016-07-13
Replace mapquest tiles in examples

jbeezley:
<http://devblog.mapquest.com/2016/06/15/modernization-of-mapquest-results-in-changes-to-open-tile-access/>


aashish24:
LGMT :+1:

PR #601, opened 2016-07-19
Fix delete polygon with stroke

jbeezley:
I noticed this while testing the examples for the release.  Trying to remove a stroked polygon was throwing an exception because of a typo and a bug in how the internal references were managed.

I also bumped to version 0.10.1 because I already published 0.10.0.  😞


manthey:
LGTM.

Issue #599, opened 2016-07-18
Add loop flag for lines


manthey:
For lines that form a closed loop, add a boolean flag indicating this so that the beginning and end are joined with proper miters.  This will also make it optional to repeat the first point as the last point.

manthey:
Added in PR #602.

PR #602, opened 2016-07-21
Add closed flag to line features

manthey:
The closed flag will automatically join the first and last points.  The last point in the line need not be duplicated.

Polygons now always use closed lines.

The picking example can take an optional parameter ?closed=true to show closed lines, allowing picking tests on the close segment.

BREAKING CHANGE: Fixed inconsistency in `pointSearch`.  The `pointSearch` function returned different keys depending on whether data was present or not in both the point and line features.  Specifically, when there is no data in a feature, or other `pointSearch` functions (in quad, polygon, and the base class feature), `pointSearch` returns {index: [...], found: [...]}.  When data is present, the point and line features returned {index: [...], data: [...]}.  This has been changed to always return {index: [...], found: [...]}.

Added unit tests for line features.

Resolves issue #599.

### DIFF
--- a/src/gl/lineFeature.js
+++ b/src/gl/lineFeature.js
@@ -407,6 +407,7 @@ var gl_lineFeature = function (arg) {
   ////////////////////////////////////////////////////////////////////////////
   this._exit = function () {
     m_this.renderer().contextRenderer().removeActor(m_actor);
+    m_actor = null;
     s_exit();
   };
 
--- a/src/gl/lineFeature.js
+++ b/src/gl/lineFeature.js
@@ -127,11 +127,11 @@ var gl_lineFeature = function (arg) {
 
   function createGLLines() {
     var data = m_this.data(),
-        i, j, k, v,
+        i, j, k, v, lidx,
         numSegments = 0, len,
         lineItem, lineItemData,
         vert = [{}, {}], vertTemp,
-        pos, posIdx3,
+        pos, posIdx3, firstpos, firstPosIdx3,
         position = [],
         posFunc = m_this.position(),
         strkWidthFunc = m_this.style.get('strokeWidth'),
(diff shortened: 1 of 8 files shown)


src/gl/lineFeature.js
```diff
@@ -151,6 +152,19 @@ var gl_lineFeature = function (arg) {
         position.push(pos.x);
         position.push(pos.y);
         position.push(pos.z || 0.0);
+        if (!j) {
+          firstpos = pos;
+        }
+      }
+      if (lineItem.length > 2 && closedFunc(data[i], i)) {
```
aashish24: Why passing two arguments to closedFunc?

manthey: Our general method of getting style information from functions is to pass two parameters: the current data item and its index.

manthey: For sub-components, like points within a line, the style information is obtained by calling a function with four parameters: the current sub-item (a vertex, for instance), the sub-item's index, the current item (the line), and the item index.

aashish24:
Looking pretty good to me, just had one questions. Also, on returning data, was the data object pointed to the original data object?

src/gl/lineFeature.js
aashish24: > Our general method of getting style information from functions is to pass two parameters: the current data item and its index.

Sure, makes sense.

manthey:
No.  When we called pointSearch, the `data` entry was an array, each element of which is one of the original values from the original data object.  The only change there was renaming pointSearch's resulting value `data` to `found` to be consistent.  `found` still is an array with each element one of the original values from the original data object.

aashish24:
> No. When we called pointSearch, the data entry was an array, each element of which is one of the original values from the original data object. The only change there was renaming pointSearch's resulting value data to found to be consistent. found still is an array with each element one of the original values from the original data object.

ah, yes, I see that now, the change is just a key change, the value remains the same. 

```
-      data: found,
 +      found: found,
```

:+1:

aashish24:
Lgtm +1

PR #603, opened 2016-07-27
Refactor how actions are specified and detected.

manthey:
This is primarily to make the code reusable in other modules.  This results in some improvements:
* Modifier keys can be ignored.  For instance, ctrl-left drag defaults to rotation, shift must not be pressed, but the state of alt and meta is irrelevant.
* You could add an action that requires two mouse buttons.

This is a breaking change.

### DIFF
--- a/src/action.js
+++ b/src/action.js
@@ -1,0 +1,17 @@
+//////////////////////////////////////////////////////////////////////////////
+/**
+ * Common object containing all action types that are provided by the GeoJS
+ * API.
+ */
+//////////////////////////////////////////////////////////////////////////////
+var geo_action = {
+  momentum: 'geo_action_momentum',
+  pan: 'geo_action_pan',
+  rotate: 'geo_action_rotate',
+  select: 'geo_action_select',
+  unzoomselect: 'geo_action_unzoomselect',
+  zoom: 'zoom',
+  zoomselect: 'geo_action_zoomselect'
+};
+
+module.exports = geo_action;
--- a/src/action.js
+++ b/src/action.js
@@ -10,7 +10,7 @@ var geo_action = {
   rotate: 'geo_action_rotate',
   select: 'geo_action_select',
   unzoomselect: 'geo_action_unzoomselect',
-  zoom: 'zoom',
+  zoom: 'geo_action_zoom',
   zoomselect: 'geo_action_zoomselect'
 };
 
--- a/src/event.js
+++ b/src/event.js
@@ -226,6 +226,63 @@ geo_event.unzoomselect = 'geo_unzoomselect';
 
 //////////////////////////////////////////////////////////////////////////////
 /**
+ * Triggered when an action is initiated with mouse down
+ *
+ * @property {object} state The action state
+ * @property {geo.mouseState} mouse The mouse state
+ * @property {object} event The triggering event
+ */
+//////////////////////////////////////////////////////////////////////////////
+geo_event.actiondown = 'geo_actiondown';
+
+//////////////////////////////////////////////////////////////////////////////
+/**
+ * Triggered when an action is being processed during mouse movement.
+ *
+ * @property {object} state The action state
+ * @property {geo.mouseState} mouse The mouse state
+ * @property {object} event The triggering event
+ */
+//////////////////////////////////////////////////////////////////////////////
+geo_event.actionmove = 'geo_actionmove';
+
+//////////////////////////////////////////////////////////////////////////////
+/**
+ * Triggered when an action is ended with a mouse up.
+ *
+ * @property {object} state The action state
+ * @property {geo.mouseState} mouse The mouse state
+ * @property {object} event The triggering event
+ */
+//////////////////////////////////////////////////////////////////////////////
+geo_event.actionup = 'geo_actionup';
+
+//////////////////////////////////////////////////////////////////////////////
+/**
+ * Triggered when an action results in a selection.
+ *
+ * @property {object} state The action state
+ * @property {geo.mouseState} mouse The mouse state
+ * @property {object} event The triggering event
+ * @property {object} lowerLeft Lower left of selection in screen coordinates
+ * @property {object} upperRight Upper right of selection in screen coordinates
+ */
+//////////////////////////////////////////////////////////////////////////////
+geo_event.actionselection = 'geo_actionselection';
+
+//////////////////////////////////////////////////////////////////////////////
+/**
+ * Triggered when an action is triggered with a mouse wheel event.
+ *
+ * @property {object} state The action state
+ * @property {geo.mouseState} mouse The mouse state
+ * @property {object} event The triggering event
+ */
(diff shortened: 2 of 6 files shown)


src/mapInteractor.js
```diff
@@ -64,28 +55,66 @@ var mapInteractor = function (args) {
...
+       * other values that are falsy are ignored.
+       *   Available actions:
+       * pan, zoom, rotate, select, zoomselect, unzoomselect, click
+       *   Available events:
+       * left, right, middle, wheel
+       *   Available modifiers:
+       * shift, ctrl, alt, meta
+       */
+      actions: [{
+        action: 'pan',
```
aashish24: @manthey should we define action and event values as some keys at one single place? for e.g. (or some other way)

user_actions = {
pan: "pan",
zoom: "zoom"
}

and then we can have: 

action: user_actions.pan

Thoughts?

src/mapInteractor.js
```diff
@@ -636,29 +632,19 @@ var mapInteractor = function (args) {
...
-    } else if (eventMatch(m_options.rotateMoveButton, m_options.rotateMoveModifiers)) {
-      action = 'rotate';
-    } else if (eventMatch(m_options.selectionButton, m_options.selectionModifiers)) {
-      action = 'select';
-    } else if (eventMatch(m_options.zoomSelectionButton, m_options.zoomSelectionModifiers)) {
-      action = 'zoomselect';
-    } else if (eventMatch(m_options.unzoomSelectionButton, m_options.unzoomSelectionModifiers)) {
-      action = 'unzoomselect';
-    }
+    action = eventMatch(m_mouse.buttons, m_mouse.modifiers, m_options.actions);
```
aashish24: Nice

src/util/init.js
```diff
@@ -291,6 +291,71 @@
...
+          action.modifiers = actionModifiers;
+        }
+      }
+    },
+
+    /**
+     * Determine if the current events and modifiers match a known action.
+     *
+     * @param {object} events: an object where each event that is currently
+     *    active is truthy.  Common events are left, right, middle, wheel.
```
aashish24: truthy?

src/util/init.js
```diff
@@ -291,6 +291,71 @@
...
+     & @param {object} modifiers: an object where each currently applied
+     *    modifier is truthy.  Common modifiers are shift, ctrl, alt, meta.
+     * @param {Array} actions: a list of actions to compare to the events and
+     *    modifiers.  The first action that matches will be returned.
+     * @returns action A matching action or undefined.
```
aashish24: should we say return a "string"? 

@returns {string} A matching action or undefined

aashish24:
Mostly looks good to me, has few suggestions.

aashish24:
@manthey are you planning to update this branch with suggestions as we discussed?

manthey:
@aashish24 Yes, soon.

aashish24:
thanks @manthey it's looking very good, will post my final comments by tomorrow morning.

src/util/init.js
aashish24: @manthey not sure if you saw this comment

src/event.js
```diff
@@ -226,6 +226,63 @@ geo_event.unzoomselect = 'geo_unzoomselect';
...
 //////////////////////////////////////////////////////////////////////////////
 /**
+ * Triggered when an action is initiated with mouse down
+ *
+ * @property {object} state The action state
+ * @property {geo.mouseState} mouse The mouse state
+ * @property {object} event The triggering event
+ */
+//////////////////////////////////////////////////////////////////////////////
+geo_event.actiondown = 'geo_actiondown';
```
aashish24: @manthey nitpick: we use camelcase style for other event for example 

geo_event.worldChanged = 'geo_worldChanged';

Should we change geo_event.actiondown to geo_event.actionDown? (and similarly for others too?)

src/mapInteractor.js
```diff
@@ -64,28 +55,82 @@ var mapInteractor = function (args) {
...
+       * modifiers: the name of a modifier or an object with modifier names for
+       *    keys and boolean values that indicates the combination of modifiers
+       *    that trigger this action.
+       * selectionRectangle: true if a selection rectangle should be shown
+       *    during the action.
+       * name: a string that can be used to reference this action.
+       * owner: a string that can be used to reference this action.
+       */
+      actions: [{
+        action: 'pan',
```
aashish24: Also here should we define actions as 

{action: geo_event.pan, ....}

also the keyword event is confusing here. Should we use 'input'? modifiers looks good to me.

src/event.js
manthey: We haven't been consistent: `geo_event.brushend = geo_brushend`, for instance.  I can change the actions to camel case or not as you prefer.  I'd be tempted to make the events consistent in casing, but that would be a breaking change for not much benefit.

src/mapInteractor.js
manthey: Actions don't necessarily correspond to events (though currently they all do), so to have constants for them I think it would be more proper to make another source file and define geo_action.pan = 'pan', etc.  What do you think?

manthey: I can change 'event' in this context to 'input'.

aashish24: > Actions don't necessarily correspond to events (though currently they all do), so to have constants for them I think it would be more proper to make another source file and define geo_action.pan = 'pan', etc. What do you think?

I like this idea :+1:

aashish24: > I can change 'event' in this context to 'input'.

thanks!

src/event.js
aashish24: > We haven't been consistent: geo_event.brushend = geo_brushend, for instance. I can change the actions to camel case or not as you prefer. I'd be tempted to make the events consistent in casing, but that would be a breaking change for not much benefit.

I think it would be nice to have the consistency IMO but we can file the bug for now and break it when we release 1.0. I would be okay with that. Thoughts?

aashish24:
other than my new comments, the branch LGTM :+1:

manthey:
@aashish24 I've added actions as a set of defined constants and updated the tests.

src/action.js
```diff
@@ -0,0 +1,17 @@
...
+ * API.
+ */
+//////////////////////////////////////////////////////////////////////////////
+var geo_action = {
+  momentum: 'geo_action_momentum',
+  pan: 'geo_action_pan',
+  rotate: 'geo_action_rotate',
+  select: 'geo_action_select',
+  unzoomselect: 'geo_action_unzoomselect',
+  zoom: 'zoom',
```
aashish24: I believe this is intentional (meaning not "geo_zoom")?

aashish24:
Looking much better @manthey just had one question on "zoom"

manthey:
@aashish24 Updated.

aashish24:
LGTM :+1: thanks

PR #604, opened 2016-07-27
Fix an interactor bug with momentum.

manthey:
Using the default map interactor options, if you right click or right drag on the map, mousemove event cease getting generated.  The action state was marked as zoom or momentum after the mouse up (since it was performing that action), and this is never cleared as clearing is done on mouse up.  This blocked generating mousemove events.

Also, tweak one of the polygon tests to try to eliminate occasional test failures.

### DIFF
--- a/src/mapInteractor.js
+++ b/src/mapInteractor.js
@@ -636,7 +636,7 @@ var mapInteractor = function (args) {
         (!m_mouse.buttons.left || m_options.click.buttons.left) &&
         (!m_mouse.buttons.right || m_options.click.buttons.right) &&
         (!m_mouse.buttons.middle || m_options.click.buttons.middle)) {
-      m_clickMaybe = true;
+      m_clickMaybe = {x: m_mouse.page.x, y: m_mouse.page.y};
       if (m_options.click.duration > 0) {
         window.setTimeout(function () {
           m_clickMaybe = false;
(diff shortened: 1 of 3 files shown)


manthey:
To see the problem, load the tile example page from http://opengeoscience.github.io/geojs/examples/tiles/, open a javascript console and type `tileDebug.map.geoOn('geo_mousemove', function (evt) { console.log(evt.map.x, evt.map.y); })`. Now when you move the mouse, you'll see screen coordinates in the console.  Focus the map window, then right click. The screen coordinates will stop updating until you left click.

src/mapInteractor.js
```diff
@@ -636,7 +636,7 @@ var mapInteractor = function (args) {
         (!m_mouse.buttons.left || m_options.click.buttons.left) &&
         (!m_mouse.buttons.right || m_options.click.buttons.right) &&
         (!m_mouse.buttons.middle || m_options.click.buttons.middle)) {
-      m_clickMaybe = true;
+      m_clickMaybe = {x: m_mouse.page.x, y: m_mouse.page.y};
```
aashish24: Looks like we
Are using clickMaybe in many ways : boolean and object? Wondering why.

manthey: We can receive a mousemove event from some browsers where there is no movement.  We need to record where the mouse down occurred, but we only care until we decide that a click is impossible.  At present, a click only gets passed through if there is no movement.  We could make it so that there is a small threshold of movement that would not trigger moving events, but would still allow clicks.

aashish24: Sounds good. I was thinking from consistency point of view.

aashish24:
LGTM  (but didn't run it yet) :+1:

PR #605, opened 2016-08-08
Change how zoom on a clamped map works.

manthey:
Before, if you have a clamped map or image and zoom in via the mouse wheel near an edge with adjacent empty space, the location of the map slides back and forth, rather than the point where you are zooming remaining under the mouse cursor.

Now, the coordinates under the mouse cursor remain the same, which means the map may not be clamped as fully as it could.  You can drag the map to reduce the empty space that clamp would have removed in the past, but not to increase it.  If you zoom in on an area off of the map, the clamping will still apply to prevent more than half of the viewport from having empty space.  This also affects how rotation clamping works.

There was an oddness where the reported map center was not the visible map center if clamping had occurred.  This has been fixed.

Also, noticed some needless code in the vgl renderer, so removed it.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -367,6 +367,9 @@ var map = function (arg) {
       return m_zoom;
     }
     var evt, bounds;
+    /* If we are zooming around a point, ignore the clamp bounds */
+    var aroundPoint = (origin && (origin.mapgcs || origin.geo) && origin.map);
+    var ignoreClampBounds = aroundPoint;
 
     /* The ignoreDiscreteZoom flag is intended to allow non-integer zoom values
      * during animation. */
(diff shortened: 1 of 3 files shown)


src/map.js
```diff
@@ -1716,12 +1739,24 @@ var map = function (arg) {
   }
 
   /**
-   * Return the nearest valid bounds maintaining the
-   * width and height. Does nothing if m_clampBounds* is
-   * false.
+   * Return the nearest valid bounds maintaining the width and height.  Does
```
aashish24: I am not entire sure about these options and will need some discussion to understand it completely.

manthey: It is easiest to show the issues that had existed and show what the new behavior allows.

For example of the current behavior, go here: http://opengeoscience.github.io/geojs/examples/tiles/?clampBoundsX=true&wrapX=false&zoom=0 using a browser with a window wider than it is tall.  This is our tile example with wrapping turned off and clamping turned on.  Hover your mouse over New Zealand, and use the mouse wheel to scroll in.  New Zealand will shift from underneath the cursor, so you zoom in on a different location.

This can be corrected by not clamping the bounds on zoom, but since clamping is a desired feature, there is code to clamp eventually (don't let the map get more than half off the screen), and to allow panning the map back to where the delta.  If you run the same example on this PR's code, New Zealand will stay under the mouse cursor.  You'll see white background to the right of New Zealand.  You can pan the map right (reducing that white space), but not left (since that increases the clamping violation).

PR #606, opened 2016-08-09
Allow showing partial tiles on the edges of an image or map.

manthey:
If the tileLayer tilesMaxBounds function is defined, some tiles may be cropped.  Tile cropping is implemented in the gl, canvas, and null renderers (but not d3).

This fixes issue #544.

### DIFF
--- a/src/gl/quadFeature.js
+++ b/src/gl/quadFeature.js
@@ -43,7 +43,7 @@ var gl_quadFeature = function (arg) {
     'void main(void) {',
     '  mediump vec4 color = texture2D(sampler2d, iTextureCoord);',
     '  if (iTextureCoord.s > crop.s || 1.0 - iTextureCoord.t > crop.t) {',
-    '    return;',
+    '    discard;',
     '  }',
     '  color.w *= opacity;',
     '  gl_FragColor = color;',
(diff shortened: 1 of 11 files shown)


src/gl/quadFeature.js
```diff
@@ -35,12 +35,25 @@ var gl_quadFeature = function (arg) {
...
       m_quads;
+  var fragmentShaderImageSource = [
+    'varying highp vec2 iTextureCoord;',
+    'uniform sampler2D sampler2d;',
+    'uniform mediump float opacity;',
+    'uniform highp vec2 crop;',
+    'void main(void) {',
+    '  mediump vec4 color = texture2D(sampler2d, iTextureCoord);',
+    '  if (iTextureCoord.s > crop.s || 1.0 - iTextureCoord.t > crop.t) {',
+    '    return;',
```
aashish24: I think discard would be a better option here.

aashish24:
@manthey so this is only going to crop towards right and bottom? What happens to the cropping towards left and top?

manthey:
@aashish24 Yes, currently all tile sets I know of have the 0, 0 pixel within a tile, so only right and bottom cropping is necessary.  All partial-tile examples I have are from medical images, and this satisfies the need there.  If you have a real-world tile set which has partial tiles on the top or left, then we can add support at that time.

src/gl/quadFeature.js
manthey: Thanks.

PR #609, opened 2016-08-26
Ignore geojson features with empty coordinates

jbeezley:
GeoJSON features with empty coordinate arrays validate correctly at geojsonlint, so we should handle them somehow.  In this PR, I just filter them out before returning from the reader otherwise the renders throw all sorts of ugly errors.


manthey:
LGTM

PR #611, opened 2016-08-31
Fix a bug in gcsTileBounds

manthey:
When asking for a specific gcs transform in gcsTileBounds, the transform was reversed.

Add a test to prove that this works as we expect.


jbeezley:
LGTM

PR #613, opened 2016-09-02
Annotation layer

manthey:
This requires PR #612.

This adds an annotation layer and an example that uses it.  Currently, there are two annotation types: polygon and rectangle.  It should be straightforward to allow the annotations to be edited and to add metadata to the annotations (such as name, changing the style, or arbitrary other data).

There are no tests yet.


aashish24:
Nice! Will start the review. I am assuming that we are going to add tests as well for this PR?

manthey:
Yes, that's why this is marked a work-in-progress.

kotfic:
@manthey Would it be a big lift to include point annotations in this PR as well?

manthey:
@kotfic It won't be much work -- I'll get to it today or tomorrow.

aashish24:
> Yes, that's why this is marked a work-in-progress.

roger that, thanks @manthey

manthey:
@kotfic Point annotations.  In the annotations example, click the middle mouse button to enter point-creation mode, then click the left mouse button to add a point.

jbeezley:
One problem with using a middle click is that apple devices don't support it.

manthey:
It is only for development -- I intend to add some control widgets to the example so you can pick what annotation you want and then use left-click for all of them.  It will go away by the time the WIP label comes off.

manthey:
In the annotation example, you can now edit annotation properties (color, stroke width, etc.).

aashish24:
> In the annotation example, you can now edit annotation properties (color, stroke width, etc.).

very cool! I guess we are still waiting for tests?

manthey:
I'll work on tests next.

kotfic:
@manthey This is really awesome! Something to consider for the example -  it would be nice if pressing escape exited polygon creation.

aashish24:
> it would be nice if pressing escape exited polygon creation.

+1

manthey:
To handle keyboard events nicely, I've used a library called [Mousetrap](https://github.com/ccampbell/mousetrap) in the past.  For just handle escape it is overkill, but I can envision more interactions (nudging points with the arrow keys, etc).  Opinions?

jbeezley:
I'm fine with mousetrap.  Looks like a pretty small library, and it would remove a lot of custom code for detecting mouse and keyboard events we have in place.

aashish24:
> To handle keyboard events nicely, I've used a library called Mousetrap in the past. For just handle escape it is overkill, but I can envision more interactions (nudging points with the arrow keys, etc). Opinions?

+1, looks like a nice little library

manthey:
Now with tests.

aashish24:
awesome @manthey thanks. I/we will try to review this asap. On this note, I guess the next task in the list would be "editable annotations" followed by "adding text to annotations". Is that correct? 

Thanks,

manthey:
@aashish24 I thought it was the opposite order: adding text, then editing, but it is your choice.

aashish24:
> @aashish24 I thought it was the opposite order: adding text, then editing, but it is your choice.

I see. For the NEX, editing has become a little bit higher priority.

src/action.js
```diff
@@ -11,7 +11,11 @@ var geo_action = {
   select: 'geo_action_select',
   unzoomselect: 'geo_action_unzoomselect',
   zoom: 'geo_action_zoom',
-  zoomselect: 'geo_action_zoomselect'
+  zoomselect: 'geo_action_zoomselect',
+
+  // annotation actions
+  annotation_polygon: 'geo_annotation_polygon',
```
aashish24: should it be annotate_polygon (verb vs noun?) similar to 'select'?

src/annotationLayer.js
```diff
@@ -0,0 +1,486 @@
...
+                         Math.pow(coord1.y - coord2.y, 2));
+    return dist;
+  };
+
+  /**
+   * Add an annotation to the layer.  The annotation could be in any state.
+   *
+   * @param {object} annotation the annotation to add.
+   */
+  this.addAnnotation = function (annotation) {
```
aashish24: +1

src/event.js
```diff
@@ -428,4 +428,57 @@ geo_event.camera.projection = 'geo_camera_projection';
 //////////////////////////////////////////////////////////////////////////////
 geo_event.camera.viewport = 'geo_camera_viewport';
 
+////////////////////////////////////////////////////////////////////////////
+/**
+ * These events are triggered by the annotation layer.
+ * @namespace geo.event.annotation
+ */
+////////////////////////////////////////////////////////////////////////////
+geo_event.annotation = {};
```
aashish24: same here verb vs noun? I do not have strong preferences and my comments are more as second thoughts.

src/annotation.js
```diff
@@ -0,0 +1,607 @@
...
+ * @param {string} [options.name] A name for the annotation.  This defaults to
+ *    the type with a unique ID suffixed to it.
+ * @param {geo.annotationLayer} [options.layer] a reference to the controlling
+ *    layer.  This is used for coordinate transforms.
+ * @param {string} [options.state] initial annotation state.  One of the
+ *    annotation.state values.
+ * @returns {geo.annotation}
+ */
+/////////////////////////////////////////////////////////////////////////////
+var annotation = function (type, args) {
```
aashish24: where are types defined?

aashish24:
I liked the design overall :+1: :+1: :+1: with separation of annotations, features, and layers.

src/action.js
manthey: I was viewing this as the action belongs to the annotation module and the action is to make a polygon.

src/annotation.js
manthey: Each annotation subclass defines the type, so this is fully extensible and types aren't predefined.  I don't expect users to call the annotation base class by itself unless they are defining a new type.  Internally, we have `polygon`, `rectangle`, and `point` right now, but I could see adding lots more -- lines, labels, ellipses, etc.

aashish24: Right, I saw the code " annotation.call(this, 'rectangle', args);" so the type names are defined by each subclass. I am wondering if could return a list of supported types (assuming after the annotations are initialized) as something like this could be useful for ui. But I don't think this is critical or needed at the moment. May be (no strong preference) in the comment, we could add a line in the doc stating that type is defined by sub-classes as they extend the base annotation class.

src/action.js
aashish24: Sure this is fine with me.

src/annotation.js
manthey: At the moment there isn't an annotation registry.  I think it would be good to have one, as this (a) allows easily enumerating available annotations, and (b) each annotation can list its required features, so when you ask for an annotation layer, instead of asking for a specific renderer or a list of required features, you could alternately supply a list of annotations that would be used and that would be used to select an appropriate renderer.

It isn't much work to add that.  I can do it now or deferred it to the next PR.

aashish24: thanks. next PR is fine with me.

PR #615, opened 2016-09-16
Improve robustness of heatmap test.

manthey:
The animation frame was mocked after an animation frame might have been requested, which could lead to a failure to step through animations properly.

### DIFF
--- a/tests/cases/heatmap.js
+++ b/tests/cases/heatmap.js
@@ -1,4 +1,4 @@
-// Test geo.core.osmLayer
+// Test geo.core.heatmap
 var geo = require('../test-utils').geo;
 var $ = require('jquery');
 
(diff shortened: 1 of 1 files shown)


tests/cases/heatmap.js
```diff
@@ -1,4 +1,4 @@
-// Test geo.core.osmLayer
+// Test geo.core.heatmap
```
aashish24: good catch

tests/cases/heatmap.js
```diff
@@ -31,6 +31,7 @@ describe('canvas heatmap feature', function () {
   });
 
   it('Setup map', function () {
+    mockAnimationFrame();
```
aashish24: why this change (moving the code to here)? I guess I have an idea but still curious

manthey:
We need to run `mockAnimationFrame` before there is a chance that something calls `requestAnimationFrame`.  `map.draw` can do that, and this was being mocked after that.

aashish24:
> We need to run mockAnimationFrame before there is a chance that something calls requestAnimationFrame. map.draw can do that, and this was being mocked after that.

Okay thanks not sure how to test the failed case but the explanation makes sense to me. 

LGTM :+1:

PR #616, opened 2016-09-23
Add an annotation registry.

manthey:
Let renderers be selected based on the annotations that will be used.

Also, there were a bunch of TODO statements asking to send a warning if items were re-registered.  These have been addressed.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -3,6 +3,9 @@ var inherit = require('./inherit');
 var geo_event = require('./event');
 var transform = require('./transform');
 var registerAnnotation = require('./registry').registerAnnotation;
+var lineFeature = require('./lineFeature');
+var pointFeature = require('./pointFeature');
+var polygonFeature = require('./polygonFeature');
 
 var annotationId = 0;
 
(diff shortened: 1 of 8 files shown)


src/annotation.js
```diff
@@ -309,6 +311,8 @@ var rectangleAnnotation = function (args) {
 };
 inherit(rectangleAnnotation, annotation);
 
+registerAnnotation('rectangle', rectangleAnnotation, {polygon: true});
```
aashish24: +1

src/annotation.js
```diff
@@ -598,6 +605,8 @@ var pointAnnotation = function (args) {
 };
 inherit(pointAnnotation, annotation);
 
+registerAnnotation('point', pointAnnotation, {point: true});
```
aashish24: My only suggestion would is that features (keys) are not visible anywhere else. Should we create a list of it somewhere at the top may be? then you can say annotation.point: true

manthey: I'm not sure having another list aids much here.  There is no strict reason why an annotation should need to use the same feature that any other annotation used (I don't know what a heatmap annotation would look like, but we could have one).

I think if we wanted to make this more based on definitions, then we should probably add a `general` key to each base feature's capabilities map which contains the plain feature's name.  Then we would import the base feature (not the renderer-specific feature) so that we could specify one of its capabilities.

manthey: I just pushed a change which does this.

kotfic: @manthey Just as a side note,  heatmap annotations look like the old-school MS Paint spray can tool. I've seen papers where this is used to allow users to generate such distributions over vernacular geographic entities (e.g.  downtown).  There might actually be some useful applications of an annotation type like this (especially in medical imaging?)

aashish24:
> I just pushed a change which does this.

Looks great!

PR #617, opened 2016-09-26
Annotation geojson

manthey:
Annotation layers can output and input geojson.  Using its own geojson will reconstruct the annotations (though they may have different internal IDs).  Modifying the geojson updates the annotations.

The annotation example has the ability to live-edit the geojson.

This requires PR #616.


manthey:
@aashish24 This is ready for review.

aashish24:
Cool! Will review it soon, thanks :clap: :clap:

src/annotationLayer.js
```diff
@@ -351,6 +365,224 @@ var annotationLayer = function (args) {
...
+   *    geojson object.  If undefined, return the current annotations as
+   *    geojson.  This may be a JSON string, a javascript object, or a File
+   *    object.
+   * @param {boolean} clear: if true, when adding annotations, first remove all
+   *    existing objects.  If 'update', update existing annotations and remove
+   *    annotations that no longer exit,  If false, update existing
+   *    annotations and leave unchanged annotations.
+   * @param {string|geo.transform} [gcs] undefined to use the interface gcs,
+   *    null to use the map gcs, or any other transform.
+   * @param {boolean} includeCrs: if true, include the coordinate system in the
```
aashish24: should we rename includeCrs to includeGcs to be consistent?

examples/common/js/examples.js
```diff
@@ -8,7 +8,7 @@ var exampleUtils = {
       '&').map(function (n) {
         n = n.split('=');
         if (n[0]) {
-          this[decodeURIComponent(n[0])] = decodeURIComponent(n[1]);
+          this[decodeURIComponent(n[0].replace(/\+/g, '%20'))] = decodeURIComponent(n[1].replace(/\+/g, '%20'));
```
manthey: `decodeURIComponent` doesn't decode `+` to a space in Chrome, but urls are allowed to use `+` to encode spaces.

src/annotationLayer.js
manthey: In geojson it is called `crs`.  Since this affects the geojson output, I figured it should be crs in this context, not gcs.

aashish24: okay thats fine, I just thought that everywhere else we used gcs but I don't have strong feeling about it.

Issue #618, opened 2016-09-26
Configurable zoom slider


dorukozturk:
@jbeezley @manthey The zoom slider defaults to upper left corner. It would be great if widgets take a position argument so that you can put them on lower left or upper right.
Example: 
http://opengeoscience.github.io/geojs/examples/widgets/docs/main.html

manthey:
Specifically, the zoom slider *ignores* the position argument.  I'll submit a PR with a quick fix shortly.

manthey:
Resolved in PR #619.

PR #619, opened 2016-09-26
Allow slider widget to be repositioned

manthey:
It would be nicer if you could change the orientation and which ends had the + and - (or even if they appeared).

You can invoke the slide with the position option, which can contain offsets from the top, left, right, or bottom (or some combination thereof).  It can also take a width and height, which are honored, though height needs to be more than 3.5 times the width or stupid things will happen.

This addresses issue #618.


manthey:
@dorukozturk Give this a try, passing something like `{position: {right: 20, top: 20}}`

aashish24:
Looks reasonable to me but I will let @dorukozturk or @jbeezley approve it.

manthey:
@jbeezley I added another change.  If you'd prefer, I could back it out and add it as a separate PR.

I also notice we don't test setting a widget position after its creation.  I'll add a test for that and for the recent change.

jbeezley:
That's fine with me.   Thanks.

manthey:
@jbeezley If this most recent change LGTY let me know (or just merge).

Issue #620, opened 2016-09-28
Point clustering flickers when transitioning between zoom levels


jbeezley:
My first thought was that it was related to mouse wheel events quickly transitioning just above and below an integer in [this code](https://github.com/OpenGeoscience/geojs/blob/master/src/pointFeature.js#L113) but some quick debugging seems to indicate it is occurring in the rendering code.

<https://jsbin.com/fuleki/edit?js,output>

aashish24:
I can see the flicker on my linux and and my mac consistently. @manthey are you going to have a look at it?

manthey:
The flicker isn't caused directly by the zoom.  I'm working on hunting it down.

manthey:
Try this out to see if it resolves the issue on your machines:

https://jsbin.com/tuyiparofu/edit?js,output

Note that PR #621 is functionally the same changes, but based on current master instead of that latest release.

manthey:
Resolved by PR #621.

PR #621, opened 2016-09-29
Reduce point clustering flicker.

manthey:
There are two different contributing factors:

(1) some times zoom rounding caused a flicker.  In floating point, 7.99999999 is essentially 8.0, but their floors are different.  In general, round zoom values to six places.

(2) gratuitous drawing during the zoom event.  After reclustering, a point feature draw() call was being made.  This is never necessary (at least on current master).


manthey:
This resolves issue #620.

PR #625, opened 2016-10-13
boundsFromZoomAndCenter returned incorrect values.

manthey:
When using any gcs other than the map's gcs, the function performed two conversions on the map center's coordinates which resulted in erroneous results.


manthey:
@jbeezley This was the error that we ran into on the HistomicsTK project getting the image bounds.

jbeezley:
Probably yes.  I'll have to check to make sure.

PR #627, opened 2016-10-17
Upload performance artifacts to s3

jbeezley:
@mgrauer This will push the notes files that we are collecting on cdash up to s3 along with some build metadata.  On travis, the files uploaded to s3 will look something like this:

```json
{
    "build_outputs": [],
    "build_timestamp": "2016-10-17T12:40:07.724645",
    "data": {
        ...
    "datasets": [],
    "git_branch": "upload-s3-artifacts",
    "git_repo": "git@github.com:OpenGeoscience/geojs.git",
    "git_sha": "3950b54d682ace0cf1b3879882d0ba147a291c2b",
    "host": "travis",
    "regeneration_command": "npm run test",
    "travis_id": "168296732",
    "vcs": "git"
}
```

It is also possible to push data from any host in the same format using the `upload_notes.py` script if the user has the aws credentials, otherwise it will just print to stdout.  I pushed historical data mined from cdash in the same format, but for some reason my scripted missed all the builds from 2016.  I'll look into that later.

### DIFF
--- a/tests/test-utils.js
+++ b/tests/test-utils.js
@@ -9,6 +9,7 @@ var $ = require('jquery');
 
 var geo = require('../src');
 var _supported = geo.gl.vglRenderer.supported;
+var bowser = require('bowser');
 
 module.exports = {};
 
@@ -286,6 +287,7 @@ module.exports.logCanvas2D = function logCanvas2D(enable) {
  * @param note: the data to send.  This will be converted to JSON.
  */
 module.exports.submitNote = function submitNote(key, note) {
+  note.browser = bowser;
   return $.ajax({
     url: '/notes?key=' + encodeURIComponent(key),
     data: JSON.stringify(note),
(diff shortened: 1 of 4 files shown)


tests/test-utils.js
```diff
@@ -286,6 +287,7 @@ module.exports.logCanvas2D = function logCanvas2D(enable) {
  * @param note: the data to send.  This will be converted to JSON.
  */
 module.exports.submitNote = function submitNote(key, note) {
+  note.browser = bowser;
```
mgrauer: What does this do?

mgrauer: Probably it provides information about the browser to be included with notes, I see that in the commit message now.

jbeezley: Indeed.  There will be huge differences in performance between a real browser and phantomjs.  This is so we can track that.

mgrauer:
@thewtex This PR contains a nice utility `upload_notes.py` that should save you a lot of time.

thewtex:
@mgrauer nice, thanks for the pointer!

Issue #624, opened 2016-10-10
Scale-able point annotation


kotfic:
Currently point annotations have a fixed radius at all zoom levels.  This creates a visually jarring experience when zooming in and out,  especially in the presence of other polygon annotations. The point annotation appears to grow and shrink in relation to other polygon annotations.

manthey:
Added in PR #628.

PR #628, opened 2016-10-19
Allow point annotations to scale with zoom.

manthey:
Added a scaled property.  If false, points are a fixed size.  If true, the first time the point annotation is rendered, it is changed to the current zoom level.  If a zoom level, the radius of the point is scaled with the map so that it is the specified radius value at the scaled property's zoom level.


manthey:
This addresses issue #624.  ping @kotfic

aashish24:
@manthey wouldn't it be better if we have this option on the pointFeature (and just pass a flag to the shader). There might be some other complications though which I haven't thought about it but may be you did?

manthey:
@aashish24 We could, but I'd rather leave that as a separate task to do later (and if we do, this could be refactored).  Fundamentally, this works in all renderers, but adds a bit of overhead in that it needs to perform some action on zoom.  If we add it to the feature itself, it would be best to go into each renderer and support changing the point size in a more direct way (so we don't have to hook to the zoom event).

As before, the user can make radius a function, e.g. `function (d) { return 10 * Math.pow(2, map.zoom() - 6); }`, and then set `feature.modified()` in the zoom event callback.

aashish24:
> @aashish24 We could, but I'd rather leave that as a separate task to do later (and if we do, this could be refactored). Fundamentally, this works in all renderers, but adds a bit of overhead in that it needs to perform some action on zoom. If we add it to the feature itself, it would be best to go into each renderer and support changing the point size in a more direct way (so we don't have to hook to the zoom event).
> 
> As before, the user can make radius a function, e.g. function (d) { return 10 * Math.pow(2, map.zoom() - 6); }, and then set feature.modified() in the zoom event callback.

Thanks after our discussion yesterday this is +1 LGTM from me. 👍

PR #629, opened 2016-10-20
Add pixel coordinate utility function.

manthey:
Add a utility function to make it easier to create maps and tile layers in pixel coordinates with common settings and functions.  It seems wasteful to have this code in each non-map example.

Fix the deepzoom example.  The IIP Server sends partial tiles at the right and bottom edge of the tile set.  Our canvas renderer happens to work with this because it doesn't scale the tile images (which means it won't work on a retina tile).  Added a capabilities flag so we can ask for that feature rather than specifying a renderer.

### DIFF
--- a/examples/deepzoom/main.js
+++ b/examples/deepzoom/main.js
@@ -1,6 +1,10 @@
+/* globals utils */
+
 // Run after the DOM loads
 $(function () {
   'use strict';
+
+  var query = utils.getQuery();
 
   // custom tileLayer for deepzoom
   var DeepZoom = function (arg) {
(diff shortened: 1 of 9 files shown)


examples/deepzoom/main.js
```diff
@@ -13,38 +17,47 @@ $(function () {
...
-  DeepZoom.defaults = $.extend({}, geo.tileLayer.defaults, {
-    minLevel: 0,
-    maxLevel: 9,
+  // We know the size of the image we are requesting, and we want to use
+  // pixel coordinates on our map.
+  var sizeX = 103583, sizeY = 70014, tileSize = 256;
+  var defaultParams = geo.util.pixelCoordinateParams(
+        '#map', sizeX, sizeY, tileSize, tileSize);
+
+  DeepZoom.defaults = $.extend({}, geo.tileLayer.defaults, defaultParams.layer, {
```
jbeezley: Would it be better to mix in the tile layer defaults inside `pixelCoordinateParams`?

jbeezley:
This is useful.  It makes me wonder if we could add a new API for non-geospatial contexts such as `geo.image` and a layer type that apply these defaults automatically.

examples/deepzoom/main.js
manthey: We could.  We mostly use `osmLayer` rather than `tileLayer`, with this example doing something different yet and subclassing `tileLayer`.  In the case of using one of the existing layers, there is no virtue of mixing it in; here there is.

manthey:
I could see having a subclass of `geo.map`, though I'm not sold on the name `geo.image`.  We also would want a variant of the `osmLayer` class (`imageLayer` if we go with `geo.image`).  Because the image size and tile size are both used in calculating map and layer parameters, I'd like to avoid having to repeat passing those values if possible.

aashish24:
geo.image could be bit confusing since we could have a geo.image object in the future to represent a image (and not a container for images) but may be its okay.

PR #633, opened 2016-10-24
Replace rather than modify default arrays.

manthey:
When constructing a mapInteractor with options, make sure the whole set is replaced.  We use jquery's extend method.  When using this on something like $.extend([1, 2, 3], [4, 5]), you end up with [4, 5, 3], but for this purpose we want to end up with just the new array.

Also, give default names and an owner to the mapInteractor map actions so that they can be more easily referenced.

### DIFF
--- a/src/mapInteractor.js
+++ b/src/mapInteractor.js
@@ -90,38 +90,54 @@ var mapInteractor = function (args) {
       actions: [{
         action: geo_action.pan,
         input: 'left',
-        modifiers: {shift: false, ctrl: false}
+        modifiers: {shift: false, ctrl: false},
+        owner: 'geo.mapInteractor',
+        name: 'button pan'
       }, {
         action: geo_action.zoom,
         input: 'right',
-        modifiers: {shift: false, ctrl: false}
+        modifiers: {shift: false, ctrl: false},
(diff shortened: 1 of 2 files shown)


src/mapInteractor.js
```diff
@@ -90,38 +90,54 @@ var mapInteractor = function (args) {
       actions: [{
         action: geo_action.pan,
         input: 'left',
-        modifiers: {shift: false, ctrl: false}
+        modifiers: {shift: false, ctrl: false},
+        owner: 'geo.mapInteractor',
```
aashish24: @manthey why we need owner information?

aashish24:
LGTM other than the side note.

src/mapInteractor.js
manthey: We don't.  It is a convenience for other applications that may want to modify the default actions.  They can ask for actions based on name and owner to be certain that it is the default action and not one that was created somewhere else.

aashish24:
> We don't. It is a convenience for other applications that may want to modify the default actions. They can ask for actions based on name and owner to be certain that it is the default action and not one that was created somewhere else.

should we document it somewhere?

PR #634, opened 2016-10-24
Support for canvas elements on quads.

manthey:
Rather than using an image, a canvas can be rendered on a quad.  On the canvas renderer, after changing the canvas quad, a draw call is sufficient to update it.  On the vgl renderer, the texture is cached, so it will take more work to update the texture.

### DIFF
--- a/src/gl/quadFeature.js
+++ b/src/gl/quadFeature.js
@@ -423,6 +423,7 @@ capabilities[quadFeature.capabilities.image] = true;
 capabilities[quadFeature.capabilities.imageCrop] = true;
 capabilities[quadFeature.capabilities.imageFixedScale] = false;
 capabilities[quadFeature.capabilities.imageFull] = true;
+capabilities[quadFeature.capabilities.canvas] = false;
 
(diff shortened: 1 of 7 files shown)


src/gl/quadFeature.js
```diff
@@ -423,6 +423,7 @@ capabilities[quadFeature.capabilities.image] = true;
 capabilities[quadFeature.capabilities.imageCrop] = true;
 capabilities[quadFeature.capabilities.imageFixedScale] = false;
 capabilities[quadFeature.capabilities.imageFull] = true;
+capabilities[quadFeature.capabilities.canvas] = false;
```
manthey: We don't need it, but I want to explicitly document what capabilities different renderers have rather than implicitly document them.

aashish24: My only concern is that now we need something similar in canvas version? my slight preference would be not to list canvas in gl and gl in canvas (and what about d3).

aashish24:
> This is if we can render a canvas on a quad, not if we can use the canvas renderer.

ah, makes sense. The name was bit confusing. LGTM then.

PR #635, opened 2016-10-25
Report where in a quad a point is located.

manthey:
When a point search is performed on a quad feature, in addition to determining if the point is in a quad, determine where it is within the quad.

### DIFF
--- a/src/quadFeature.js
+++ b/src/quadFeature.js
@@ -138,10 +138,12 @@ var quadFeature = function (arg) {
    */
   ////////////////////////////////////////////////////////////////////////////
   this.pointSearch = function (coordinate) {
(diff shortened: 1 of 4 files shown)


src/quadFeature.js
```diff
@@ -161,12 +163,27 @@ var quadFeature = function (arg) {
             util.pointInPolygon(coordinate, poly2)) {
           indices.push(quad.idx);
           found.push(data[quad.idx]);
+          coordbasis = util.pointTo2DTriangleBasis(
+            coordinate, poly1[0], poly1[1], poly1[2]);
+          if (!coordbasis || coordbasis.x + coordbasis.y > 1) {
```
aashish24: some extra comments would be useful here

PR #637, opened 2016-10-27
Add a pixelmap feature

manthey:
This takes an image where the color values are treated as indices into the data data.  This is used to generate a new color for each index.  The resultant image is rendered as a quad.  The feature is available in the canvas renderer.

The pixelmap image can be specified as a url or given as a Image element.  The pixelmap feature will be a deferred jquery object if the image needs to be loaded.

This depends on PRs #633, #634, #635, and #636.

I've left the history of this branch as it has code that shows how this could be extended to non-canvas renderers (though it is slower).

The example uses a pixelmap of the United States and some political data that is probably already out of date.


manthey:
@aashish24 All dependent PRs have been merged in.  This is ready for review.

src/canvas/pixelmapFeature.js
```diff
@@ -0,0 +1,34 @@
...
+//////////////////////////////////////////////////////////////////////////////
+var canvas_pixelmapFeature = function (arg) {
+  'use strict';
+
+  if (!(this instanceof canvas_pixelmapFeature)) {
+    return new canvas_pixelmapFeature(arg);
+  }
+  pixelmapFeature.call(this, arg);
+
+  var object = require('./object');
```
aashish24: curious: why you had to do this?

src/pixelmapFeature.js
```diff
@@ -0,0 +1,451 @@
...
+      m_this.style('url', val);
+      m_this.dataTime().modified();
+      m_this.modified();
+    }
+    return m_this;
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Get the maximum index value from the pixelmap.  This is a value present in
```
aashish24: Nitpick: in could be in the next line since its probably on column 79-80

src/pixelmapFeature.js
```diff
@@ -0,0 +1,451 @@
...
+  };
+
+  ////////////////////////////////////////////////////////////////////////////
+  /**
+   * Get/Set mapColor accessor
+   *
+   * @returns {geo.pixelmap}
+   */
+  ////////////////////////////////////////////////////////////////////////////
+  this.mapColor = function (val) {
```
aashish24: @manthey I am wondering why don't we just call it color like other features? since it is primarily a color transfer function? Also, I would think this should be a style element.

src/pixelmapFeature.js
```diff
@@ -0,0 +1,451 @@
...
+        } else {
+          /* we were given a url, so construct a new image */
+          m_srcImage = new Image();
+          // Only set the crossOrigin parameter if this is going across origins.
+          if (src.indexOf(':') >= 0 &&
+              src.indexOf('/') === src.indexOf(':') + 1) {
+            m_srcImage.crossOrigin = this.style.get('crossDomain')() || 'anonymous';
+          }
+        }
+        m_srcImage.onload = function () {
```
aashish24: I believe this code is only needed for else if and else?

src/pixelmapFeature.js
```diff
@@ -0,0 +1,451 @@
...
+
+  /**
+   * Compute information for this pixelmap image.  It is wasterful to call this
+   * if the pixelmap has already been prepared (it is invalidated by a change
+   * in the image).
+   */
+  this._preparePixelmap = function () {
+    var i, idx, pixelData;
+
+    if (!m_srcImage || !m_srcImage.complete || !m_srcImage.naturalWidth ||
```
aashish24: since this is a repeating code,  may be  in this or next PR we could write a function instead?

src/pixelmapFeature.js
```diff
@@ -0,0 +1,451 @@
...
+        pixelData[i * 4 + 2] = color[2];
+        pixelData[i * 4 + 3] = color[3];
+      }
+    }
+    /* Place the updated area into the canvas */
+    m_info.context.putImageData(
+      m_info.imageData, 0, 0, 0, Math.floor(updateFirst / m_info.width),
+      m_info.width, Math.ceil((updateLast + 1) / m_info.width));
+
+    /* If we haven't made a quad feature, make one now.  The quad feature needs
```
aashish24: Nitpick (can be done in another PR): Spacing between . (dot) and the.

src/pixelmapFeature.js
aashish24: ah. I see later that it is a style element so why we need this method?

aashish24:
Awesome feature 👏 👏 just has some minor things in the review.

src/canvas/pixelmapFeature.js
manthey: In our various renderers, we have renderer-specific changes to objects, mostly to make a renderer-specific draw method.

src/pixelmapFeature.js
manthey: 'in' ends at column 79.  Do you really want to change to less than 80 line lengths?

src/pixelmapFeature.js
manthey: In many features, we have certain styles also accessible by functions (position often is, even if we store it in style as well).  This allows for an easy side effect, such as only marking this as changed if it is actually different.

The problem with this approach in general is that we have different behaviors if these properties are set via style versus the special function -- for instance, `style('position', <value>)` won't update the dataTime and will always updated the modified time, even if the passed value is the same as the previous one.  We might want to refactor the feature style method so that it is easier to have convenience functions and the behavior is identical regardless of whether the convenience function or style is used.  However, that should be a general change, and not in this PR.

manthey: The only feature that has a plain `color` style is the solid-color quad feature.  All other features have styles name `fillColor`, `strokeColor`, and probably others.  My logic was to name it with a prefix to be clear what it was (but perhaps it doesn't help).  We can rename this to `color` if you think that is better.

src/pixelmapFeature.js
manthey: The onload function is used on both a newly created Image and on an existing Image that isn't complete.  It is within the `else if` clause on line 180.

src/pixelmapFeature.js
manthey: I'll make a utility function.

src/canvas/pixelmapFeature.js
aashish24: okay, thanks.

src/pixelmapFeature.js
aashish24: its fine then. on web-page it looks as if it was outside that's why I added probably

src/pixelmapFeature.js
aashish24: heatmap uses `color`. color could be static or a transfer function.

aashish24: my preference would be slightly more towards`color`

src/pixelmapFeature.js
aashish24: I am fine with that if we are consistent, may be we can change them at some point but i think right now we use single space. Not a roadblock for this PR though.

src/pixelmapFeature.js
aashish24: > The problem with this approach in general is that we have different behaviors if these properties are set via style versus the special function -- for instance, style('position', <value>) won't update the dataTime and will always updated the modified time, even if the passed value is the same as the previous one. We might want to refactor the feature style method so that it is easier to have convenience functions and the behavior is identical regardless of whether the convenience function or style is used. However, that should be a general change, and not in this PR.

Sure, let's keep this method and then we can revisit it once we refactor style function.

aashish24: color would be better I think.

manthey: Okay.  I'll switch it to `color`.

aashish24:
> A quick grep shows 400+ uses of two spaces and 50 uses of 1 space. I think this should be left to the discretion of whomever writes comments, much like oxford commas, tight or loose punctuation, and other optional grammar rules.

sure, as I referred earlier, I am not in favor of making it part for this PR.

manthey:
@brianhelba We've renamed the `mapColor` style and function to `color`.

aashish24:
thanks @manthey. LGTM 👍

aashish24:
Nice. Talked to @jbeezley and sounds like we are ready for another point release.

PR #640, opened 2016-10-28
Remove jQuery as a Bower dependency

brianhelba:
This brings the list of Bower dependencies in line with the NPM dependencies.


aashish24:
@brianhelba commit 👏

Issue #526, opened 2016-02-03
Small lag between positions of d3 and vgl layers


manthey:
If you pan a map with both d3 and vgl layers, the two layers are briefly offset from where they should be, as if their view points are updated at different times.  They always stop with the correct alignment, but the movement looks odd.

manthey:
Fixed in PR #644.

PR #644, opened 2016-11-09
Remove the lag between layers.

manthey:
Pedantically, this wasn't a lag, but rather missed frames.

A variety of actions can trigger rerendering.  We want to perform animations before other updates.  To ensure that renderings are well ordered and not called too often, we were removing an animation frame request and then creating a new one to ensure that the callback happened at the end of the animation frame cycle.  However, sometimes this would defer the animation frame callback until a later cycle rather than just at the end of the next cycle.

By using a single animation queue, we ensure that get all of the callbacks we want in the animation cycle.  This has the fringe benefit that it has less overhead in Chrome.  Furthermore, by using a shared array for the animation queue, two map instances can be synchronized, whereas, before, there was likely to be a lag between them.

A few other minor optimizations have been made: vgl textures are built in the _build step not in the animation frame.  The vgl camera is only updated just as needed, so if two pan events happen in a single frame, only one update will occur.

Resolves issue #526.

### DIFF
--- a/src/d3/d3Renderer.js
+++ b/src/d3/d3Renderer.js
@@ -40,7 +40,6 @@ var d3Renderer = function (arg) {
       m_diagonal = null,
       m_scale = 1,
       m_transform = {dx: 0, dy: 0, rx: 0, ry: 0, rotation: 0},
-      m_renderAnimFrameRef = null,
       m_renderIds = {},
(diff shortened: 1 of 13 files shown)


aashish24:
looks reasonable to me; I thought we merged this already but I didn't +1 it.

PR #646, opened 2016-11-09
Don't pixel align if the map isn't aligned.

manthey:
If the map is not at an integer offset on the page, don't try to align texels.  This only applies to the vgl renderer.

### DIFF
--- a/src/camera.js
+++ b/src/camera.js
@@ -303,10 +303,17 @@
 
     /**
      * Getter/setter for the viewport.
+     *
+     * The viewport consists of a width and height in pixels, plus a left and
+     * top offset in pixels.  The offsets are only used to determine if pixel
+     * alignment is possible.
      */
     Object.defineProperty(this, 'viewport', {
       get: function () {
-        return {width: this._viewport.width, height: this._viewport.height};
+        return {
+          width: this._viewport.width, height: this._viewport.height,
+          left: this._viewport.left, top: this._viewport.top
+        };
       },
       set: function (viewport) {
         if (!(viewport.width > 0 &&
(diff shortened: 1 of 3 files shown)


src/camera.js
```diff
@@ -303,10 +303,17 @@
...
      * Getter/setter for the viewport.
+     *
+     * The viewport consists of a width and height in pixels, plus a left and
+     * top offset in pixels.  The offsets are only used to determine if pixel
+     * alignment is possible.
      */
     Object.defineProperty(this, 'viewport', {
       get: function () {
-        return {width: this._viewport.width, height: this._viewport.height};
+        return {
```
aashish24: why not just return the viewport object?

src/gl/vglRenderer.js
```diff
@@ -189,8 +189,11 @@ var vglRenderer = function (arg) {
       m_lastZoom = map.zoom();
       cam.setViewMatrix(view, true);
       cam.setProjectionMatrix(proj);
+      var viewport = camera.viewport;
       if (proj[1] || proj[2] || proj[3] || proj[4] || proj[6] || proj[7] ||
           proj[8] || proj[9] || proj[11] || proj[15] !== 1 || !ortho ||
+          (viewport.left && viewport.left % 1) ||
```
aashish24: some comments would be nice here on reasoning behind 1%

src/camera.js
manthey: To prevent a consumer of this from mutating the internal object.

src/gl/vglRenderer.js
manthey: Okay.  I thought it was a familiar way to check if a value was an integer.

src/camera.js
aashish24: Sure, could we not use jquery extend in that case (like we use it elsewhere?) but I am fine with this code given your explanation.

src/gl/vglRenderer.js
aashish24: for some but also it would be nice to explain the code block around it if possible.

src/camera.js
manthey: In theory, it should be more efficient to do this ourselves than use jQuery's extend, since we know what keys are of interest.

aashish24: ok, sounds good.

PR #648, opened 2016-11-16
Fix gaps in tiles when they aren't aligned.

manthey:
Non pixel-perfect tile alignment, where there are no lower level tiles, was showing slight gaps between the tiles.  The code to crop right and bottom edges was being invoked, because the texture coordinates of an unaligned tile occasionally are greater than 1.

To test this would require a new selenium test, since it is purely an artifact of how textures are rendered on quads in the vgl renderer.

### DIFF
--- a/src/gl/quadFeature.js
+++ b/src/gl/quadFeature.js
@@ -42,7 +42,7 @@ var gl_quadFeature = function (arg) {
     'uniform highp vec2 crop;',
     'void main(void) {',
     '  mediump vec4 color = texture2D(sampler2d, iTextureCoord);',
-    '  if (iTextureCoord.s > crop.s || 1.0 - iTextureCoord.t > crop.t) {',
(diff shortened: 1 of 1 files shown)


aashish24:
> The code to crop right and bottom edges was being invoked, because the texture coordinates of an unaligned tile occasionally are greater than 1.

How could it be greater than one? Is it because of some computation?

manthey:
The s/t coordinate of the texture is at the center of the fragment pixel being rendered.  When this is pixel aligned, it is well formed, and always between 0 and 1.  When it is in't pixel aligned, the renderer spans (w+1) x (h+1) pixels, with the pixels at x=0, y=0, x=w, or y=h with partial opacity (which are composited together).  The s/t values are at the centers of those pixels, which can be just over 1 based on how the texture rounding works

aashish24:
okay, thanks for the explanation. LGTM :+1:

Issue #204, opened 2014-10-05
Provide ability to draw anti-aliased lines in OpenGL


aashish24:

aashish24:
* Support other line joints as well.

manthey:
Added by PR #649.

PR #649, opened 2016-11-21
Refactor GL line rendering

manthey:
This resolves issues #204 and #596.

Previously, lines were not rendered correctly for several conditions: (a) miter locations were sometimes miscalculated, resulting in inconsistent line widths.  (b) when the miter limit was exceeded, an attempt to prevent an excessively long line produced skewed segments.  (c) On the inside of acute angles on short line segments, some of the triangles used to render lines were twisted.  (d) when lines overlap, the overlap was inconsistent.

All of this has been fixed.  Additionally, we now support:
- line caps: butt (default and what was avaulable before), round, square.  These can vary by point.
- line joins: miter (default), round, bevel, miter-clip.  These can vary by point.  Miter and miter-clip have a configurable miter limit (one value for the whole feature) which, if the miter would be longer than this, switches to using a bevel or a clipped miter.
- antialiasing on edges and end caps.  This can be specified for the whole feature.
- strokeOffset: This can vary from -1 (shift left) to 1 (shift right), where 0 is the line centered on the vertices, and 1 is with the edge of the line on the vertices.

All of this is done by rendering two triangles per line segment and performing much of the computation in the vertex and fragment shaders.

There are a few limitations to the current implementation:
- there is finite precision in the calculations, resulting in occasional artifacts along miter joins.  These are mostly only noticeable on very wide lines using stroke offsets, and, even then, are subtle.
- When the line width changes per vertex, sometimes the inside joins are not technically correct.  Instead of using the angle of the lines meeting, the angle of the edges of the lines meeting would need to be calculated.
- On short line segments, one edge of the miter can be by itself, which would ideally either be antialiased or wouldn't be trimmed.  Currently, on wide short lines with multiple segments, the mitered edge can look blocky in limited instances.
- When a line segment has zero length in screen space, it is not drawn at all.  Adjacent line segments may not be drawn exactly as expected.  Also, if line caps are used, perhaps zero-length segments should still be visible.

This method uses 18/15 as much memory as lines used before.

There is a debug flag to show all of the pixels sent to the fragment shader.  If debug is not specified when the feature is created, the debug code is not even compiled into the fragment shader.

There could be some efficiency improvements in the shaders.  For instance, it might worth it to have a quick test for simple line segments without joins.

The new features and changes are exposed in one of the selenium tests (test/selenium/glLines/?wide=true), which can take a variety of query options, such as strokeOffset, lineCap, lineJoin, miterLimit, antialiasing, strokeWidth, strokeColor, strokeOpacity, and debug.  This should be turned into an example that also tests number of lines.


manthey:
I'm marking this as a WIP until I add an example and add some additional tests.

manthey:
This is ready for review.

aashish24:
👏 👏  started the review today...hoping to get it done in next few days.

src/d3/lineFeature.js
```diff
@@ -105,6 +106,10 @@ var d3_lineFeature = function (arg) {
 
       m_renderer._drawFeatures(m_style);
     });
+    for (i = data.length; i < m_maxIdx; i += 1) {
```
aashish24: Why this is required now?

src/feature.js
```diff
@@ -31,7 +31,7 @@ var feature = function (arg) {
 
   var m_this = this,
       s_exit = this._exit,
-      m_selectionAPI = arg.selectionAPI === undefined ? false : arg.selectionAPI,
+      m_selectionAPI = arg.selectionAPI === undefined ? false : !!arg.selectionAPI,
```
aashish24: I have heard good and bad about using double negation (https://www.sitepoint.com/javascript-double-negation-trick-trouble/)

src/d3/lineFeature.js
manthey: It was always required; not having it was a bug.  In the past, if you reduced the number of lines you had in the d3 renderer, they wouldn't be taken away.

src/feature.js
manthey: We could use `Boolean()` instead.  `!!` is substantially faster than `Boolean()` in Chrome, though, and, it is a common enough javascript idiom, that I think it is fine.

manthey:
@aashish24 One question I meant to ask before: do we want to set `antialiasing` to a default value?  Currently, it defaults to `0`.  I think lines actually look best when it is `sqrt(2)` or so.

src/feature.js
aashish24: Right !! Is fast and i read about that earlier. Inam fine with !! For now

aashish24: I meant in chrome

aashish24:
+2 for a value of 2. I think thank makes sense.

src/d3/lineFeature.js
aashish24: Ah thanks

manthey:
I've pushed a change to make the default antialiasing 2.

aashish24:
Roger that @manthey working on my review

src/gl/lineFeature.js
```diff
@@ -71,53 +113,233 @@ var gl_lineFeature = function (arg) {
...
+          '  }',
+          '  float angleCB = atan(deltaCB.y / aspect, deltaCB.x);',
+          // values we need to pass along
+          '  strokeColorVar = vec4(strokeColor, strokeOpacity);',
+          // extract values from our flags field
+          '  int vertex = int(mod(flags, 4.0));',
+          '  int nearMode = int(mod(floor(flags / 4.0), 8.0));',
+          '  int farMode = int(mod(floor(flags / 32.0), 8.0));',
+          '  float offset = mod(floor(flags / 256.0), 2048.0) / 1023.0;',
+          '  if (offset > 1.0)  offset -= 2048.0 / 1023.0;',
```
aashish24: would be nice to have explanation for 2048/1023.0 division

src/gl/lineFeature.js
```diff
@@ -60,8 +95,15 @@ var gl_lineFeature = function (arg) {
...
           'uniform float aspect;',
+          'uniform float miterLimit;',
+          'uniform float antialiasing;',
 
           'varying vec4 strokeColorVar;',
+          'varying vec4 subpos;',  /* px, py, length - px, width */
+          'varying vec4 info;',  /* near mode, far mode, offset */
+          'varying vec4 angles;', /* near angle cos, sin, far angle cos, sin */
+
+          'const float PI = 3.14159265358979323846264;',
```
aashish24: perhaps we can define it some where else (utils?) so that we don't have to define it multiple times.

src/gl/lineFeature.js
```diff
@@ -71,53 +113,233 @@ var gl_lineFeature = function (arg) {
...
           '  }',
-          '  const float PI = 3.14159265358979323846264;',
-          '  vec4 worldPos = projectionMatrix * modelViewMatrix * vec4(pos.xyz, 1);',
-          '  if (worldPos.w != 0.0) {',
-          '    worldPos = worldPos/worldPos.w;',
+          /* convert coordinates.  We have four values, since we need to
+           * calculate the angles between the lines formed by prev-pos and
+           * pos-next, and between pos-next and next-far, plus know the angle
+           *   (prev)---(pos)---(next)---(far) => A---B---C---D */
+          '  vec4 A = projectionMatrix * modelViewMatrix * vec4(prev.xyz, 1);',
```
aashish24: worth using a function?

src/gl/lineFeature.js
manthey: Let's do that when we address issue #655 (in a different PR).

Alternately, we could use the value in javascript, i. e., `'const float PI = ' + Math.PI.toFixed(20) + ';'`.  We aren't guaranteed the same precision choices in javascript and WebGL, though.  The differences almost certainly don't matter much here.

src/gl/lineFeature.js
manthey: Sure.

src/gl/lineFeature.js
aashish24: sounds good to me.

PR #650, opened 2016-11-22
Fix the map jumping at the end of some transitions.

manthey:
Sometimes, when zooming, the map would jump at the end of the transition.  This was caused by the zoom level reaching its final value, then a last animation frame firing, and asking to recenter the map when the center was already correct.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -1109,7 +1109,7 @@ var map = function (arg) {
     }
 
     var defaultOpts = {
-      center: m_this.center(undefined, null),
+      center: undefined,
       zoom: m_this.zoom(),
       rotation: m_this.rotation(),
       duration: 1000,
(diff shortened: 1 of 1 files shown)


manthey:
@aashish24 I think this was the problem you saw zooming with your laptop.

aashish24:
@manthey is this what could cause the issue on some systems (like mine)? I will give it a try.

aashish24:
@manthey seems to have fixed the issue! thanks. The code looks reasonable to me. How did you reproduce this bug on your system?

manthey:
It happens much more rarely using a mouse wheel than you had with your track pad, but I noticed it on rare occasions, added some debug logging, and when it occurred, looked at what happened.

PR #654, opened 2016-12-12
When adding annotations, add a gcs options.

manthey:
New annotations should default to the interface gcs, not the map gcs.  Note that this breaks the current interface for annotationLayer.addAnnotation.

### DIFF
--- a/src/annotationLayer.js
+++ b/src/annotationLayer.js
@@ -200,14 +200,24 @@ var annotationLayer = function (args) {
    * Add an annotation to the layer.  The annotation could be in any state.
    *
    * @param {object} annotation the annotation to add.
+   * @param {string|geo.transform} [gcs] undefined to use the interface gcs,
+   *    null to use the map gcs, or any other transform
    */
-  this.addAnnotation = function (annotation) {
+  this.addAnnotation = function (annotation, gcs) {
     var pos = $.inArray(annotation, m_annotations);
     if (pos < 0) {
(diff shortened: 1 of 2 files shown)


manthey:
@danlamanna This will make it so you don't have to call `transform.transformCoordinates` when you add a new annotation.

src/annotationLayer.js
```diff
@@ -348,7 +358,7 @@ var annotationLayer = function (args) {
           state: geo_annotation.state.create,
           layer: this
         });
-        this.addAnnotation(m_this.currentAnnotation);
+        this.addAnnotation(m_this.currentAnnotation, null);
```
aashish24: it could be little confusing in the way we are using null or undefined. I would suggest that if it is null or undefined, we fall back to map.gcs(). Should we create keys perhaps? map_gcs, map_ingcs? Thoughts?

manthey: This is done throughout the other modules, so that `none` is the special value that can be used internally to mean use the `map.gcs()` while not forcing a user to constantly pass gcs parameters when what they mostly want is the `map.ingcs()`.

I would **not** make the undefined value go to `map.gcs()`, as that is certainly not what a user wants (they want the interface gcs, not the display gcs).

Using `map_gcs` instead of `none` would be okay, but this really should be a constant defined somewhere.  In which case, I lament that we don't have one `constants` module.  The user already has to import events and actions to get constants.

aashish24:
thanks @manthey

PR #656, opened 2016-12-14
Removed deprecated/unused files

aashish24:
As per the release 1.0 plan, removed unused / deprecated code.


aashish24:
Coverage is :up:

manthey:
I think you meant to remove `src/gl/geomFeature.js`, too.

aashish24:
> I think you meant to remove src/gl/geomFeature.js, too.

Yes, I missed that out in my commit. Should be fixed now.

aashish24:
@jbeezley should I generate new geo.js? Or do we that at the point of release?

manthey:
That gets done at release.

aashish24:
> thanks, that's what I thought so.

PR #658, opened 2017-01-03
Test examples

manthey:
This adds tests for the lines example and a framework for testing examples in general (issue #329).

It renders the example to a browser, then takes a screenshot of the iframe that contains the example.  The browser must be on the same machine as the karma test server for this to work.


manthey:
Our examples use bootswatch, which is dependent on fonts.googleapi.com, so these tests *require* web access.  Furthermore, the tests are brittle in that Firefox sometimes hasn't fully loaded the fonts when everything else is ready.

It would be better if we used something like google-fonts-offline and rebuilt the bootswatch sass files to point to local versions of these files.

manthey:
In the future, as we add example tests, we'll probably want to reference local map tiles rather than ones from openstreetmap.  This could be redirected to the karma server via a hosts entry in travis, for instance, but that doesn't conveniently extend to local testing.

aashish24:
> It would be better if we used something like google-fonts-offline and rebuilt the bootswatch sass files to point to local versions of these files.

+1. care to create an issue for this?

docs/developers.rst
```diff
@@ -109,6 +109,12 @@ test command or set this parameter in CMake.
     build correctly.  Try running ``ccmake /path/to/geojs`` for a full
     list of configuration options.
 
+Examples should be tested by creating an entry in the ``tests/example-cases/``
```
aashish24: +1

manthey:
By using a proxy-auto-config file and some cleverness in the karma middleware, we can avoid external web access during the example tests.  For now, I'll leave the reference to google fonts.

aashish24:
@manthey I am running into a minor issue: (via Google search I found an issue related to karma config but haven't tried it yet). Any thoughts? before I spent more time on it?

Time: 1790ms
                 Asset    Size  Chunks             Chunk Names
tests/test-examples.js  741 kB       0  [emitted]  tests/test-examples.js
chunk    {0} tests/test-examples.js (tests/test-examples.js) 264 kB [rendered]
    [0] ./tests/test-examples.js 302 bytes {0} [built]
    [1] ./tests/example-cases .*\.js$ 175 bytes {0} [built]
    [2] ./tests/example-cases/lines.js 831 bytes {0} [optional] [built]
    [3] ./~/jquery/dist/jquery.js 258 kB {0} [built]
    [4] ./tests/image-test.js 5.59 kB {0} [built]
webpack: bundle is now VALID.
10 01 2017 11:09:09.055:INFO [karma]: Karma v0.13.22 server started at http://localhost:9876/
10 01 2017 11:09:09.061:WARN [launcher]: Can not load "FirefoxWithProxy"!

manthey:
The "FirefoxWithProxy" is defined in karma-base.js and uses the tests/data/proxy-for-tests.pac file.

These worked in a Docker container without any issue, so I'm suspect your issue has something to do with either not building something or not having the correct dependencies installed.

manthey:
Can you run the ffheadless tests (that is, do the Firefox tests run that don't require a proxy)?

aashish24:
> These worked in a Docker container without any issue, so I'm suspect your issue has something to do with either not building something or not having the correct dependencies installed.

Let me do a clean build, it is possible that my environment is corrupt because of my recent experimentation

aashish24:
Much better but still seeing some issues

webpack: bundle is now VALID.
10 01 2017 11:47:15.874:INFO [karma]: Karma v0.13.22 server started at http://localhost:9876/
10 01 2017 11:47:15.891:INFO [launcher]: Starting browser Firefox
10 01 2017 11:47:22.984:INFO [Firefox 50.0.0 (Ubuntu 0.0.0)]: Connected on socket nhKBph1NJgxUDNoMAAAA with id 96354308
Firefox 50.0.0 (Ubuntu 0.0.0): Executed 0 of 3 SUCCESS (0 secs / 0 secs)
Firefox 50.0.0 (Ubuntu 0.0.0) lines example basic FAILED
	Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. in /home/chaudhary/tools/geojs/src.git/node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 1973)
	attemptAsync/timeoutId<@/home/chaudhary/tools/geojs/src.git/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1973:23

manthey:
Do you have `imagemagick` installed?  We're using its `import` function to get the screen shot for examples.

aashish24:
I thought I did but it looks like I didn't. Running again, thanks.

aashish24:
@manthey and i spent some quality time together and for some reason it looks like the test is timing out even though it actually is producing the right image. I will file a bug so that I can get back to it in the future if still relevant. Other than that it LGTM (pending lines branch).

aashish24:
Ref: https://github.com/OpenGeoscience/geojs/issues/659

Issue #431, opened 2015-08-06
Replace inline styles with a distributed CSS file


jbeezley:

manthey:
Done in PR #660.

PR #660, opened 2017-01-16
Use a css file rather than inline styles.

manthey:
This resolves issue #431.

This also includes a fix for a lag in the heatmap feature layer.   The transform needed to be set in an animation frame that is scheduled via the map's scheduling mechanism.


manthey:
This could be two separate PRs if that is preferred.

aashish24:
LRST 👍

PR #661, opened 2017-01-20
Add keyboard support for map interaction.

manthey:
Add some css to show when the map gets focus.  This is only added if the map has the appropriate class, which is controlled by a flag in the keyboard settings for the interactor.

As some background for this PR, having keyboard control is desirable for annotation editing, as it will allow precise nudging of locations.


src/mapInteractor.js
```diff
@@ -409,6 +450,120 @@ var mapInteractor = function (args) {
...
+      if (meta.hasOwnProperty(value)) {
+        if ((meta[value].shift === undefined || evt.shiftKey === !!meta[value].shift) &&
+            (meta[value].ctrl === undefined || evt.ctrlKey === !!meta[value].ctrl) &&
+            (meta[value].alt === undefined || evt.altKey === !!meta[value].alt) &&
+            (meta[value].meta === undefined || evt.metaKey === !!meta[value].meta)) {
+          factor = value;
+        }
+      }
+    }
+    console.log(JSON.stringify([meta, evt, factor]));  //DWM::
```
aashish24: Debug left out?

manthey: Yep.

PR #662, opened 2017-01-20
Blog lines examples

manthey:
Add the hidden example used for the lines blog comparison grid.

Probably the most useful piece of this is the the ability to compare different versions of geojs and other libraries.


examples/index.jade
```diff
@@ -3,7 +3,7 @@ extends ./common/templates/index.jade
 mixin thumbnail(ex)
   .col-md-4
     .thumbnail
-      a(href="#{ex.path}")
+      a(href="#{ex.path}/")
```
aashish24: I may know the reason but curious on why we had to do this?

manthey: On demo.kitware.com, we are proxying from a directory (demo.kitware.com/geojs/).  The code in the node express server that redirects directories strips off the proxy location (e.g., `demo.kitware.com/geojs/examples/lines` => `demo.kitware.com/examples/lines/`).  Rather than figure out how to better proxy node express, if we just ask for where we want to go to start with, it saves a redirect and goes to the right place.

aashish24: Okay, thanks thats what i thought

PR #664, opened 2017-01-27
Improve some errors and some coverage of map.js.

manthey:
Throw errors, not strings.

Increase tests for the map class.

The map.fileReader function was documented in the geoJSON example as taking a fileReader instance, but it actually took a string and called createFileReader.  This function now can take either and has tests.

Change where coverage reports are written and collated from.  The web-gl and examples tests both use firefox, and their coverage clobbered each other.  We aren't using the examples tests to measure coverage, but rather than allow this to happen, change the coverage directories so that they no longer conflict.  Otherwise, if the tests were run in the order ff->examples->combine coverage, we would get erroneous coverage information.

### DIFF
--- a/karma-cov.conf.js
+++ b/karma-cov.conf.js
@@ -7,12 +7,17 @@ var path = require('path');
  * Return URL friendly browser string
  */
 function browser(b) {
+  /* The browser string starts with the basic browser name (Firefox, PhantomJS,
+   * etc.  Split on the first space or dash to isolate this name. */
   return b.toLowerCase().split(/[ /-]/)[0];
 }
 
 function subdir_name(b) {
   var subdir = browser(b);
   if (process.env.GEOJS_TEST_CASE) {
+    /* Use thes test case and as part of the name so that different tests end
+     * up in different coverage directories.  Get the last file name without
+     * extenskon from the test case for this purpose. */
     var parts = /^(.+\/)*(([^\/]+)\.[^\/.]*|[^\/.]+)$/.exec(process.env.GEOJS_TEST_CASE);
     if (parts) {
       subdir += '_' + (parts[3] || parts[2]);
--- a/karma-cov.conf.js
+++ b/karma-cov.conf.js
@@ -10,16 +10,27 @@ function browser(b) {
   return b.toLowerCase().split(/[ /-]/)[0];
 }
 
+function subdir_name(b) {
+  var subdir = browser(b);
+  if (process.env.GEOJS_TEST_CASE) {
+    var parts = /^(.+\/)*(([^\/]+)\.[^\/.]*|[^\/.]+)$/.exec(process.env.GEOJS_TEST_CASE);
+    if (parts) {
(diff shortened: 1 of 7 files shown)


karma-cov.conf.js
```diff
@@ -10,16 +10,27 @@ function browser(b) {
   return b.toLowerCase().split(/[ /-]/)[0];
 }
 
+function subdir_name(b) {
```
aashish24: can we put some description on the regex?

karma-cov.conf.js
```diff
@@ -10,16 +10,27 @@ function browser(b) {
...
 
   karma_config.reporters = ['progress', 'coverage'];
   karma_config.coverageReporter = {
     reporters: [
-      {type: 'html', dir: 'dist/coverage/', subdir: browser},
-      {type: 'cobertura', dir: 'dist/cobertura/', file: 'coverage.xml', subdir: browser},
-      {type: 'json', dir: 'dist/coverage/json/', subdir: browser},
-      {type: 'lcovonly', dir: 'lcov', subdir: browser},
+      {type: 'html', dir: 'dist/coverage/', subdir: subdir_name},
+      {type: 'cobertura', dir: 'dist/cobertura/', file: 'coverage.xml', subdir: subdir_name},
```
aashish24: @jbeezley would know better.

aashish24:
Not sure what's the deal with the codecov/patch, the code looks reasonable to me but perhaps @jbeezley can weigh on the karma configuration.

karma-cov.conf.js
jbeezley: The purpose is to put the coverage for the test groups into different subdirectories.  That said, the regex could use a comment describing what it does.

jbeezley:
The patch contains changes to uncovered error conditions that don't really need test coverage.  I think we can safely ignore it.

karma-cov.conf.js
```diff
@@ -7,12 +7,17 @@ var path = require('path');
...
+   * etc.  Split on the first space or dash to isolate this name. */
   return b.toLowerCase().split(/[ /-]/)[0];
 }
 
 function subdir_name(b) {
   var subdir = browser(b);
   if (process.env.GEOJS_TEST_CASE) {
+    /* Use thes test case and as part of the name so that different tests end
+     * up in different coverage directories.  Get the last file name without
+     * extenskon from the test case for this purpose. */
```
jbeezley: *extension

Also, do you mean the last element of the path?  The comment makes it sound like the test case is an array of files.

tests/cases/mapInteractor.js
```diff
@@ -1682,4 +1682,7 @@ describe('mapInteractor', function () {
     expect(interactor.keyboard(keyboardSettings)).toBe(interactor);
     expect(map.node().hasClass('highlight-focus')).toBe(true);
   });
+  it('Test touch interaction', function () {
+    //DWM::
+  });
```
jbeezley: Did you mean to include this?

PR #665, opened 2017-02-06
Add a map.screenshot() function.

manthey:
This returns either a data url or an HTMLCanvasElement.

You do not need to preserve drawing buffers for this to work -- that doesn't alway work in Chrome in any case, so WebGL layers are rerendered as part of the screenshot process.

This needs testing.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -1546,6 +1546,68 @@ var map = function (arg) {
   };
 
   /**
+   * Get a screen-shot of all or some of the canvas layers of map.  Note that
+   * webGL layers are rerendered, even if
+   *   window.contextPreserveDrawingBuffer = true;
+   * is set before creating the map object.  Chrome, at least, may not keep the
+   * drawing buffers if the tab loses focus (and returning focus won't
+   * necessarily rerender).
+   *
+   * @param {object|array|undefined} layers: either a layer, a list of
+   *      layers, or falsy to get all layers.
+   * @param {string} type: see canvas.toDataURL.  Defaults to 'image/png'.
+   *    Alternately, 'canvas' to return the canvas element (this can be used
+   *    to get the results as a blob, which can be faster for some operations
+   *    but is not supported as widely).
+   * @param {Number} encoderOptions: see canvas.toDataURL.
+   * @returns {string}: data URL with the result.
+   */
+  this.screenshot = function (layers, type, encoderOptions) {
+    // ensure layers is a list of all the layres we want to include
+    if (!layers) {
+      layers = m_this.layers();
+    } else if (!Array.isArray(layers)) {
+      layers = [layers];
+    }
+    // filter to only the included layers
+    layers = layers.filter(function (l) { return m_this.layers().indexOf(l) >= 0; });
+    // sort layers by z-index
+    layers = layers.sort(
+      function (a, b) { return (a.zIndex() - b.zIndex()); }
+    );
+    // create a new canvas element
+    var result = document.createElement('canvas');
+    result.width = m_width;
+    result.height = m_height;
+    var context = result.getContext('2d');
+    // start with a white background
+    context.fillStyle = 'white';
+    context.fillRect(0, 0, result.width, result.height);
+    // for each layer, copy all canvases to our new canvas.  If we ever support
+    // non-canvases, add them here.  It looks like some support could be added
+    // with a library such as rasterizehtml (avialable on npm).
+    layers.forEach(function (layer) {
+      $('canvas', layer.node()).each(function () {
+        if (layer.renderer().api() === 'vgl') {
+          layer.renderer()._renderFrame();
+        }
+        var transform = $(this).css('transform');
+        // if the canvas is being transformed, apply the same transformation
+        if (transform && transform.substr(0, 7) === 'matrix(') {
+          context.setTransform.apply(context, transform.substr(7, transform.length - 8).split(',').map(parseFloat));
+        } else {
+          context.setTransform(1, 0, 0, 1, 0, 0);
+        }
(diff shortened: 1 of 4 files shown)


manthey:
@aashish24 This provides as screenshot of all canvas layers (or a subset of canvas layers).  It doesn't require preserveDrawingBuffer to be set ahead of time.

I still need to add tests and remove the functionally duplicated code in the testing utilities.

If we want to get non-canvas layers, it looks like there is a javascript library that can render html to canvas to some degree that we can look into.

aashish24:
> @aashish24 This provides as screenshot of all canvas layers (or a subset of canvas layers). It doesn't require preserveDrawingBuffer to be set ahead of time.

Cool! I will have my first pass on it today. 



> If we want to get non-canvas layers, it looks like there is a javascript library that can render html to canvas to some degree that we can look into.

that would be desirable but I am okay with if that is done in another PR.

src/map.js
```diff
@@ -1546,6 +1552,69 @@ var map = function (arg) {
...
+   *      layers, or falsy to get all layers.
+   * @param {string} type: see canvas.toDataURL.  Defaults to 'image/png'.
+   *    Alternately, 'canvas' to return the canvas element (this can be used
+   *    to get the results as a blob, which can be faster for some operations
+   *    but is not supported as widely).
+   * @param {Number} encoderOptions: see canvas.toDataURL.
+   * @returns {string|HTMLCanvasElement}: data URL with the result or the
+   *    HTMLCanvasElement with the result.
+   */
+  this.screenshot = function (layers, type, encoderOptions) {
```
aashish24: I like the API

src/map.js
```diff
@@ -1546,6 +1552,69 @@ var map = function (arg) {
...
+    layers = layers.sort(
+      function (a, b) { return (a.zIndex() - b.zIndex()); }
+    );
+    // create a new canvas element
+    var result = document.createElement('canvas');
+    result.width = m_width;
+    result.height = m_height;
+    var context = result.getContext('2d');
+    // start with a white background
+    context.fillStyle = 'white';
```
aashish24: @manthey do we not need to set the transparency? I believe the later drawImage call will update the final value to the combined value of all opacities.

src/map.js
```diff
@@ -1546,6 +1552,69 @@ var map = function (arg) {
...
+    context.fillRect(0, 0, result.width, result.height);
+    // for each layer, copy all canvases to our new canvas.  If we ever support
+    // non-canvases, add them here.  It looks like some support could be added
+    // with a library such as rasterizehtml (avialable on npm).
+    layers.forEach(function (layer) {
+      $('canvas', layer.node()).each(function () {
+        if (layer.renderer().api() === 'vgl') {
```
aashish24: Would be nice to have it not check for vgl but I understand why we have to do this.

src/map.js
manthey: The drawImage properly uses the opacity of individual pixels in the canvas, so layers composite properly.  But... I did forget that we adjust the opacity of the containing div of the layer when you set the layer opacity, so I'll need to take that into account.

src/map.js
manthey: I original thought that if we preserved the drawing buffers, this wouldn't be necessary, but they aren't preserved when the browser tab is switched in any case, so this ends up being necessary.

aashish24:
> The drawImage properly uses the opacity of individual pixels in the canvas, so layers composite properly. But... I did forget that we adjust the opacity of the containing div of the layer when you set the layer opacity, so I'll need to take that into account.

Awesome. I was wondering about that.

aashish24:
One another think I can think of is a GeoJS logo or something of that sort (it could be also used for other PR purposes as well) in the future. What you think? 

I haven't tried running the branch locally but the changes looks reasonable to me now.

aashish24:
Also a widget might be useful for screencapture (along with widget to change the basemap).

danlamanna:
> One another think I can think of is a GeoJS logo or something of that sort (it could be also used for other PR purposes as well) in the future. What you think?

You mean watermarking it? As a consumer of open source software, leaving a watermark of your brand on my images is what would make me not use the software anymore.

kotfic:
It seems like adding a widget could be handled in a separate PR?

manthey:
Adding a logo on top of stuff is tricky, as it has to be small to not obscure anything but large enough to be visible, and it has to work on a variety of background.  I'd rather see that done as a optional map widget, and then we'd just get that as part of the screenshot if was deliberately present.  It should definitely be opt-in, not opt-out.

aashish24:
> It seems like adding a widget could be handled in a separate PR?

yes, all of these features  @manthey and I are discussing is for another PR (and to add to the list of issues). 

* screenshot for non canvas layers
* wartermarking 
* widgets for screencapture / basemaps

aashish24:
> You mean watermarking it? As a consumer of open source software, leaving a watermark of your brand on my images is what would make me not use the software anymore.

@danlamanna wartermaking is not necessarily of geojs logo but for a particular project or company logo. This is something used by the PR folks mostly for presentations or for branding purposes. We can have it off by default.

aashish24:
@danlamanna please see here: http://leafletjs.com/examples/extending/watermark.html

danlamanna:
Whatever, as long as we don't force it upon the user or leave it as the default or anything like that. #stillOpposed

aashish24:
> Whatever, as long as we don't force it upon the user or leave it as the default or anything like that. #stillOpposed

+1, totally agree.

aashish24:
@manthey I was hoping (not sure if possible in pure JS environment as I haven't looked) but plottly does have this UI for downloading the plot as png. https://plot.ly/~ElPolloFrio/1911

Issue #666, opened 2017-02-10
Extend screenshot to non-canvas layers


manthey:
The map.screenshot() function only gets canvas layers right now.

To get svg layers is straightforward, BUT it will require returning a Promise rather than the screenshot right away, as there is no way to ensure synchronous loading.

It is possible to get HTML layers (and the attribution element), by rendering into an svg foreignobject, and then rendering that into a canvas.

manthey:
I tried the rasterizehtml library, but it didn't handle external image properly (so svg tile layers weren't rendered), and other svg layers appeared with an some offset.

I have a simple function that can render svg layers now, but it is asynchronous.

Is there are objection to switching the screenshot to return a promise which would then return its results?

That is, instead of 
```
  var base64PNG = map.screenshot()
```
you would have
```
  map.screenshot().then(function (result) {
    var base64PNG = result;
  });
```

manthey:
A few other possibilities:

We could trigger an event when the screenshot is ready, so that clients could listen for that instead of dealing with the promise.

We could add an option that would ensure the map was idle and rendered before taking the screenshot.

jbeezley:
I would go with both an event and a screenshot.

aashish24:
> I would go with both an event and a screenshot.

+1. so event and promise both is what I am assuming here.

jbeezley:
Yes, `s/screenshot/promise/`, it's been one of those days.

manthey:
Resolved by PR #667.

PR #667, opened 2017-02-14
Include svg layers in screenshots.

manthey:
The screenshot function now returns a Promise.  It also can take an optional wait parameter to wait until the map is idle or just until a render animation frame has occurred.  An event is triggered when the screenshot is ready.

If we need it in the future, we can issue events during the preparation phase.

Update tests for promise-based screenshots.

A few of the selenium-based tests have been moved to the headless firefox-based tests.

This address some of issue #666.  Widget (html) layers and the attribution are not yet included.


aashish24:
@manthey is this ready for review I am assuming?

manthey:
@aashish24 I had thought so.  However, yesterday I added most of the support for arbitrary layers -- widget layers (even SVG) won't screenshot properly as it currently stands as not all css gets applied properly.  I should have an update later today that resolves all of that.

manthey:
@aashish24 Ready for review.

This now fully resolves issue #666.

As a quick test in any of the examples, type the following in the console:
```
var map = $('#map').data('data-geojs-map');  // get a reference to the map
map.screenshot({wait:'idle'}).then(function (res) {  // take a screenshot
   $('body').append($('<img>').attr('src',res));  // when it is ready, add it to the page
   $('#map').css('display', 'none');  // and hide the original map
});
```

As mentioned in the last commit, screenshots of non-canvas layers will _not_ work in IE.

In any browser, fonts will look a bit different (slightly fuzzier), because they are rendered through SVG `foreignObject`, and SVG deliberately doesn't use hinting within the `foreignObject`.  There may be a way around that, but I haven't figured it out.

aashish24:
Roger that @manthey I am on it now. 

> As a quick test in any of the examples, type the following in the console:

Cool! Let me build geojs in my new laptop and I will give it a shot.

src/map.js
```diff
@@ -1547,20 +1547,46 @@ var map = function (arg) {
...
-    $a.appendTo(m_this.node());
+    /* Only add the element if there is at least one attribution */
+    if ($('span', $a).length) {
+      $a.appendTo(m_this.node());
+    }
     return m_this;
   };
 
   /**
+   * Draw a layer image to a canvas contex.  The layer's opacity and transform
```
aashish24: contex --> context?

src/map.js
```diff
@@ -1571,14 +1597,53 @@ var map = function (arg) {
...
+      if (opts.wait === 'idle') {
+        m_this.onIdle(waitForRAF);
+      } else {
+        waitForRAF();
+      }
+      return defer.then(function () {
+        return m_this.screenshot(layers, type, encoderOptions, optsWithoutWait);
+      });
+    }
+    defer = $.when();
```
aashish24: do we ever return this resolved promise (since after this line we re-assign defer)?

aashish24: nm, I see the code below.

src/util/init.js
```diff
@@ -687,6 +687,149 @@
...
+            styleElem.text(css);
+            fetch.resolve();
+          });
+          deferList.push(fetch);
+        }
+        $('head', container).append(styleElem);
+      });
+
+      $.when.apply($, deferList).then(function () {
+        var svg = $('<svg xmlns="http://www.w3.org/2000/svg"><foreignObject width="100%" height="100%"></foreignObject></svg>');
```
aashish24: @manthey should we wrap these long lines?

src/util/init.js
```diff
@@ -687,6 +687,153 @@
     },
 
     /**
+     * Escape any character in a string that has a code point >= 127.
+     *
+     * @param {string} text: the string to escape.
+     * @returns {string}: the escaped string.
+     */
+    escapeUnicodeHTML: function (text) {
```
aashish24: @manthey where were some specific unicode char you observed?

manthey: The `&copy;` (&copy;) symbol in the attribution.  Generally, the SVG foreignObject seems to handle this as `&#169;`, but not as unescaped unicode.

aashish24:
@manthey I see the HTML text is also getting blurry. Is that expected?

src/util/init.js
aashish24: okay thanks!

manthey:
Yes -- It can probably be addressed, but I'd rather defer that to another PR.

Basically, SVG `foreignObject` renders the HTML, and then we render the SVG as an image.  SVG doesn't ever use font hinting (so you don't get the pixel-targeted font variations, nor the subpixel anti-aliasing).  I don't know if rendering it at a higher resolution and down-sampling would help, or if there is a way to get it to use font-hinting.

aashish24:
> Yes -- It can probably be addressed, but I'd rather defer that to another PR.

Sounds good to me. I do not know if there is an issue for this. If not can you create one?

manthey:
Regarding font blurriness: are you on a platform that has a non-unity device scale?  For instance, on my Windows laptop, Windows is set to use 125% enlargement by default.  When Chrome or Firefox renders an image, it scales it UP so that it is 125% of the original size, which naturally makes it look blurry.  The original GeoJS render uses vector fonts at the appropriate resolution, but the generated screenshot is NOT the apparent resolution of the screen, but rather the scaled version.

This is probably the main component of whatever blurriness you are seeing.  What remains beyond that may be inconsequential (load the screenshot in something like GIMP, and you'll see that it is not the same as the resolution of the screen).

This means that we may want an option to scale by the device-pixel-ratio (or not), which would probably default to doing so.

PR #669, opened 2017-02-28
Add a screenshot button in the examples header.

manthey:
I don't know if we want this added to the examples, but it helps show off the feature, and is simpler than doing a proper screenshot widget.  If you think it is undesirable, we can get rid of the PR.

Another way to do this is to just show a save dialog with the screenshot (and not show a thumbnail of it).  In this version you can right-click on the image and copy it instead of being forced to download it.

### DIFF
--- a/examples/common/css/examples.css
+++ b/examples/common/css/examples.css
@@ -18,3 +18,14 @@ html, body {
   height: calc(100% - 60px);
   overflow: hidden;
 }
+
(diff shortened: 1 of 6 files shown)


jbeezley:
I think the feature is okay.  Just a couple minor notes:
1. The button shouldn't appear in the header on the main examples page.
2. It should use a different pointer when mousing over the element.

PR #670, opened 2017-02-28
Make it easier to disable or replace keyboard actions.

manthey:
If keyboard.actions is specified in the interactor constructor, replace the set of keyboard actions rather than modify it.  Otherwise, it is too annoying to disable or replace.


manthey:
@brianhelba As you asked for, and with a test, too.

brianhelba:
@OpenGeoscience/owners Is it possible to cut a bugfix release for this change? That would let me close out a bug in my project's use of GeoJS.

jbeezley:
@brianhelba Sure, I'll have a release out soon.

PR #676, opened 2017-03-11
Increased code-coverage for choropleth 

aashish24:


manthey:
I notice that we still have one line uncovered; that of direct creation.  This can be tested like so:
```
    it('direct creation', function () {
      var map, layer, choropleth;
      map = create_map();
      layer = map.createLayer('feature', {renderer: 'vgl'});
      choropleth = geo.choroplethFeature({layer: layer});
      expect(choropleth instanceof geo.choroplethFeature).toBe(true);
    });
```

aashish24:
> it('direct creation', function () {
>       var map, layer, choropleth;
>       map = create_map();
>       layer = map.createLayer('feature', {renderer: 'vgl'});
>       choropleth = geo.choroplethFeature({layer: layer});
>       expect(choropleth instanceof geo.choroplethFeature).toBe(true);
>     });

I was wondering about that but i realized that we don't do that in general. I can add this snippet.

Also, I realized that sometimes the GL headless tests are failing on travis. It seems random to me and related to failed download but i have not looked closely.

manthey:
In many of the unit tests we create the feature explicitly for some of the tests, so this gets tested as part of that (e.g., the polygonFeature tests).  

There are some speed tests that occasionally fail.  These are set to be pretty low, but they still sometimes fail; we could set the threshold even lower, or try to run them up to _n_ times to see if they pass.  In theory, if we make some change that affects our speed, then these should pick that up.

Also, it looks like some package was recently changed so that we may now need a more recent version of npm to run tests.

aashish24:
> Also, it looks like some package was recently changed so that we may now need a more recent version of npm to run tests.

that could be. It was trying to download a package and failed last time I saw the log

aashish24:
@manthey this is ready for review . The gl tests are failing but that is because of the other issues which I think you are taking care of in your #677 branch.

manthey:
Travis is running slowly today.  I've submitted PR #679 to fix the node version and to relax some timing requirements.  Hopefully once that passes and merges in, this will pass, too.

aashish24:
Branch is updated with recent changes from master that should fix the travis tests.

aashish24:
@manthey looks like we still have some the timing issue with glLineOpts..but I will merge this branch since it seems orthogonal to my changes as you approved it.

manthey:
Agreed.

PR #677, opened 2017-03-15
Convert all selenium tests to use xvfb.

manthey:
The d3 renderer was issuing an undefined event when it was resized.  This has been corrected to issue the d3.rescale event.

The tests in testing/test-cases/selenium-tests/featureVisible are already handled by tests/cases/feature.js, so they have been removed without conversion.

Taking a screenshot of mixed d3 and vgl layers could erroneously place the d3 layers on top.  This has been fixed.


manthey:
There may be other files we can get rid of now that selenium is no longer in use.

There are some files in the `testing` directory which probably should be moved to the `tests` directory.

manthey:
@aashish24 I'm sure some of these tests are still brittle, but I think this is ready for review.

aashish24:
+2

PR #679, opened 2017-03-15
Update the version of node used in travis.

manthey:
Some of our node packages now require a newer version of node, so make travis use it.  Our docs show us using version 4, so switch to that.

Also, relax speed test requirements to make it easier to pass on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: required
 dist: trusty
 
 node_js:
-  - '0.12'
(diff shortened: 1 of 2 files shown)


aashish24:
@manthey I did it on my #676 but I can take out the node version bump  thanks for the fixes..

aashish24:
LGTM +2 if it passes on travis (/pr is pending). Once this goes to master, I can update my #676

PR #681, opened 2017-03-28
Added line annotations.

manthey:
Added line annotations.

Rectangle annotations can be created by clicking opposite points rather than dragging the outline.  This allows them to be created on a touch device.

Added style and editstyle convenience functions to annotations -- rather than have to ask for `annotation.options('style')`, you can do `annotation.style()`.

Fix an issue with the touch handler that produced an offset to the touch event.

Fix an issue in the line feature that didn't recompute the search data when just styles were changed.

Fix an issue with the lines demo where the tooltip wouldn't always show.


src/annotationLayer.js
```diff
@@ -482,8 +499,6 @@ var annotationLayer = function (args) {
         case 'point':
           position = [feature.position()(data, data_idx)];
           break;
-        default:
```
aashish24: Cannot see in diff.. but i guess we do not need return here?

src/annotationLayer.js
```diff
@@ -652,10 +678,12 @@ var annotationLayer = function (args) {
                * hasn't been tested.
                */
               var style = {};
-              $.each(['fill', 'fillColor', 'fillOpacity', 'line', 'polygon',
-                      'position', 'radius', 'stroke', 'strokeColor',
-                      'strokeOpacity', 'strokeWidth', 'uniformPolygon'
-                  ], function (keyidx, key) {
+              $.each([
+                'closed', 'fill', 'fillColor', 'fillOpacity', 'line',
+                'lineCap', 'lineJoin', 'polygon', 'position', 'radius',
```
aashish24: should we maintain this list explicitly as some object?

src/mapInteractor.js
```diff
@@ -652,10 +652,9 @@ var mapInteractor = function (args) {
     m_touchHandler.lastEventType = evt.which;
     m_touchHandler.lastEvent = evt;
     /* convert touch events to have page locations */
-    var offset = $node.offset();
     if (evt.pageX === undefined && evt.center !== undefined && evt.center.x !== undefined) {
-      evt.pageX = evt.center.x + offset.left;
```
aashish24: No need for offset now? Or is that included now.

manthey: We just talked about, but I'll add a note for the historical record: I was mistaken in adding the offset originally, but didn't notice until I added lines.  This fixed it.

src/annotationLayer.js
manthey: There are only three conditions, so we don't need to.

Issue #674, opened 2017-03-02
Layers should have visible and selectionAPI functions


manthey:
We can toggle the visibility and interactivity of individual features with the visible() and selectionAPI() functions.  It would be nice if these were exposed for layers, so an entire layer (regardless of type) could be hidden or selections disabled (though the interaction option might only be appropriate for feature layers).

Currently, you can hide a layer by changing is opacity, but it still interacts and renders.

manthey:
A `visible()` will be added in an upcoming PR.

We currently have an `active` flag we can pass when we create a layer.  The only affect this has is whether the layer has a css class of `active`, which sets the layer's css to `pointer-events: auto` (otherwise it is `none`).

Once a layer is created, we don't have an API for changing the `active` flag (it would be trivial to have one) other than manually toggling the layer's css class.

Should a `selectionAPI` state of a layer match that of its `active` state?  For features, we disable event hooks when we turn off the selection API.  I imagine that we would cascade that down from the layer, so toggling the layer would do the same.  The `active` state, on the other hand, determines if the layer receives mouse (pointer) events itself, and is orthogonal to our selection API (since that is based on the _map_ receiving mouse events).

Or should `selectionAPI` be orthogonal to `active`?

manthey:
PR #682 adds `visible()` and `selectionAPI()` functions to layers.  If either is false, then subcomponents (such as features) act as if their values are also false.  The `active` setting on a layer can now be changed (it is independent of `selectionAPI`).

manthey:
Added in PR #682.

PR #682, opened 2017-03-28
Add a visible() and selectionAPI() functions to layers.

manthey:
Note that, much like feature level visibility, you might need to call layer.draw() after making the layer visible to ensure an update.  For feature layers, visibility cascades down to individual features (which disables their interaction).

This resolves issue #674.


aashish24:
Looking into it now :) I am surprised we didn't have this before.

src/feature.js
```diff
@@ -413,24 +413,37 @@ var feature = function (arg) {
   ////////////////////////////////////////////////////////////////////////////
   /**
    * Get/Set visibility of the feature
+   *
+   * @param {boolean|undefined} val: undefined to return the visibility, a
+   *    boolean to change the visibility.
+   * @param {boolean} direct: if true, when getting the visibility, disregard
```
aashish24: use of direct in two different ways is little confusing but I see a reason.

src/pixelmapFeature.js
```diff
@@ -339,13 +339,14 @@ var pixelmapFeature = function (arg) {
       m_quadFeature = m_this.layer().createFeature('quad', {
         selectionAPI: false,
         gcs: m_this.gcs(),
-        visible: m_this.visible()
+        visible: m_this.visible(undefined, true)
```
aashish24: it would have been nice to just call it like m_this.visible(true). Should we switch the args placeholder?

aashish24: but that means that the user would have to provide answer to direct in their call, unless they are okay with passing undefined for direct in their code.

aashish24:
I am going back and forth on direct. Typically in the past for state change / query I used bitwise operators that could provide bit more flexibility and clarification but that means that would have remember the magic does 

For now like the names enough that I +2 it.

src/feature.js
manthey: `direct` is only ever passed internally -- the user should never have to reference it.  It is only so we know what the base visibility of a feature is regardless of the layer.

src/pixelmapFeature.js
manthey: The *user* just uses `m_this.visible(true)`.  This is just done, because we don't want to change the state of visibility based on the *level*.

src/feature.js
aashish24: right but I could see a use-case (perhaps more complicated) where a user would like to know the visible state of a feature not w.r.t to its parent. But I agree that most likely they would just query the state w.r.t parent.

src/pixelmapFeature.js
aashish24: Roger that, on the same page.

aashish24:
In summary: 

* User would mostly call visible(true)
* When they really want to ensure they would call visible(true, true) [in this case I would have preferred `force` or `override` as arg names but `direct` works as well. 

+2

manthey:
The user really shouldn't ever call `visible(true, true)`.  Since the only time that `visible(true)` on a feature doesn't do something is if the layer is hidden, then when the layer's visibility changes, the *layer* uses the `visible(..., true)` to ensure that the interface hooks are enabled.  If the layer's visibility isn't changing, adding the direct flag doesn't do anything useful when setting the feature visibility.  When fetching the feature's visibility, it does allow telling whether the feature would be visible IF the layer is visible.

PR #684, opened 2017-03-30
Add unit tests for the point feature.

manthey:
Fixed some issues with the pointFeature:
- Setting the position on construction resulted in different behavior than setting it afterwards.
- Setting the clustering on construction allowed non-default values, but during the setter would only allow defaults.
- If set, the clustering could not be changed except by unsetting it and resetting it.
- The maximum radius of any point is used when deciding where to search for points, but it was potentially not calculated until after it was used.

Made some improvements:
- The position function wraps the position in another function to handle clustering.  If the position is not a function, then only one function call is made rather than two.
- Removed the unused internal _boundingBox function from the pointFeature.
- Added some comments about when you would want to use a different primitiveShape with the gl point feature.

### DIFF
--- a/src/gl/pointFeature.js
+++ b/src/gl/pointFeature.js
@@ -38,6 +38,11 @@ var gl_pointFeature = function (arg) {
       m_pixelWidthUniform = null,
       m_aspectUniform = null,
       m_dynamicDraw = arg.dynamicDraw === undefined ? false : arg.dynamicDraw,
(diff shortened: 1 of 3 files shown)


manthey:
I think this PR illustrates why we want to have _unit_ tests on everything -- I discovered a few bugs and inconsistencies while writing them that weren't exposed in our various test cases.

PR #685, opened 2017-03-30
Replace vendored wigglemaps with kdbush

jbeezley:
Depends on #684.


aashish24:
🎉 We have not 92+% coverage, thanks @jbeezley

Issue #683, opened 2017-03-29
Update styles from arrays


manthey:
Previously, we've animated line and point features by getting access to the WebGL `Float32Array`s used for specific styles, such as opacity or color, modified those arrays, and then rerendered.  This is very fast, but requires knowledge of internal data structures.  For unclustered points, for instance, each point has a a record that has some number of vertices per point (which could be 1, 3, or 4 depending on how we constructed the point feature).  The user has to know to use this value.

As an alternative, if we added a function to set styles from arrays, the generic mode would just update that style by converting the array to a function:
```
this.styleFromArray = function (key, styleArray, refresh) {
  this.style(key, function (d, i) {
    return styleArray[i];
  });
  if (refresh && this.visible()) {
    this.draw();
  }
  return(this);
}
```
The `styleArray` contents in this example would be _not_ be functions.  This means for lines and polygons, there would be one value for each line or polygon regardless of how many line segments or vertices they had.

Perhaps this could also take a dictionary of style keys and their associated arrays as a convenience to avoid having to call the function multiple times for multiple concurrent style changes.

For features where we chose to optimize things, the subclass could provide an extra method to accelerate updates.  For instance, in the gl_pointFeature, this function could be:
```
s_styleFromArray = <parent>.styleFromArray;
this.styleFromArray = function (key, styleArray, refresh) {
  if (this.visible() {
    var vpf, mapper, buffer, numPts, value, i, j, v;
    switch (key) {
      case 'fillOpacity': case 'strokeOpacity':
        numPts = this.data().length;
        mapper = this.actors()[0].mapper();
        buffer = mapper.getSourceBuffer(key);
        vpf = this.verticesPerFeature();
        if (!buffer || !numPts || numPts * vpf !== buffer.length) {
          break;
        }
        for (i = 0, v = 0; i < numPts; i += 1) {
          value = styleArray[i];
          for (j = 0; j < vpf; j += 1, v += 1) {
            buffer[v] = value;
          }
        }
        mapper.updateSourceBuffer(key);
        // I'll have to check if this causes a refresh or not, but we don't need to refresh from the superclass call
        refresh = false;
        break;
    }
  }
  // we always need to update the style to keep the overall state consistent.
  return s_styleFromArray(key, styleArray, refresh);
}
```

[...]

PR #687, opened 2017-04-05
Update styles from arrays.

manthey:
This allows a convenient way to animate across styles, and also provides a way where, if appropriate, rerendering can be optimized based on just what has changed.

Per-vertex styles can be set on lines and polygons by passing an array of arrays.

Some styles on points and contours were being queried via functions of the form styleFunc(data) where they should have been queried as styleFunc(data, index).  These have been changed.

Speed up paths for other features will be added later.


manthey:
This resolves issue #683.

Optimizations for lines and polygons can be done later.  Optimizations for positions will take more work.

src/feature.js
```diff
@@ -356,6 +360,67 @@ var feature = function (arg) {
 
   ////////////////////////////////////////////////////////////////////////////
   /**
+   * Set style(s) from array(s).  For each style, the array should have one
+   * value per data item.  The values are not converted or validated.  Color
+   * values should be objects with r, g, b values on a scale of [0, 1].  If
+   * invalidate values are given the behavior is undefined.
+   *   For features where this._subcomponentStyles is an object and a style
```
aashish24: Since this is a public API, listing private API (from documentation point of view) could be confusing to the user.

src/feature.js
```diff
@@ -356,6 +360,67 @@ var feature = function (arg) {
...
+  ////////////////////////////////////////////////////////////////////////////
+  this.updateStyleFromArray = function (keyOrObject, styleArray, refresh) {
+    if (typeof keyOrObject !== 'string') {
+      $.each(keyOrObject, function (key, value) {
+        m_this.updateStyleFromArray(key, value);
+      });
+    } else {
+      var fallback;
+      if (keyOrObject.toLowerCase().match(/color$/)) {
+        fallback = {r: 0, g: 0, b: 0};
```
aashish24: Should be define these defaults as someplace else where it is more visible?

src/gl/pointFeature.js
```diff
@@ -60,7 +62,7 @@ var gl_pointFeature = function (arg) {
     '  precision highp float;',
     '#endif',
     'attribute vec3 pos;',
-    'attribute float rad;',
+    'attribute float radius;',
```
aashish24: +1

src/feature.js
```diff
@@ -44,6 +44,10 @@ var feature = function (arg) {
       m_dependentFeatures = [],
       m_selectedFeatures = [];
 
+  // subclasses can add keys to this for styles that apply to subcomponents of
+  // data items, such as individual vertices on lines or polygons.
+  this._subcomponentStyles = {};
```
aashish24: should we just call it componentStyles? or perFeatureStyle? sub and component in my understanding convey very similar message.

src/feature.js
manthey: I'll improve the comment.

src/feature.js
manthey: It really is intended to be part of the private API, so it should only matter to developers.  This is the list of styles that are per-vertex rather than per-feature when there is a distinction between vertices and features (i.e., not on point features).  There could be a feature where the sub-feature is not a vertex but something else, so this isn't strictly a list of vertex styles.  I don't know what a better name would be.  `perFeatureStyle` is definitely wrong, as these apply at a finer granularity than features.  `_subfeatureStyles` might be better.  `_stylesAtFinerGranularityThanPerFeature` is too long.

src/feature.js
manthey: This is the only one, because colors are always coerced into the rgb object and other functions except them not to have undefined values.  Moving it elsewhere will just obscure what happens, I think.

src/feature.js
aashish24: I like _subfeatureStyles 👍

src/feature.js
aashish24: Sure I am fine with being here for now and we can revisit if needed to be in the future.

PR #688, opened 2017-04-26
Move mock VGL functions to util

manthey:
This will make in convenient for other projects to test geojs in non-webgl browsers.

Also, this now allow multiple contexts to be mocked.


manthey:
I expect the travis tests to fail since Firefox 53 has been released.  It has changed something in the button padding that is messing things up.

jbeezley:
I'm okay with the changes.  Do you need to update the screenshots for Firefox 53?

PR #696, opened 2017-05-02
Better checks for external modules.

manthey:
D3 and HammerJS are intended to be optional modules.  Consumers of geojs (at least those that don't use webpack) improperly report that these modules are always available.

### DIFF
--- a/src/d3/d3Renderer.js
+++ b/src/d3/d3Renderer.js
@@ -635,7 +635,11 @@ registerRenderer('d3', d3Renderer);
    * @returns {boolean} true if available.
    */
   d3Renderer.supported = function () {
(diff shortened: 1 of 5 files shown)


jbeezley:
We might also want to change the public symbol for hammerjs in the webpack config (https://github.com/OpenGeoscience/geojs/blob/master/webpack.config.js#L46).  If you include hammerjs as a global, it is exported to `window.Hammer`.

manthey:
@jbeezley Is the change I made sufficient?  Is there some sensible way to add a test for gracefully not using Hammer if it isn't available?

jbeezley:
The change looks good.  I don't know of any good ways of testing it.

PR #698, opened 2017-05-19
Update the documentation in src/map.js.

manthey:
Only the first 1/3 has been done, mainly to get feedback if this is what we want.

I've removed a few bits of vestigial code -- map.scale and map.maxZoom were unused.  The pan event claimed it included more information than it did, and included an extra empty element.

We could get rid of the map.origin() code -- we don't use it or test it other than simple coverage.

Perhaps we should have a more central place to document datatypes used in the documentation -- many are in the mapInteractor code.


manthey:
@jbeezley and @aashish24 Here is my first take on trying to update map.js's documentation.  I've only done the first 1/3 of the file or so, since I want to make sure we are happy with how things are being done.  I have questions about how jsdoc generates event documentation (see my comment on issue #448).

manthey:
I've gone through all of src/map.js.  Criticism welcome.

src/map.js
```diff
@@ -523,17 +520,19 @@ var map = function (arg) {
...
+   *    center, it is converted from this gcs to the map projection.  The
+   *    returned center is converted from the map projection to this gcs.
+   * @param {boolean} [ignoreDiscreteZoom] if true, ignore the discreteZoom
    *    option when determining the new view.
-   * @param {boolean} ignoreClampBounds if true or 'limited', ignore the
-   *    clampBoundsX options (up to a point, see fix_bounds) when determining
-   *    the new view.
-   * @returns {Object|geo.map}
+   * @param {boolean|'limited'} [ignoreClampBounds] if true ignore the
+   *    clampBounds* options when determining the new view.  When 'limited',
```
aashish24: why star after clampBounds?

src/map.js
```diff
@@ -596,8 +597,9 @@ var map = function (arg) {
   /**
    * Remove layer from the map
    *
-   * @param {geo.layer} layer that should be removed from the map
-   * @return {geo.map}
+   * @param {geo.layer?} layer layer that should be removed from the map
```
aashish24: was ? intended?

src/map.js
```diff
@@ -835,7 +819,9 @@ var map = function (arg) {
 
   ////////////////////////////////////////////////////////////////////////////
   /**
-   * Manually force to render map
+   * Redraw the map and all its layers.
```
aashish24: +1

src/map.js
manthey: shortened form of `clampBoundsX and clampBoundsY options`.  Obviously not clear enough, so I shouldn't use the shortened form

aashish24:
@jbeezley @manthey should we have a style doc somewhere that can summarize @jbeezley suggestions?

src/map.js
```diff
@@ -865,12 +853,13 @@ var map = function (arg) {
...
-   * @param {string|object|undefined} readerOrName: undefined to get the
-   *    current reader, an instance of a file reader to set the reader, or a
-   *    name to create a file reader (see utils.createFileReader for options).
-   * @param {object} opts: options for creating a file reader.  If this
-   *    includes layer, use that layer, otherwise create a layer using these
-   *    options.
+   * @param {string|object} [readerOrName]: undefined to get the current
+   *    reader, an instance of a file reader to set the reader, or a name to
+   *    create a file reader.
+   * @param {object} [opts]: options for creating a file reader when the reader
```
jbeezley: It doesn't seem to change the output here, but for consistency it would be good to remove the `:` in these declarations.

src/map.js
```diff
@@ -667,12 +669,13 @@ var map = function (arg) {
 
   ////////////////////////////////////////////////////////////////////////////
   /**
-   * Resize map (deprecated)
+   * Resize map in pixels.
```
manthey: We use this function when the window is resized via the `resizeSelf` function.  We don't do *anything* with x and y (just with width and height).

Since prior to this PR, it has been labelled deprecated for a while, we could remove the function and roll the needed functionality into `resizeSelf`.  We could remove x and y from the `geo.event.resize` event properties, too.

src/map.js
```diff
@@ -357,14 +341,18 @@ var map = function (arg) {
   /**
    * Get/Set zoom level of the map
    *
-   * @param {number} val if undefined, return the current zoom level.
+   * @param {number} [val] if undefined, return the current zoom level.
```
manthey: Let us do that, too.  I'd say the only exception is when the first word is a literal, in which case it can be in monospace.

src/map.js
```diff
@@ -412,19 +400,22 @@ var map = function (arg) {
...
-   * @param {boolean} ignoreClampBounds if true or 'limited', ignore the
-   *    clampBoundsX options (up to a point, see fix_bounds) when determining
-   *    the new view.
-   * @returns {geo.map}
-   * @fires geo.event.event:pan
+   * @param {boolean|'limited'} [ignoreClampBounds] if true ignore the
+   *    clampBoundsX and clampBoundsY options when determining the new view.
+   *    When 'limited', the clampBoundsX and clampBoundsY options are
+   *    selectively enforced so that the map will not end up more out of bounds
+   *    than its current state.
```
manthey: Yes.

src/map.js
manthey: I meant to remove the `:`.

manthey:
@jbeezley Short of refactoring or removing `resize`, I've addressed your comments.  I'll not that `resize` is also called from the `size` method, so it might make more sense to move its code there and have `resizeSelf` call `size`.

aashish24:
@manthey I am fine with changes, although I wish we had a doc style checker. Manually checking doc changes seems harder than it looks. 

@jbeezley are we still waiting on changes?

src/map.js
```diff
@@ -1726,19 +1743,22 @@ var map = function (arg) {
...
+   * synchronization.
    *
-   * @param {function} callback function to call during the animation frame.
-   *    It is called with an animation epoch, exactly as requestAnimationFrame.
-   * @param {boolean|'remove'} [action=false] falsy to only add the callback if
-   *    it is not already scheduled.  'remove' to remove the callback (use this
-   *    instead of cancelAnimationFrame).  Any other truthy value moves the
-   *    callback to the end of the list.
-   * @returns {integer} An integer as returned by window.requestAnimationFrame.
+   * @param {function} Xallback function to call during the animation frame.
```
jbeezley: Typo

src/map.js
```diff
@@ -969,15 +981,15 @@ var map = function (arg) {
...
    * Get or set the min/max zoom range.
    *
-   * @param {object} [arg] The zoom range
-   * @param {number} [arg.min] The minimum zoom level
-   * @param {number} [arg.max] The maximum zoom level
-   * @param {boolean} [noRefresh] if true, don't update the map if the zoom
+   * @param {object} [arg] The zoom range.
+   * @param {number} [arg.min] The minimum zoom level.
+   * @param {number} [arg.max] The maximum zoom level.
+   * @param {boolean} [noRefresh] if `true`, don't update the map if the zoom
```
jbeezley: Capitalize and end in a period.

src/map.js
```diff
@@ -1750,7 +1750,7 @@ var map = function (arg) {
    * callbacks will be called in a single time slice, providing better
    * synchronization.
    *
-   * @param {function} Xallback function to call during the animation frame.
+   * @param {function} Callback function to call during the animation frame.
```
jbeezley: The parameter name is "callback", so this should be "callback Function".

aashish24:
:tada:

PR #700, opened 2017-06-01
Fix a bug with polygon stroke styles.

### LINKED ISSUE #699: Stroke style is wrong on polygons with holes
A simple reproducible case in geojson:
```
{
  "geometry":{
    "coordinates":[
      [[-90,30],[-90,40],[-80,40],[-80,30],[-90,30]],
      [[-87,33],[-87,37],[-83,37],[-83,33],[-87,33]]
    ],
    "type":"Polygon"
  },
  "properties":{"strokeColor":"black"},
  "type":"Feature"
}
```
If pasted into the geojson example, the hole does not have the same stroke as the outside of the polygon.  Rather, it has the default stroke style.

manthey:
If a polygon has a hole, then stroke styles were misapplied to polygons.  This was caused by there being more than one stroke per polygon and the stroke styles being called with polygon indices.

Fixes #699.

### DIFF
--- a/src/polygonFeature.js
+++ b/src/polygonFeature.js
@@ -151,6 +151,32 @@ var polygonFeature = function (arg) {
 
   ////////////////////////////////////////////////////////////////////////////
   /**
+   * Get the style for the stroke of the polygon.  Since polygons can have
+   * holes, the number of stroke lines may not be the same as the number of
+   * polygons.  If the style for a stroke is a function, this calls the
+   * appropriate value for the polygon.  Any style set for a stroke line should
+   * be wrapped in this function.
+   *
+   * @param {(object|function)?} styleValue The polygon's style value used for
+   *    the stroke.  This should be m_this.style(<name of style>) and not
+   *    m_this.style.get(<name of style>), as the result is more efficient if
+   *    the style is not a function.
+   * @returns {object|function} A style that can be used for the stroke.
+   * @private
+   */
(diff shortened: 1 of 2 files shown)


src/polygonFeature.js
```diff
@@ -151,6 +151,32 @@ var polygonFeature = function (arg) {
...
   /**
+   * Get the style for the stroke of the polygon.  Since polygons can have
+   * holes, the number of stroke lines may not be the same as the number of
+   * polygons.  If the style for a stroke is a function, this calls the
+   * appropriate value for the polygon.  Any style set for a stroke line should
+   * be wrapped in this function.
+   *
+   * @param {(object|function)?} styleValue The polygon's style value used for
+   *    the stroke.  This should be m_this.style(<name of style>) and not
+   *    m_this.style.get(<name of style>), as the result is more efficient if
```
aashish24: I have a feeling that this comment can be made more readable. Also when we refer to get, are we talking about Map.prototype.get()?

manthey: Our style function can either be called like `style('strokeColor')`, which returns the value set for strokeColor, or as `style.get('strokeColor')`, which *always* returns a function for strokeColor.  Here, if we know that the value is *not* a function, then we can pass it along to the stroke, whereas if it *is* a function, we have to get the appropriate value for the polygon by calling that function with the polygon's values.

PR #702, opened 2017-06-05
Remove the map.resize method.

manthey:
We marked the `map.resize` method as deprecated in version 0.6 (commit 27e6b36).  We completely ignore the `x` and `y` coordinates passed to it.  This change removes the resize and the internal `m_x` and `m_y` variables, moving the necessary functionality to the size method.

This also modifies the resize event by removing the `x` and `y` parameters.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -99,8 +99,6 @@ var map = function (arg) {
       s_exit = this._exit,
       // See https://en.wikipedia.org/wiki/Web_Mercator
       // phiMax = 180 / Math.PI * (2 * Math.atan(Math.exp(Math.PI)) - Math.PI / 2),
-      m_x = 0,
-      m_y = 0,
       m_node = $(arg.node),
       m_width = arg.width || m_node.width() || 512,
       m_height = arg.height || m_node.height() || 512,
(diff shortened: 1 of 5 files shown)


manthey:
I'm not committed to this change -- but we should either remove the deprecated function or address the fact we ignore the x and y parameters.

src/map.js
```diff
@@ -649,7 +647,30 @@ var map = function (arg) {
         height: m_height
       };
     }
-    m_this.resize(0, 0, arg.width, arg.height);
+    // store the original center and restore it after the resize
+    var oldCenter = m_this.center();
```
aashish24: should we make this a internal call _restoreView() or something?

manthey: Unless there is some reason to call `size` without some of the side effects (emitting a `resize` event and marking the map as modified), I don't see any benefit to calling another function here.

aashish24: perhaps my intentions were not clear. I was wondering if there is a strong reason to wrap these few lines of code as internal method. If not, then I am good with these changes :+1:

aashish24: @manthey thoughts? As mentioned earlier, I am good with these changes (please have a look at my previous comment).

manthey: I don't think this needs to be in an internal method.  It is specific to resizing, and that can all go through the `size` method.

aashish24: okay sounds good to me.

PR #704, opened 2017-06-07
Update the documentation on events.

manthey:
Remove `type` as a field in events, as we usually use `event` (and `event` is always present).

Make some events more consistent.

This also updates documentation in typedef.js and object.js.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -3,8 +3,8 @@
  * Common object containing all event types that are provided by the GeoJS
  * API.  Each property contained here is a valid target for event handling
  * via {@link geo.object#geoOn}.  The event object provided to handlers is
- * different for each event type.  Each handler is generally be called
- * with the <code>this</code> context being the class that caused the event.<br>
+ * different for each event type.  Each handler is generally called with the
+ * `this` context being the class that caused the event.<br>
  * <br>
  * The following properties are common to all event objects:
  *
@@ -143,7 +143,7 @@ geo_event.mousemove = 'geo_mousemove';
 //////////////////////////////////////////////////////////////////////////////
 /**
  * Triggered on `mouseup` events that happen soon enough and close enough to a
- * mousedown event.  The event object extends {@link geo.mouseState}.
+ * `mousedown` event.  The event object extends {@link geo.mouseState}.
  *
  * @event geo.event.mouseclick
  * @property {geo.mouseButtons} buttonsDown The buttons that were down at the
@@ -428,7 +428,7 @@ geo_event.pixelmap = {
    *
    * @event geo.event.pixelmap.prepared
    * @type {object}
-   * @property {geo.pixelmapFeature} pixelmap The pixelamp object that was
+   * @property {geo.pixelmapFeature} pixelmap The pixelmap object that was
    *    prepared.
    */
   prepared: 'geo_pixelmap_prepared'
--- a/src/event.js
+++ b/src/event.js
@@ -3,15 +3,16 @@
  * Common object containing all event types that are provided by the GeoJS
  * API.  Each property contained here is a valid target for event handling
  * via {@link geo.object#geoOn}.  The event object provided to handlers is
- * different for each event type.  Each handler will generally be called
- * with a the <code>this</code> context being the class that caused the event.<br>
+ * different for each event type.  Each handler is generally be called
+ * with the <code>this</code> context being the class that caused the event.<br>
  * <br>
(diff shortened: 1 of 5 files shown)


manthey:
@jbeezley Is there a better way to show the data that is passed to an event?  Currently, the geo.event lists some properties at the top, then each event lists properties specific to that event; it may not be obvious to the new user that the `event` and `geo` properties always exist.

Also, the geo.event.feature events should reference the main events, but I'm not sure how to get jsdoc to create the appropriate link.

jbeezley:
I don't have any specific advice unfortunately.  I suspect we will need to customize the template or create a jsdoc plugin.  I can spend some time on it in a different PR.

manthey:
This is a question for future documentation work:  The map class and some features, such as contour, take an object with many parameters.  I think these really should be specified using `@typedef` (currently, in map.js it is just listed as a long set of options in the class specification)?   If we do so, should the typedefs be moved to `typedef.js` or left in the appropriate feature?  In the map, this means that the class specification would just be `@param {geo.mapSpec} arg The map specification object`.

src/event.js
```diff
@@ -3,15 +3,16 @@
  * Common object containing all event types that are provided by the GeoJS
  * API.  Each property contained here is a valid target for event handling
  * via {@link geo.object#geoOn}.  The event object provided to handlers is
- * different for each event type.  Each handler will generally be called
- * with a the <code>this</code> context being the class that caused the event.<br>
+ * different for each event type.  Each handler is generally be called
```
jbeezley: `s/be //`

src/event.js
```diff
@@ -3,15 +3,16 @@
...
+ * with the <code>this</code> context being the class that caused the event.<br>
```
jbeezley: Might as well just use markdown syntax instead of `<code>`.

src/event.js
```diff
@@ -167,25 +125,29 @@ geo_event.draw = 'geo_draw';
...
 //////////////////////////////////////////////////////////////////////////////
 geo_event.mousemove = 'geo_mousemove';
 
 //////////////////////////////////////////////////////////////////////////////
 /**
- * Triggered on every 'mousedown' over the map's DOM element.  The event
- * object extends {@link geo.mouseState}.
- * @mixes geo.mouseState
+ * Triggered on `mouseup` events that happen soon enough and close enough to a
+ * mousedown event.  The event object extends {@link geo.mouseState}.
```
jbeezley: <pre>`mousedown`</pre>

src/event.js
```diff
@@ -404,31 +422,40 @@ geo_event.feature = {
...
 geo_event.pixelmap = {
-  /* The image associated with the pixel map url has been prepared and rendered
-   * once. */
+  /**
+   * Report that an image associated with a pixel map has been prepared and
+   * rendered once.
+   *
+   * @event geo.event.pixelmap.prepared
+   * @type {object}
+   * @property {geo.pixelmapFeature} pixelmap The pixelamp object that was
```
jbeezley: pixelmap

src/typedef.js
```diff
@@ -81,44 +89,124 @@
...
- * brush action
+ *      brush action.
+ */
+
+/**
+ * The conditions that are necessary to make an action occur.
+ *
+ * @typedef geo.actionRecord
+ * @type {object}
+ * @property {string} action The name of the action, from geo.action.
```
jbeezley: `{@link geo.action}`

src/typedef.js
```diff
@@ -81,44 +89,124 @@
...
+ * @property {string} [name] A human-readable name that can be used to locate
+ *      or filter actions.
+ * @property {string|object} input The name of an input that is used for the
+ *      action, or an object with input names as keys and boolean values of
+ *      inputs that are required to occur or required to not occur to trigger
+ *      the action.  Input names include `left`, `right`, `middle` (for mouse
+ *      buttons), `wheel` (the mouse wheel), `pan` (touch pan), `rotate` (touch
+ *      rotate).
+ * @property {string|object} [modifiers] The name of a modifier key or an
+ *      object withe modifiers as the keys and boolean values.  The listed
```
jbeezley: typo

src/event.js
```diff
@@ -194,7 +156,8 @@ geo_event.mouseclick = 'geo_mouseclick';
 /**
  * Triggered on every 'mousemove' during a brushing selection.
```
jbeezley:
```
`mousemove`
```

PR #706, opened 2017-06-12
Update eslint and enable linting of docstrings

jbeezley:
This updates to eslint v3.  Technically, v4 is out as of yesterday, but our peer dependencies haven't been updated yet.  There are a number of new rules (mostly whitespace) that needed to be fix.  Thankfully most of the changes only required running `eslint --fix .`.

The main purpose of this PR is to add linting of our docstrings.  The rules I choose are designed to enforce the style of the docstrings added in `map.js` in #698.  I have the rules set to warnings for now until we get around to fixing them all, which could take a while.

I also made a PR against eslint to add enforcement of parameter and return value descriptions (https://github.com/eslint/eslint/pull/8716).  If and when this is merged, we can enforce our docstring conventions there as well.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,19 +1,19 @@
 {
     "extends": "semistandard",
     "rules": {
-        "indent": [2, 2, {"VariableDeclarator": 2, "SwitchCase": 1}],
-        "padded-blocks": 0,
-        "space-before-function-paren": [2, {"anonymous": "always", "named": "never"}],
-        "one-var": 0,
-        "no-throw-literal": 0,
-        "new-cap": 0,
-        "spaced-comment": 0,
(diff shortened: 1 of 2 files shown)


jbeezley:
Now that I look at it again, I think I'll change some of the new options to reduce the number of changes here.  Rebase incoming.

jbeezley:
I found there are some big changes to how indentation is handled in 4.0.  It also provides more flexibility in configuration.  I reverted the eslint update for now.  I'll make another pass at it once our peer dependencies are all compatible with 4.0.

jbeezley:
I'm getting all sorts of random errors on travis about missing node modules during the build.  I had tried to upgrade npm, but that didn't seem to help, so I reverted that change.

PR #709, opened 2017-06-14
Improve the documentation of feature, contourFeature, and util/index.

manthey:
Increase what is accepted by the css color parser (noticed while documenting colors).

Reorder some of the code in util/index.js to put similar things together.


aashish24:
@manthey looks like this needs updating?

manthey:
@aashish24 It already has master merged in.  The codecov complaints are from the first push before I changed the rules to allow small deviations; I don't know how to get codecov to update its comment on the checks without deleting and recreating the PR.  I have another change set that builds on this one, that I would rather see as a separate PR than conflated with this one.  This PR can be reviewed and merged as is.

manthey:
@aashish24 With recent changes to our codecov process, it now passes all checks.

PR #713, opened 2017-06-16
Remove rows of slashes.

### LINKED ISSUE #712: eslinting of jsdoc rules isn't always applied
Currently, we are only throwing warnings for jsdoc linting errors.  However, we aren't throwing all the warnings we should be.  Specifically, when we have placed a comment between a function or class jsdoc comment and the function or class, eslint doesn't check that line.  We often do this like so (this won't trigger a warning):
```
  /**
   * Initialize
   *
   * @param {geo.contourFeature.spec} arg The contour feature specification.
   */
  this._init = function (arg) {
```

If the lines of 
```
```
are removed, so we have
```
  /**
   * Initialize
   *
   * @param {geo.contourFeature.spec} arg The contour feature specification.
   */
  this._init = function (arg) {
```

eslint properly checks and throws the expected warning.

We either need to adjust how eslint parses this, or just remove the lines of slashes when they precede function or class definitions (which would be my preference).

manthey:
They confuse eslint.

In some instances, we used block comments of the form
```
///////////
// comment
///////////
```
These have been changed to
```
/*
 * comment
 */
```

Fixes #712.


manthey:
The eslint jsdoc warnings went from 163 to 893...

PR #715, opened 2017-06-22
Increase options for tile url templates.

manthey:
Before, we would translate `{x}` -> tile x, `{y}` -> tile y, `{z}` ->
tile layer, and `{s}` -> one of the subdomains defined in the layer.
This makes these case insensitive, supports an options `$` prefix, and
allows several additional subdomain specifications.  Specifically,
`{s:abc}`, `{S:abc}`, `{a-c}`, and `{a,b,c}` are all equivalent.  If the
subdomain is in the template string, the layer's subdomain list is
ignored.

When we refactor the documentation for the tileLayer, this information
should be included in the subdomain and template string docs.


manthey:
I refactored the code so that the regex and subdomains are only parsed once when the url is set instead of each time a tile is requested.

PR #718, opened 2017-07-10
Finish removing event `type` properties.

### LINKED ISSUE #703: Event property type is not used consistently
In `src/events.js`'s documentation, we claim that all events have a property called `type` which contains the string that identifies the event.  However, in `src/object.js`, we store the event identifier in the event object's `event` property.  We *also* use the `type` property on a handful of events from `map.js`: `layerAdd`, `layerRemove`, `resize`, `draw`, `drawEnd`.

I think we should remove the `type` property, but that would be a breaking change.  Alternately, we could assign the event identifier to *both* the `event` and `type` properties.

manthey:
We store the event name in the `event` proeprty of an event.  In a few
places were also storing it in the `type` property.  The `type` property
wasn't used consistently and has mostly been removed.  This finishes
removing it.

This closes #703.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -682,7 +682,6 @@ var map = function (arg) {
     m_this.center(oldCenter);
 
     m_this.geoTrigger(geo_event.resize, {
-      type: geo_event.resize,
       target: m_this,
       width: m_width,
       height: m_height


aashish24:
Nice 🎉

PR #719, opened 2017-07-17
Annotation labels

manthey:
Support annotation labels.

This adds a new feature type, `geo.textFeature`, where the data items have a position, text, and styling.  Currently this is only available for the canvas renderer.  Annotations now have `label`, `showLabel`, `labelStyle`, and `description` properties.  The annotation layer has a `showLabels` option for toggling all labels.  This supports rotation, scaling, offset, shadows, and strokes on text features.

The annotation example exposes all of these options.

Documentation and tests are fully updated.


src/textFeature.js
```diff
@@ -0,0 +1,207 @@
+var inherit = require('./inherit');
```
aashish24: the API is nice @manthey

src/canvas/textFeature.js
```diff
@@ -0,0 +1,214 @@
...
+
+  arg = arg || {};
+  textFeature.call(this, arg);
+  object.call(this);
+
+  /**
+   * @private
+   */
+  var m_this = this,
+      m_defaultFont = 'bold 16px sans-serif',
```
aashish24: should we use font size 12? as default which I have seen is more common but perhaps 12 is too small?

manthey: I picked `bold 16px sans-serif` because I thought it was a good default for the annotations.  Generally, I think a text feature will be used for labels not blocks of text, so a larger size is more likely to be appropriate.  I could be persuaded otherwise.

manthey:
Do we want `showLabels` to default to `true`?  If so, anything that currently uses annotations that switches to the new code will suddenly have labels, which could either be good or bad depending on what that program expects.

src/annotation.js
```diff
@@ -1229,18 +1444,33 @@ lineRequiredFeatures[lineFeature.capabilities.basic] = [annotationState.create];
 registerAnnotation('line', lineAnnotation, lineRequiredFeatures);
 
 /**
- * Point annotation class
+ * Point annotation classa.
```
aashish24: classa? --> class?

src/annotationLayer.js
```diff
@@ -6,25 +6,41 @@ var registry = require('./registry');
...
- * @class geo.annotationLayer
+ * @class
+ * @alias geo.annotationLayer
  * @extends geo.featureLayer
- * @param {object?} options
- * @param {number} [options.dblClickTime=300]  The delay in milliseconds that
- *    is treated as a double-click when working with annotations.
- * @param {number} [options.adjacentPointProximity=5]  The minimum distance in
+ * @param {object} [args] Layer options.
+ * @param {number} [args.dblClickTime=300] The delay in milliseconds that is
```
aashish24: I am wondering if 300ms is longer than usual but this is not something for this PR

examples/annotations/main.css
```diff
@@ -4,7 +4,7 @@
   position: absolute;
   left: 10px;
   top: 80px;
-  z-index: 1;
+  z-index: 1000;
```
manthey: When layers are children of other layers, we set a z-index on them.  When we have multiple layers, this conflicts with the z-index context of the controls.  The easy solution is to move the controls to an arbitrary high z-index.  It might be better to ensure that the map is a stacking context, which on some browsers would be adding css `isolation: isolate`, but that doesn't work everywhere and other methods of creating a stacking context involve altering css properties on the map div which might be undesirable in general (we could do so in the examples, though).

src/annotation.js
manthey: Dumb typo.

src/annotationLayer.js
manthey: iOS uses 300 ms, Windows uses 500 ms.

src/canvas/textFeature.js
```diff
@@ -0,0 +1,214 @@
...
+        mapRotation = map.rotation(),
+        mapZoom = map.zoom(),
+        fontFromSubValues, text, pos, visible, color, blur, stroke, width,
+        rotation, rotateWithMap, scale, offset,
+        transform, lastTransform = util.mat3AsArray();
+
+    /* If any of the font styles other than `font` have values, then we need to
+     * construct a single font value from the subvalues.  Otherwise, we can
+     * skip it. */
+    fontFromSubValues = ['fontStyle', 'fontVariant', 'fontWeight', 'fontStretch', 'fontSize', 'lineHeight', 'fontFamily'].some(function (key) {
```
manthey: Done.

src/annotationLayer.js
aashish24: good to know, thanks

PR #720, opened 2017-07-21
Example building for new website

matthewma7:
This is related to 

This PR modifies the existing templates to be Jekyll compatible and removes some unnecessary build steps for the new website. 
This PR also put in small changes such as the location of **data** folder for examples and update thumbnail images for one example.

This change is potentially breaking some tests, but the function is ready to be used.


aashish24:
@matthewma7 as discussed, if you can add the css circle cropping and improve the example demo page, I can add the content and push to your repo.

jeffbaumes:
I like how this is looking so far.

PR #723, opened 2017-07-26
Add isObject utility function.  Use Array.isArray.

manthey:
Fix jsdocs on all touched files.  Add tests.

Prior to this, we were using `instanceof Object` and `instanceof Array`.  For testing the upcoming tutorials branch, it is handy to be able to pass an object created in a parent window to a map running in an iframe.  However, `<parent object> instanceof Object` returned false in a map in the iframe, which makes it hard to interact with the map.  Similarly, the MDN docs recommend using `Array.isArray` over `instanceof` for this same reason.  Using `instanceof` to detect our own classes and `Float32Array` and similar structures still works without causing issues.  It is possible we will want to do something similar for detecting HTML Image and Canvas elements, but for now these have been left alone.

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -254,7 +254,9 @@ transform.transformCoordinates = function (
     }
     return {x: output.x, y: output.y};
   }
-  if (Array.isArray(coordinates) && coordinates.length === 1 && util.isObject(coordinates[0]) && 'x' in coordinates[0] && 'y' in coordinates[0]) {
+  if (Array.isArray(coordinates) && coordinates.length === 1 &&
+      util.isObject(coordinates[0]) && 'x' in coordinates[0] &&
+      'y' in coordinates[0]) {
     output = trans.forward({x: coordinates[0].x, y: coordinates[0].y, z: coordinates[0].z || 0});
     if ('z' in coordinates[0]) {
       return [output];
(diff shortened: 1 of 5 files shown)


aashish24:
@manthey did we search for all instance of?

manthey:
@aashish24 Yes.

src/transform.js
```diff
@@ -246,14 +247,14 @@ transform.transformCoordinates = function (
...
-  if (coordinates instanceof Object && 'x' in coordinates && 'y' in coordinates) {
+  if (util.isObject(coordinates) && 'x' in coordinates && 'y' in coordinates) {
     output = trans.forward({x: coordinates.x, y: coordinates.y, z: coordinates.z || 0});
     if ('z' in coordinates) {
       return output;
     }
     return {x: output.x, y: output.y};
   }
-  if (coordinates instanceof Array && coordinates.length === 1 && coordinates[0] instanceof Object && 'x' in coordinates[0] && 'y' in coordinates[0]) {
+  if (Array.isArray(coordinates) && coordinates.length === 1 && util.isObject(coordinates[0]) && 'x' in coordinates[0] && 'y' in coordinates[0]) {
```
aashish24: long line? Could we wrap it?

manthey: Done.

PR #725, opened 2017-08-01
Add tutorial infrastructure.

manthey:
This adds a method for creating tutorials.  There is a basic tutorial which is mostly defined in a pug template.  Code blocks are user-editable, and will update a live iframe with the results.  The pug template can define tests for code blocks which are run as part of the test suite.

Fixed some inconsistent use of jQuery $.Deferred where sometimes we added the `new` operator before calling it, which is unnecessary.


aashish24:
@matthewma7 would it be possible to update the io pages to include tutorials as well?

matthewma7:
@aashish24 Yes, I'll try to integrated in the the new site. @manthey just a heads up, I may need some input from you in the process.

PR #728, opened 2017-08-04
Fix import order in tutorials.

manthey:
geo.ext.js has to be imported before geo.js for Hammer to be used.  With this change, the tutorials work on an iPad.


manthey:
@matthewma7 This fixes a problem with touch interaction on the tutorials.

matthewma7:
@manthey Cool, thanks!

PR #721, opened 2017-07-21
[WIP] New gh-pages website

matthewma7:
This is a new gh-pages website.
Live demo can be found https://matthewma7.github.io/geojs/

This is paired with the changes on this branch https://github.com/OpenGeoscience/geojs/tree/website_examples

I started with as gh-pages supported Jekyll project but later found out the existing custom site generating with pug/jade. The previous building is still needed, but it is used for building individual example pages only. So, the ph-pages site building involves two different technology at this moment.
(All old example url should still work)

To build or the develop the website, follow the steps in the [readme](https://github.com/OpenGeoscience/geojs/blob/website/README.md) file.

**There are tasks needs to be completed before merging this PR.** (publish the website)
Below are the items I can think of. More items are welcomed.  Once related information is gathered, I can make the changes or changes can be committed directly to this branch. 
- Fill all text placeholder
  - The highlight section of the welcome page
  - The description in "get started" section
  - Example and tutorial page description
- Review all text content
- Determine the order of examples
  There is a requirement that we want to order the examples not just alphabetically, now the order is controlled by the https://github.com/OpenGeoscience/geojs/blob/website/_data/examples.yml file, it also controls what shows on the examples page.
- Review the download page 
   https://matthewma7.github.io/geojs/download
   I put together an initial download page, but because my unfamiliarity with GeoJS, the content of the page needs to be reviewed. Especially, the part about geo.js and geo.ext.js.
-  The two WMS examples need to be updated. It seems there is a CORS issue with the WMS server.
-  Potentially update the jsdocs template to follow the new website design
-  We want to have some tutorials, the website has a placeholder for them, but the tutorials are pending implementation.

PR #729, opened 2017-08-07
New gh-pages website

matthewma7:
This PR supersedes https://github.com/OpenGeoscience/geojs/pull/720 and https://github.com/OpenGeoscience/geojs/pull/721. The other two PR are outdated because of project structure change, now we build the whole website locally then copy to gh-pages branch like before.

This PR creates new scripts for building examples and tutorials for the website, and build the website itself. 

fireshot capture 2 - geojs - welcome - http___localhost_4000_geojs_


manthey:
It looks like all of the examples duplicate their titles and about text from their example.json file in their index.pug file.  Ideally, we shouldn't have this duplicated, as we are bound to fail to update one or the other eventually.

manthey:
The server referenced by WMS examples is probably offline, and just a proxy is returning the CORS errors.  It was offline for a long while before, but eventually came back.  We could switch to using the NOAA WMS server that I know of, but the examples would be fundamentally different.

matthewma7:
@manthey I totally agree, one thing shouldn't be defined in two places, I'll look into it next time. I think the CORS error for WMS service has been there for a couple of weeks, maybe we should have a plan B if it doesn't come back in time.

matthewma7:
@manthey Cool, the text makes the site looks finished now.

aashish24:
@matthewma7 @manthey is this ready for review?

matthewma7:
@aashish24 Yes, this is ready for review.

manthey:
Can we pull the vendor part of the the agency theme from an npm package like `startbootstrap-agency`?

matthewma7:
@manthey I think it's not possible to use the npm package anymore. A large amount of code has been removed and refactored. The only thing intact is the name.

website/_config.yml
```diff
@@ -0,0 +1,83 @@
+# Hexo Configuration
+## Docs: https://hexo.io/docs/configuration.html
+## Source: https://github.com/hexojs/hexo/
+
+# Site
+title: Hexo
```
manthey: Even if this isn't used anywhere, we should probably update it with better information (or at least remove the placeholder data).

website/scaffolds/draft.md
```diff
@@ -0,0 +1,4 @@
+---
```
manthey: If the `scaffolds` directory isn't used, we should remove it.  If it has utility, we should document it.

manthey:
I'd rather not see things like the font-awesome fonts in the repo if we can get them from another location.

website/_config.yml
```diff
@@ -0,0 +1,83 @@
...
+
+# Writing
+new_post_name: :title.md # File name of new posts
+default_layout: layout
+titlecase: false # Transform title into titlecase
+external_link: true # Open external links in new tab
+filename_case: 0
+render_drafts: false
+post_asset_folder: false
+relative_link: false
```
manthey: Can we set this to `true`?  Doing so will also require changing `website/themes/agency/layout/layout.ejs` (see comment there).  By using relative links, you can use any webserver to test the website, not just use `haxo server`, and, if we deploy to a sample to a proxied container, the website will still work.

website/themes/agency/layout/layout.ejs
```diff
@@ -0,0 +1,133 @@
...
+                <a class="navbar-brand" href="<%- url_for("/") %>"><img class="navbar-logo" src="<%- url_for("images/logo_transparent.png") %>" /></a>
+            </div> 
+            
+            <!-- Collect the nav links, forms, and other content for toggling -->
+            <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
+                <ul class="nav navbar-nav navbar-right">
+                    <li class="hidden">
+                        <a href="#page-top"></a>
+                    </li>
+                    <% var path = page_url(path); %>
```
manthey: If we use relative links, this line needs to be removed, and the following line, instead of testing `path === config.root`, we can test `path === "index.html"`.

matthewma7:
@manthey I addressed improvements you pointed out. but I wasn't able to use the startbootstrap-agency from npm because that version is based on bootstrap 4 but I am using bootstrap 3.

This PR is ready for review.

manthey:
@matthewma7 I pushed a change which automatically generates the example and tutorial lists, plus reduces code duplication in some build files.

manthey:
Assuming you approve the PR I made, I see one remaining issue I'd like to address on this PR: the "source code" view from the examples has poor styling.  Specifically, the `PRE` blocks shouldn't set the background and border, and the `.content` blocks shouldn't set the margin and padding.

These are set by the bootstrap styling, which changes the docco styling.

matthewma7:
@manthey Turns out the docco files are preprocessed accidentally. Docco outputs full HTML page already. I think we can leave it as that for now.

matthewma7:
@manthey Cool, Thanks for the reviewing and improvements for this PR. 😉

aashish24:
great job @matthewma7 and @manthey thanks for getting it in master.

PR #730, opened 2017-08-08
Handle tap touch actions. 

manthey:
This specifically supports styli on iPads.

Update documentation on `mapInteractor`.

### DIFF
--- a/src/mapInteractor.js
+++ b/src/mapInteractor.js
@@ -553,7 +553,7 @@ var mapInteractor = function (args) {
         return;
       }
     }
-    var evtType = /^(.*)(start|end|move)$/.exec(evt.type);
+    var evtType = /^(.*)(start|end|move|tap)$/.exec(evt.type);
     if (!evtType || evtType.length !== 3) {
       endIfBound = true;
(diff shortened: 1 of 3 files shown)


src/mapInteractor.js
```diff
@@ -923,6 +936,12 @@ var mapInteractor = function (args) {
     if (m_paused) {
       return;
     }
+    /* In some scenarios, we get both a tap event and then, somewhat later, a
+     * set of mousedown/mouseup events.  Ignore the spurious down/up set if we
+     * just handled a tap. */
+    if (m_touchHandler && m_touchHandler.lastEventType === 'tap' && (new Date()).valueOf() - m_touchHandler.lastTime < 1000) {
```
aashish24: worth wrapping this long line? May be we could have a test for long lines - if makes sense we could allow devs to have length of 79/80 or 120 or not :)

manthey: I don't like having a hard limit on line length.  It makes it a pain to put things like reference urls in comments.   In this case, I'll wrap the line.

PR #731, opened 2017-08-18
Color legend widget

matthewma7:
Implementation of color legend widget.

Support both discrete and continuous scale.
Scale calculation is based on d3 scale.
Discrete color scale supports linear log, sqrt, pow, quantile, ordinal.
Continuous color supports linear log, sqrt, pow. Continuous color supports two color transition.

This is is WIP PR, will test use this widget in a project first.


aashish24:
this is looking awesome @matthewma

matthewma7:
This is ready for review

examples/color-legend/example.json
```diff
@@ -0,0 +1,10 @@
+{
+  "path": "color-legend",
+  "title": "Color legends",
+  "exampleCss": ["main.css"],
```
manthey: For examples with empty css files, we really should set `exampleCss: []` and shouldn't include a blank css file.

src/ui/colorLegendWidget.js
```diff
@@ -0,0 +1,379 @@
...
+  var m_categories = [];
+
+  var oldInit = this._init;
+  // get the widget container ready
+  this._init = function () {
+    oldInit();
+    var canvas = m_this.canvas();
+    d3.select(canvas)
+      .style({
+        'display': 'none',
```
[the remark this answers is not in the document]
matthewma7: Totally, I was wondering about this, but I didn't see much style file in geojs, so I removed style file and stayed with inline style, but it totally works and make sense to have a ui.style file.

tests/cases/colorLegend.js
```diff
@@ -0,0 +1,187 @@
...
+    var legends = $(container).find('.legend');
+    $(legends[0]).find('svg>rect')[0].dispatchEvent(mousemove);
+    $(legends[0]).find('svg>rect')[0].dispatchEvent(mouseout);
+    expect($(container).find('.color-legend-popup').text()).toBe('100 - 200');
+    $(legends[1]).find('svg>rect')[0].dispatchEvent(mousemove);
+    expect($(container).find('.color-legend-popup').text()).toBe('46');
+    $(legends[1]).find('svg>rect')[0].dispatchEvent(mouseout);
+    done();
+  });
+});
```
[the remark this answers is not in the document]
matthewma7: Wow, how do you notice such small misses, thank you for such careful review.
I think I tested popup mouseout event, did the coverage says otherwise? Do you know how to see the line by line code coverage for the colorLegend file?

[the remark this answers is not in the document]
jbeezley: https://codecov.io/gh/OpenGeoscience/geojs/src/2d_legend/src/ui/colorLegendWidget.js

Also, if you run the tests with coverage on locally, it will dump an html coverage report that you can browse.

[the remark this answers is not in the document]
matthewma7: @jbeezley, Thank you! That's really helpful.

matthewma7:
@manthey, Thank you very much for your detailed previous reviews. They totally make sense.
I have made corresponding changes. 
This is ready for further review.

(I don't know why codecov indicates two line is not covered, I have a expect() that tests the results from that two lines)

src/ui/widget.js
```diff
@@ -2,6 +2,18 @@ var inherit = require('../inherit');
...
 /**
+ * @typedef {object} geo.gui.widget.position
+ * @property {string|number} top The position to the top of the container.
+ * A string css position or a number. If a number is used, it will be treated as px value.
+ * @property {string|number} right The position to the right of the container.
+ * Value is used similarly to the top property.
+ * @property {string|number} bottom The position to the bottom of the container.
+ * Value is used similarly to the top property.
+ * @property {string|number} left The position to the left of the container.
+ * Value is used similarly to the top property.
```
manthey: All of these properties are optional.  Further, we should add:
```
@property {*} [...] Additional css properties that affect position are
  allowed.  See the css specification for details.
```

manthey:
With PR #733, you can remove `examples/color-legend/index.pug` and `examples/color-legend/main.css`, and remove the lines that have `path` and `example.css` from `examples/color-legend/example.json`.

matthewma7:
OMG, I finally fixed the build. @manthey, could you take a quick look?

src/ui/colorLegendWidget.js
```diff
@@ -0,0 +1,421 @@
...
+      .call(function (g) {
+        g.call(axis);
+      });
+  };
+
+  /**
+   * Formatter of number that tries to maximize the precision
+   * while making the output shorter.
+   * @param {number} number to be formatted
+   * @param {precision} precision number of decimal precision will be tried to be kept
```
manthey: I think the data type should just be `number`.  I'd recommend changing this to `@param {number} precision Maximum number of decimal places that are kept.`.

src/ui/colorLegendWidget.js
```diff
@@ -0,0 +1,421 @@
...
+  this._hidePopup = function () {
+    m_this.popup.transition()
+      .duration(200)
+      .style('opacity', 0);
+  };
+
+  return this;
+};
+
+function getPrecision(a) {
```
manthey: This function needs to be commented (while you at it, _hidePopup, above, needs it too).

src/ui/colorLegendWidget.js
```diff
@@ -0,0 +1,421 @@
...
+  if (!isFinite(a)) return 0;
+  var e = 1, p = 0;
+  while (Math.round(a * e) / e !== a) { e *= 10; p++; }
```
manthey: If you pass a value with a magnitude less than `1 / Number.MAX_VALUE`, the process can get stuck in an infinite loop (if `Number.MIN_VALUE` is less than `1 / Number.MAX_VALUE` , which is true in my version of Chrome).  Perhaps we should return if `isFinite(a * e)` is false.

Chrismarsh:
I realize this is closed, however I have a question regarding how to use this. When I add a contour plot as well as one of these legends, the colouring of the contour plot doesn't match the legend's colour. Is this to be expected?

manthey:
Are you using the same colors for the contour plot and the legend?  They can be specified independently;  there is nothing that forces them to be the same, though you should be able to use the same colors for both.

If you have a code snippet, we can advise further.

PR #735, opened 2017-09-14
Use Chrome for tests.

manthey:
This is in addition to Firefox and PhantomJS for the basic tests, and as a replacement for the image-based tests.  The tutorials tests pass on all browsers locally (Chrome, Firefox, and PhantomJS), but the Firefox variant occasionally times out on travis, so it has been disabled.

Unlike Firefox, Chrome doesn't report where on the screen it is located.  To work around this, we run Chrome in full-screen kiosk mode and explicitly position the window.

The travis build process now uploads failing image tests to data.kitware.com using a private api_key.  This makes debugging travis failures a little easier.  This is based on a script that was used by Candela for a while.

Thresholds for any performance metrics have been made very lax -- xvfb Chrome is much slower at some things than xvfb Firefox.  One image comparison is different between Ubuntu 16.04 with Chrome stable and travis's Ubuntu 14.04 with Chrome stable.  Visually it looks like the travis version is darker (sort of like its opacity is greater).  This test has been made laxer to pass in both places.  This is still better than Firefox, which renders very different fonts and buttons between travis and Ubuntu 16.04.

Many of the cmake tests have been renamed.  The npm executable is consistently referenced.

Strictly, one of the gl-headless tests isn't headless (the two-map test).  Perhaps the example-tests really should be renamed to headed-tests and the two-map test could be moved to it.  This might make the tests faster, but otherwise probably doesn't matter.

The contour example has added query parameters to make it easier to test manually.  This could be moved to a different PR.

The blog-lines example has been refactor to give a more consistent finished-loading indication.

Some code has been changed to reduce node warnings (specifically, how buffers are allocated), and also to reduce linting warnings.

More context attributes can be set in the webgl context to make the tests between travis and Ubuntu 16.04 more consistent.


aashish24:
> Strictly, one of the gl-headless tests isn't headless (the two-map test). Perhaps the example-tests really should be renamed to headed-tests
+1

karma-base.js
```diff
@@ -173,7 +173,7 @@ var notes_middleware = function (config) {
         });
       } else if (request.method === 'GET') {
         var src = path.resolve(image_path, query.name + '.png');
-        var img = new Buffer(fs.readFileSync(src)).toString('base64');
+        var img = Buffer.from(fs.readFileSync(src)).toString('base64');
```
aashish24: I am curious - can you provide some more information on new vs Buffer.from

package.json
```diff
@@ -89,14 +90,17 @@
     "build-tutorials": "webpack --config webpack-tutorials.config.js",
     "lint": "eslint --cache .",
     "puglint": "pug-lint src examples",
-    "test": "GEOJS_TEST_CASE=tests/test-unit.js karma start karma-cov.conf.js --single-run --browsers PhantomJS",
+    "test": "GEOJS_TEST_CASE=tests/test-unit.js karma start karma-cov.conf.js --single-run --browsers ChromeHeadlessTouch,FirefoxTouch,PhantomJS",
```
aashish24: :+1:

karma-base.js
manthey: https://github.com/nodejs/node/wiki/Breaking-changes-between-v5-and-v6

src/map.js
```diff
@@ -1652,9 +1655,6 @@ var map = function (arg) {
     // for each layer, copy to our new canvas.
     layers.forEach(function (layer) {
       var opacity = layer.opacity();
-      if (opacity <= 0) {
-        return;
```
aashish24: +1

karma-base.js
aashish24: thanks

PR #738, opened 2017-09-20
Improve jsdoc documentation.

manthey:
This fixes the jsdoc linting errors for several files and improves the auto-generated documentation.

### DIFF
--- a/src/camera.js
+++ b/src/camera.js
@@ -310,8 +310,10 @@
     Object.defineProperty(this, 'viewport', {
       get: function () {
         return {
-          width: this._viewport.width, height: this._viewport.height,
-          left: this._viewport.left, top: this._viewport.top
+          width: this._viewport.width,
+          height: this._viewport.height,
+          left: this._viewport.left,
+          top: this._viewport.top
         };
       },
       set: function (viewport) {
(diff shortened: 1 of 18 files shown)


src/camera.js
```diff
@@ -649,6 +655,14 @@
 
     /**
      * Public exposure of the viewFromCenterSizeRotation function.
+     * @param {object} center
+     * @param {number} center.x
+     * @param {number} center.y
```
aashish24: should we make a note of units for size (pixels) and center?

aashish24: Also, the private function has the explanation of what the function does but not the public function. The API docs generated shows only one line. thoughts?

src/featureLayer.js
```diff
@@ -200,15 +214,18 @@ var featureLayer = function (arg) {
...
    */
   this._exit = function () {
     m_this.clear();
     s_exit();
   };
 
   /**
-   * Draw
+   * Draw.  If the layer is visible, call the parent class's draw function and
+   * the renderer's _render function.
```
aashish24: I am thinking that mentioning a private method might be less useful. We could just state the this calls the renderer render function.

src/layer.js
```diff
@@ -428,7 +454,7 @@ var layer = function (arg) {
   };
 
   /**
-   * Clean up resouces
+   * Clean up resouces.
```
aashish24: resources?

src/layer.js
```diff
@@ -444,29 +470,37 @@ var layer = function (arg) {
...
+   *
+   * @returns {number} The height of the parent map in pixels.
    */
   this.height = function () {
     return m_this.map().size().height;
   };
 
   /**
-   * Get or set the current layer opacity.
+   * Get or set the current layer opacity.  The opacity is in teh range [0-1].
```
aashish24: teh = the?

PR #742, opened 2017-10-12
Make a tutorial that acts as an editor

manthey:
Make a tutorial that acts as an editor for easy exploration.

This is my equivalent of using a fiddle site.  The code is stored as an ugly blob in the query string, so it is completely cookie and state free.

In general, any tutorial will store code in its url if 'keep=true' is added to the url when it is loaded.  The editor tutorial always does this.

Allow resizing the tutorial pane.


aashish24:
@manthey hoping to get to review it today.

aashish24:
I think having this is really useful for sharing purposes...

Issue #690, opened 2017-04-27
Demo WMS server used in examples is down


jbeezley:
We are using an external WMS server for the following two examples:
1. http://opengeoscience.github.io/geojs/examples/sld/
2. http://opengeoscience.github.io/geojs/examples/wms/

According to http://demo.geohealthcheck.org/resource/17, it has been down for at least a month.  Should we remove the examples?  Is there a different server we can use?

aashish24:
perhaps we can use the wms server from massgis: We can find few examples here: https://wiki.state.ma.us/display/massgis/GeoServer+-+SLD+-+Polygon+Example+as+Points

jbeezley:
It looks like it's back up.

manthey:
The server we've been using has gone up and down.  It now reports CORS errors.

We need to find a server that is public and supports something interesting for sld requests.

manthey:
This was corrected by switching to https.

PR #743, opened 2017-10-13
Update examples.

manthey:
Tweak a few examples to fix some inconsistencies and errors.  The goal is for every example to start in an interesting spot, and, moreover, for the thumbnail of the example to be representative of the example.

Added an optional `thumbquery` parameter to example.json files.  If present, this should be added to the base url when generating the thumbnail for the example.

Adjusted the zoom and center of several examples to start in more interesting locations.

Fixed some style issues.

If a heatmap was panned before it was ready, it could throw an error.

If a reprojection's viewport corners were ill-defined in the projection, the map would fail to draw.

For the WMS and SLD examples, switch the URL used for the geoserver to https.  This works -- the http server no longer does (this resolves issue #690).

For the SLD example, add controls to allow setting the base url, layer name, and projection.  Clean up some of the pug code.  Update the thumbnail to match what the default settings show.


examples/sld/main.js
```diff
@@ -12,40 +12,36 @@ var layer = {
   projection: 'EPSG:3785'
 };
 
+var baseUrl = 'https://demo.boundlessgeo.com/geoserver/ows';
```
aashish24: +1

src/tileLayer.js
```diff
@@ -1416,16 +1416,23 @@ module.exports = (function () {
...
           mapzoom = map.zoom(),
           roundzoom = this._options.tileRounding(mapzoom),
-          unit = map.unitsPerPixel(zoom === undefined ? roundzoom : zoom);
+          unit = map.unitsPerPixel(zoom === undefined ? roundzoom : zoom),
+          gcsPt;
       if (pt === undefined) {
         var size = map.size();
         pt = {x: size.width / 2, y: size.height / 2};
       }
+      /* displayToGcs can fail under certain projections.  If this happens,
```
aashish24: how can we replicate this behavior in our current version?

aashish24: just curious but agree on the statement

manthey: In the reprojection example, choose a projection that curves at the edges, (ESRI:53009 or any Mo
llweide will work).  Zoom out.  If you zoomed out far enough, this code path is triggered.

aashish24: Got it, thanks.

PR #744, opened 2017-10-31
Fix polygon fill and opacity interaction.

manthey:
When using a constant opacity but a non-constant fill, the fill would appear to not be honored.


aashish24:
@matthewma7 this should fix your use-case I believe. Perhaps you want to go back to changing fill and not opacity since that would optimize the code I think.

PR #746, opened 2017-11-15
Update renderer documentation.

manthey:
This is part of the gradual process of improving our documentation.

### DIFF
--- a/src/d3/d3Renderer.js
+++ b/src/d3/d3Renderer.js
@@ -3,10 +3,19 @@ var registerRenderer = require('../registry').registerRenderer;
 var renderer = require('../renderer');
 
 /**
- * Create a new instance of class d3Renderer
+ * Create a new instance of class d3Renderer.
  *
  * @class geo.d3.renderer
  * @extends geo.renderer
+ * @param {object} arg Options for the renderer.
+ * @param {geo.layer} [arg.layer] Layer associated with the renderer.
+ * @param {HTMLElement} [arg.canvas] Canvas element associated with the
+ *   renderer.
+ * @param {boolean} [arg.widget=false] Set to `true` if this is a stand-alone
+ *   widget.  If it is not a widget, svg elements are wrapped in a parent
+ *   group.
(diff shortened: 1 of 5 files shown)


src/d3/d3Renderer.js
```diff
@@ -59,6 +70,10 @@ var d3Renderer = function (arg) {
   /**
    * Meta functions for converting from geojs styles to d3.
    * @private
+   * @param {function|object} f The style value or function to convert.
+   * @param {function} [g] An optional function that returns a boolean; if it
```
aashish24: I am trying to remember..we are not using square braces for f because it could be an object?

manthey: In jsdoc brackets around a parameter indicate that it is optional.  Here, `f` is required, but `g` is optional.

src/d3/d3Renderer.js
```diff
@@ -502,9 +569,17 @@ var d3Renderer = function (arg) {
     }
```
aashish24: should we return m_this here as well?

src/renderer.js
```diff
@@ -23,28 +27,37 @@ var renderer = function (arg) {
...
+   * @param {HTMLElement} [val] If `undefined`, return the current canvas
+   *    element, otherwise set the canvas element and mark the renderer as
+   *    modified.
+   * @returns {HTMLElement|this} The current canvas element or the renderer
+   *    instance.
    */
   this.canvas = function (val) {
     if (val === undefined) {
       return m_canvas;
-    } else {
```
aashish24: +1 :)

src/d3/d3Renderer.js
manthey: This is called by `renderAnimationFrame`, so any return value will be ignored.

src/d3/d3Renderer.js
aashish24: ah.. that's it. thanks for reminding that.

PR #747, opened 2017-11-15
Preserve annotation ids when reloading geojson.

manthey:
Fix some of the documentation.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -50,6 +50,9 @@ var annotation = function (type, args) {
   delete m_options.annotationId;
   if (m_id === undefined || (m_options.layer && m_options.layer.annotationById(m_id))) {
     annotationId += 1;
+    if (m_id !== undefined) {
+      console.warn('Annotation id ' + m_id + ' is in use; using ' + annotationId + ' instead.');
+    }
     m_id = annotationId;
   } else {
     if (m_id > annotationId) {
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -45,11 +45,19 @@ var annotation = function (type, args) {
     return new annotation(type, args);
   }
 
-  annotationId += 1;
   var m_options = $.extend({}, {showLabel: true}, args || {}),
-      m_id = annotationId,
-      m_name = m_options.name || (
-        type.charAt(0).toUpperCase() + type.substr(1) + ' ' + annotationId),
+      m_id = m_options.annotationId;
+  delete m_options.annotationId;
+  if (m_id === undefined || (m_options.layer && m_options.layer.annotationById(m_id))) {
+    annotationId += 1;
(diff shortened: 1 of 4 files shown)


manthey:
Prior to this, no effort was made to keep annotation IDs the same between generating geojson and setting them via geojson.  This now tries to use the same IDs.  As an example, go to [this example](http://opengeoscience.github.io/geojs/examples/annotations/?geojson=%7B%22type%22%3A%22FeatureCollection%22%2C%22features%22%3A%5B%7B%22type%22%3A%22Feature%22%2C%22geometry%22%3A%7B%22type%22%3A%22Point%22%2C%22coordinates%22%3A%5B-118.87264913546323%2C36.69629877019643%5D%7D%2C%22properties%22%3A%7B%22annotationType%22%3A%22point%22%2C%22name%22%3A%22Camp%22%2C%22annotationId%22%3A20%2C%22fill%22%3Atrue%2C%22fillColor%22%3A%22%2300ff00%22%2C%22fillOpacity%22%3A0.25%2C%22radius%22%3A10%2C%22scaled%22%3Afalse%2C%22stroke%22%3Atrue%2C%22strokeColor%22%3A%22%23000000%22%2C%22strokeOpacity%22%3A1%2C%22strokeWidth%22%3A3%7D%7D%5D%7D).  The point annotation's ID will be 1.  On the current PR it will be 20.

src/annotation.js
```diff
@@ -45,11 +45,19 @@ var annotation = function (type, args) {
...
 
-  annotationId += 1;
   var m_options = $.extend({}, {showLabel: true}, args || {}),
-      m_id = annotationId,
-      m_name = m_options.name || (
-        type.charAt(0).toUpperCase() + type.substr(1) + ' ' + annotationId),
+      m_id = m_options.annotationId;
+  delete m_options.annotationId;
+  if (m_id === undefined || (m_options.layer && m_options.layer.annotationById(m_id))) {
+    annotationId += 1;
```
jbeezley: I wonder if we should emit a warning here if the "requested" id is ignored.  Even better, perhaps if we are wanting to preserve id's across contexts, maybe we should just make them uuid's.  I think the value is fairly opaque with the exception of the default title.

Also, the id attribute isn't documented.  Maybe it's better to just keep it as a private interface due to the id conflict issues.  Thoughts?

manthey: The default title was my main motivation of using low integers.  We could use uuids for the ids, change the accessor function from `id` to `_id` to indicate that it is private, and use a different counter for generating mostly distinct names.

jbeezley: I'm okay with leaving the switch to uuid until later.  I think it would be good to issue a warning here to avoid confusion given how Emory is using these id's.

PR #750, opened 2017-12-04
Make maps better at exiting and reloading.

manthey:
Reduce repeated test code.  This reuses or replaces the #map node in the headless tests rather than creating different named nodes for many different tests.  This reduces the boiler-plate code for creating and destroying maps in the tests.

Reusing elements exhibited some problems in map teardown and recreation.  Specifically: (a) When the map was exiting some functions could be invoked with no-longer-valid renderer values.  (b) An animation frame could fire after the vgl renderer had exited.  (c) The scene object reset its children incorrectly.  (d) The sliderWidget turned off all layer events for zoom rather than just its own.  (e) Tests that mock VGL must clean up their maps before restoring VGL, or when the map is exited it may not have the appropriate state information.

Fix some inconsistencies on widgets.  For instance, if a widget is position relative to the screen and then changed to relative to the map, it wouldn't update on pan, but if it started as relative to the map, it would update on pan.

Update jsdocs on widgets.

### DIFF
--- a/src/gl/quadFeature.js
+++ b/src/gl/quadFeature.js
@@ -147,7 +147,7 @@ var gl_quadFeature = function (arg) {
    * Build this feature.
    */
   this._build = function () {
-    var mapper, mat, prog, srctex, unicrop, geom;
+    var mapper, mat, prog, srctex, unicrop, geom, context;
 
     if (!m_this.position()) {
       return;
@@ -174,10 +174,11 @@ var gl_quadFeature = function (arg) {
       unicrop = new vgl.uniform(vgl.GL.FLOAT_VEC2, 'crop');
       unicrop.set([1.0, 1.0]);
       prog.addUniform(unicrop);
+      context = m_this.renderer()._glContext();
       prog.addShader(vgl.getCachedShader(
-          vgl.GL.VERTEX_SHADER, vgl.GL, vertexShaderImageSource));
+          vgl.GL.VERTEX_SHADER, context, vertexShaderImageSource));
       prog.addShader(vgl.getCachedShader(
-          vgl.GL.FRAGMENT_SHADER, vgl.GL, fragmentShaderImageSource));
+          vgl.GL.FRAGMENT_SHADER, context, fragmentShaderImageSource));
       mat.addAttribute(prog);
       mat.addAttribute(new vgl.blend());
       /* This is similar to vgl.planeSource */
@@ -214,9 +215,10 @@ var gl_quadFeature = function (arg) {
       prog.addUniform(new vgl.projectionUniform('projectionMatrix'));
       prog.addUniform(new vgl.floatUniform('opacity', 1.0));
       prog.addUniform(new vgl.uniform(vgl.GL.FLOAT_VEC3, 'vertexColor'));
+      context = m_this.renderer()._glContext();
       prog.addShader(vgl.getCachedShader(
-          vgl.GL.VERTEX_SHADER, vgl.GL, vertexShaderColorSource));
-      prog.addShader(vgl.utils.createFragmentShader(vgl.GL));
+          vgl.GL.VERTEX_SHADER, context, vertexShaderColorSource));
+      prog.addShader(vgl.utils.createFragmentShader(context));
       mat.addAttribute(prog);
       mat.addAttribute(new vgl.blend());
       /* This is similar to vgl.planeSource */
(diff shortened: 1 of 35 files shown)


src/map.js
```diff
@@ -1482,7 +1482,9 @@ var map = function (arg) {
       if (m_discreteZoom) {
         m_this.zoom(Math.round(m_this.zoom()));
       }
-      m_this.interactor().options({discreteZoom: m_discreteZoom});
+      if (m_this.interactor()) {
```
aashish24: can we have a situation where we do not have interactor?

src/sceneObject.js
```diff
@@ -178,7 +178,7 @@ var sceneObject = function (arg) {
    * Free all resources and destroy the object.
    */
   this._exit = function () {
-    m_this.children = [];
+    m_children = [];
```
aashish24: was this a bug?

manthey: Yes.

src/map.js
manthey: Yes, if you want to display a static map.  We have a test without an interactor.

aashish24: thanks!

src/ui/widget.js
```diff
@@ -171,21 +206,30 @@ var widget = function (arg) {
         // if the property is a number, add px to it, otherwise set it to the
         // specified value.  Setting a property to null clears it.  Setting to
         // undefined doesn't alter it.
-        if (/^\s*(\-|\+)?(\d+(\.\d*)?|\d*\.\d+)([eE](\-|\+)?\d+)?\s*$/.test(position[cssAttr])) {
+        if (/^\s*(-|\+)?(\d+(\.\d*)?|\d*\.\d+)([eE](-|\+)?\d+)?\s*$/.test(position[cssAttr])) {
```
aashish24: whats change here? sorry it is hard to debug

tests/cases/map.js
```diff
@@ -849,15 +843,15 @@ describe('geo.core.map', function () {
 
   describe('Internal methods', function () {
     it('resizeSelf', function () {
-      var m = create_map();
-      expect(m.size()).toEqual({width: 500, height: 500});
-      $('#map-create-map').css({width: '400px', height: '400px'});
-      expect(m.size()).toEqual({width: 500, height: 500});
+      var m = createMap();
+      expect(m.size()).toEqual({width: 640, height: 360});
```
aashish24: why change the size here?

src/ui/widget.js
manthey: I removed some needless escaping; that is all.

tests/cases/map.js
manthey: I'm using a generic test element that is 640 x 360 (for no particular reason that I know of).  Before, this create a map that was a different size; since we don't really care what the size is (only that it is what we expect), when calling the generic function, test that it is the size we expect.

manthey:
I think @matthewma7 found a condition we should guard against, so I'll have at least one more commit to this branch.

manthey:
Before this PR, we weren't exiting the vgl renderer the correct way, so instead of doing a targeted cleanup, it did a general one.  When that was fixed, it exposed the fact that the quads were using a generic context rather than a specific context for their shader cache.  This would hang on to the context. 
 Now they use the correct context, and we forceably try to lose the context as soon as possible (which helps avoid exhausting the available GL contexts).

manthey:
@aashish24 Since I made changes since you approved it, can you make sure you are happy with those changes? (or @matthewma7 could also approve).

tests/cases/map.js
aashish24: ah okay. thanks.

src/gl/quadFeature.js
```diff
@@ -174,10 +174,11 @@ var gl_quadFeature = function (arg) {
       unicrop = new vgl.uniform(vgl.GL.FLOAT_VEC2, 'crop');
       unicrop.set([1.0, 1.0]);
       prog.addUniform(unicrop);
+      context = m_this.renderer()._glContext();
       prog.addShader(vgl.getCachedShader(
-          vgl.GL.VERTEX_SHADER, vgl.GL, vertexShaderImageSource));
+          vgl.GL.VERTEX_SHADER, context, vertexShaderImageSource));
       prog.addShader(vgl.getCachedShader(
-          vgl.GL.FRAGMENT_SHADER, vgl.GL, fragmentShaderImageSource));
+          vgl.GL.FRAGMENT_SHADER, context, fragmentShaderImageSource));
```
aashish24: +1 how did we miss this 😄

PR #751, opened 2017-12-07
Don't automatically install website node modules.

manthey:
When geojs is used as a node module in another project, we should install all of the website's node modules.  Doing so bloats the deployment directory of other projects.  This should be a safe change unless we expect other projects to use the generated geojs website as components within those projects.

This skips installing 10,000 files that total 76 Mb.


manthey:
@matthewma7 I don't think harms anything.

PR #752, opened 2017-12-07
Add setup instruction to the webiste README.md file

matthewma7:
Add an instruction to the website environment setup.
Changes a wording of a step.


matthewma7:
@manthey I just realize we could add a simple instruction. And, the word setup feels better to me, but I am open to any suggestions.

PR #755, opened 2017-12-21
Fix running headed chrome tests with proxy.

manthey:
When we run headed tests on Chrome, we weren't invoking the proxy to serve tiles locally and skip serving fonts.  This means that the tests were running against internet data sources, and therefore subject to the whims of connectivity and changes in OSM data.

Install Google fonts via an npm package and include them as local files for the examples and tutorials.  This avoids having to fetch them each time and also allows the headless tests to have them with certainty.

Fixed the webpack public path for examples and tutorials.


karma-base.js
```diff
@@ -250,8 +250,17 @@ module.exports = function (config) {
         ]
       },
       ChromeWithProxy: {
-        base: 'ChromeFull',
+        // inheriting from ChromeFull ignores the flags in this entry, so we
+        // need to inherit from Chrome
+        base: 'Chrome',
         flags: [
+          '--no-sandbox',  // necessary to run tests in a docker
```
aashish24: @jbeezley could you review the karma conf?

PR #757, opened 2018-01-11
Fix keyboard unbindings.

manthey:
There was a bug when unbinding key handlers.  Mousetrap always expects an array of key sequences, and we were passing one at a time.

Also, jQuery's toggleClass treats 0 as true, so ensure that we always use a boolean.


manthey:
@matthewma7 This should fix the issues you were seeing.

PR #758, opened 2018-01-11
Fix a bug in pointSearch when a feature's gcs was not the default.

manthey:
`pointSearch` gets a point in the map's interface gcs.  Features' gcs defaults to the map's interface gcs.  If the feature has a different gcs, some pointSearch functions did the wrong thing; this has been fixed, and the feature resolution of gcs has been slightly simplified.

### DIFF
--- a/src/feature.js
+++ b/src/feature.js
@@ -518,8 +518,8 @@ var feature = function (arg) {
    */
   this.gcs = function (val) {
     if (val === undefined) {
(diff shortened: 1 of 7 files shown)


manthey:
This issue was discovered as I was working on editing annotations; since it is separable, it is presented as a stand-alone PR.  It will be needed soon.

PR #761, opened 2018-01-15
Use the average perimeter as the center of a polygon or line.

manthey:
When a polygon has many points near each other in one area and few along the opposite part of the perimeter, this produces a nicer center.

### DIFF
--- a/src/util/index.js
+++ b/src/util/index.js
@@ -746,6 +746,39 @@ var util = module.exports = {
   },
 
   /**
+   * Return the coordinate associated with the center of the perimeter formed
+   * from a list of points.
+   *
+   * @param {geo.geoPosition[]} coor An array of coordinates.
+   * @returns {geo.geoPosition|undefined} The position for the center, or
+   *    `undefined` if no such position exists.
+   */
+  centerFromPerimeter: function (coor) {
+    var position, p0, p1, w, sumw, i;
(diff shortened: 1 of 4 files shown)


manthey:
@matthewma7 Would this be useful as a stand-alone function in utils?  It could be named something like `utils.centerFromPerimeter(arrayOfPoints)` where the points are objects with `x`, `y` values.

matthewma7:
@manthey Yes, I think this could be useful, at this moment I have a use case for it. 
I think the function signature you suggested looks good.

src/util/index.js
```diff
@@ -745,6 +745,39 @@ var util = module.exports = {
...
 
+  /**
+   * Return the coordinate associated with the center of the perimeter formed
+   * from a list of points.
+   *
+   * @param {geo.geoPosition[]} coor An array of coordinates.
+   * @returns {geo.geoPosition|undefined} The position for the center, or
+   *    `undefined` if no such position exists.
+   */
+  centerFromPerimeter: function (coor) {
```
aashish24: @manthey it would be good to describe the math in words to explain what is done in this snippet.

manthey: @aashish24 I added a bit more to the function comment.

aashish24: thanks, I found it useful.

PR #762, opened 2018-01-15
When doing a pointSearch on polygons, include strokes.

manthey:
Prior to this, only the interior of the polygon was considered.


manthey:
Note that `pointSearch` will also find a polygon when the point is located in its interior, even if the `fill` style is `false`.  If this behavior is undesired, then the caller is expected to prune out the response (or to have used a line feature).

PR #763, opened 2018-01-22
Shift many references in annotations code of this to m_this.

manthey:
This removes the burden on binding events to the class instance.


src/annotation.js
```diff
@@ -665,7 +669,7 @@ var rectangleAnnotation = function (args) {
           selectionRectangle: true
         }];
       default:
-        return [];
+        return s_actions.apply(this, arguments);
```
aashish24: just checking.. should it be this or m_this?

PR #765, opened 2018-01-22
Update jquery.

manthey:
This shouldn't result in any functional change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4315,9 +4315,9 @@
       "dev": true
     },
     "jquery": {
-      "version": "2.2.4",
-      "resolved": "https://registry.npmjs.org/jquery/-/jquery-2.2.4.tgz",
-      "integrity": "sha1-LInWiJterFIqfuoywUUhVZxsvwI=",
+      "version": "3.3.1",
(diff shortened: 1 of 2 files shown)


manthey:
I was mistaken -- jQuery 2 -> 3 refactors the Deferred object.  This functionally changes the order that callback occur.  Before, in jQuery 2, `var d = $.Deferred(); d.then(function () { console.log('success'); }, function () { console.log('fail'); }); d.resolve(); console.log('end');` would result in printing `success` then `end`.  In jQuery 3, this results in printing `end` then `success` (but a callback specified by a deferred's `done` method appears to be invoked immediately).

manthey:
I'm going to close this until I have a working branch.  Anyone else welcome to address it in the meantime.

brianhelba:
I agree with your summary of the most impactful change: `.then` and `.catch` now always happen asynchronously, whereas `.done`, `.fail`, and `.always` can happen synchronously if the Deferred / Promise has already been resolved at the time of attachment.

Feel free to ping me if you need feedback or review, since I had to deal with this when upgrading jQuery in Girder.

Issue #630, opened 2016-10-20
Enable users to modify / edit annotations via mouse 


aashish24:

manthey:
As per a discussion, once an annotation is in edit mode, it will change its appearance and have several handles that indicate where a user can drag and manipulate it.  For our currently defined annotations:
- point:
  - the point itself is the handle and allows repositioning the point point
- rectangle
  - corner handles: change the position of the two adjacent sides (resize the rectangle)
  - edge handles: change the position of one side
  - center handle: reposition the rectangle
  - outside handle: rotate the rectangle
- polygon:
  - vertex handles: change the position of the vertex, altering the two adjacent edges.  If one vertex is dragged on top of another vertex, it will delete the vertex that has been dragged
  - edge handles: when dragged, create a new vertex and divide the side into two edges
  - centroid handle: reposition the polygon
  - outside handle: rotate the polygon
- line:
  - vertex handles: change the position of the vertex, altering the two adjacent edges.  If one vertex is dragged on top of another vertex, it will delete the vertex that has been dragged.  If the line is not closed and the first and last vertices are made coincident, this should close the line (though there might need to be a way to suppress this)
  - edge handles: when dragged, create a new vertex and divide the side into two edges
  - centroid handle: reposition the line
  - outside handle: rotate the line
  - ? the may need to be some way to break a closed line.

Initially the handles will just be shown as point features.  In the future we may want them to be fancier.  If a manipulation mode is not allowed, the associated handle shouldn't appear (for instance, rotation may not be allowed).

Selecting annotations for editing and turning off edit mode could be the first steps in implementing edit.

Eventually we would like to be able to focus a specific handle and control it via the keyboard, but that can be added after mouse controls.

manthey:
Added in PR #769.

PR #769, opened 2018-01-25
Edit annotations

manthey:
When enabled, hovering over an annotation will highlight it.  Clicking on the annotation will place it in edit mode.  Dragging the edit handles will alter the annotation.  Escape or clicking outside of edit handles will exit from edit mode.

This PR can stand on its own, but there are a number of future improvements that are desired:
- escape should revert the edit rather than just exit edit mode (and enter should exit edit mode, keeping the edit)
- tab should switch between annotations; when an annotation is selected via the enter key, tab would switch between edit handles.
- edit handles should be adjustable with the arrow keys
- meta keys (shift, ctrl) should affect how moving edit handles is controlled (for instance, shift might constrain a rectangle's aspect ratio, or ensure 90 degree rotations).  I'm not sure of the specifics.
- the edit handles should be fancier than just points.  I'd rather see squares on the vertices, a curved double-headed arrow on the rotation handle, and a double-headed arrow on the resize handle.  This would benefit from a more generic marker option for points.
- we should test with a touch device.
- for lines, the ability to break or join a line could be disabled.
- There could be other constraints, too, like discrete rotations or scales.


manthey:
Note that this depends on a number of other PRs and has them as commits (#761, #763, #764, #767, #768).  It could be rebased as those get merged in.

manthey:
Resolves #630.

aashish24:
@manthey this will take some time to review but overall the feature is great!

manthey:
Now that all relevant previous PRs are merged to master, I've rebased this to resolve merge conflicts.

src/annotationLayer.js
```diff
@@ -127,7 +131,14 @@ var annotationLayer = function (args) {
     if (evt.state && evt.state.actionRecord &&
         evt.state.actionRecord.owner === geo_annotation.actionOwner &&
         m_this.currentAnnotation) {
-      update = m_this.currentAnnotation.processAction(evt);
+      switch (m_this.mode()) {
+        case m_this.modes.edit:
```
aashish24: nice and clean.. :+1:

src/annotationLayer.js
```diff
@@ -172,16 +183,118 @@ var annotationLayer = function (args) {
     }
   };
 
+  /**
+   * Select or deselect an edit handle.
+   *
+   * @param {geo.event} evt The mouse move event.
```
aashish24: is this only applicable/callable when "The mouse move"?

src/annotationLayer.js
```diff
@@ -172,16 +183,118 @@ var annotationLayer = function (args) {
...
+   * @returns {this}
+   */
+  this._selectEditHandle = function (evt, enable) {
+    if (!evt.data || !evt.data.editHandle) {
+      return;
+    }
+    $.each(m_features[geo_annotation._editHandleFeatureLevel], function (type, feature) {
+      feature.feature.modified();
+    });
+    m_this.currentAnnotation.selectEditHandle(evt.data, enable);
```
aashish24: curious...here we are passing enable as it is and 2 lines below we are converting into true boolean representation. I think I may know why but just double checking.

src/annotation.js
```diff
@@ -15,11 +15,45 @@ var annotationId = 0;
...
+  },
+  fillOpacity: function (d) {
+    return d.selected ? 0.5 : 0.25;
+  },
+  radius: function (d) {
+    return d.type === 'edge' || d.type === 'rotate' ? 8 : 10;
+  },
+  scaled: false,
+  stroke: true,
+  strokeColor: {r: 0, g: 0, b: 1},
```
aashish24: I would probably make the default color to orange. What you think @matthewma7 ?

src/annotation.js
```diff
@@ -141,6 +175,36 @@ var annotation = function (type, args) {
...
+      return;
+    }
+    offset = offset || 0;
+    rotation = rotation || 0;
+    for (i = 0; i < coor.length; i += 1) {
+      pos = map.gcsToDisplay(coor[i], null);
+      maxr2 = Math.max(maxr2, Math.pow(pos.x - dispCenter.x, 2) + Math.pow(pos.y - dispCenter.y, 2));
+    }
+    r = Math.sqrt(maxr2) + offset;
+    pos = map.displayToGcs({x: dispCenter.x + r * Math.cos(rotation), y: dispCenter.y - r * Math.sin(rotation)}, null);
```
aashish24: long line?

src/annotation.js
```diff
@@ -575,6 +748,260 @@ var annotation = function (type, args) {
...
+        },
+        map = m_this.layer().map(),
+        p0 = map.gcsToDisplay(handle.center, null),
+        p1 = map.gcsToDisplay(handle.resizePosition, null),
+        p2 = map.gcsToDisplay({
+          x: handle.resizePosition.x + delta.x,
+          y: handle.resizePosition.y + delta.y
+        }, null),
+        d01 = Math.pow(Math.pow(p1.y - p0.y, 2) + Math.pow(p1.x - p0.x, 2), 0.5) - handle.handle.style.resizeHandleOffset,
+        d02 = Math.pow(Math.pow(p2.y - p0.y, 2) + Math.pow(p2.x - p0.x, 2), 0.5) - handle.handle.style.resizeHandleOffset,
```
aashish24: wrap these long lines?

src/annotation.js
```diff
@@ -575,6 +748,260 @@ var annotation = function (type, args) {
...
+      x: start[index].x + delta.x,
+      y: start[index].y + delta.y
+    };
+    if (layer.displayDistance(curPts[index], null, start[index], null) <= aPP) {
+      /* if we haven't moved at least aPP from where the vertex started, don't
+       * allow it to be merged into another vertex.  This prevents small scale
+       * edits from collapsing immediately. */
+    } else if (layer.displayDistance(curPts[index], null, curPts[(index + 1) % curPts.length], null) <= aPP) {
+      near = (index + 1) % curPts.length;
+    } else if (layer.displayDistance(curPts[index], null, curPts[(index + curPts.length - 1) % curPts.length], null) <= aPP) {
```
aashish24: same here..wrap?

src/annotationLayer.js
manthey: This gets called with the mouseon and mouseoff events.  Those only occur on a mouse move or when the mouse move is retriggered.  When we want to make sure touch devices are fully supported, we should trigger mouseon and mouseoff events (and mouseover and mouseout) before the touch causes a click (but that is an issue for another PR).

src/annotation.js
```diff
@@ -876,6 +1306,84 @@ var rectangleAnnotation = function (args) {
     }
   };
 
+  /**
+   * Process any edit actions for this annotation.
+   *
+   * @param {geo.event} evt The action event.
+   * @returns {boolean|string} `true` to update the annotation, falsy to not
```
aashish24: 'true' --> Truthy to update the annotation? To keep it consistent with other docs?

src/action.js
```diff
@@ -15,7 +15,8 @@ var geo_action = {
   // annotation actions
   annotation_line: 'geo_annotation_line',
   annotation_polygon: 'geo_annotation_polygon',
-  annotation_rectangle: 'geo_annotation_rectangle'
+  annotation_rectangle: 'geo_annotation_rectangle',
+  annotation_edit_handle: 'geo_annotation_edit_handle'
```
aashish24: should we just call it annotation_edit? or would that be confusing?

src/annotationLayer.js
manthey: Most of javascript handles truthy or falsy values they way you'd expect.  However, jQuery's toggleClass does *not*, and must take either an actual boolean or a function that returns a boolean.

aashish24: ah, thanks good to know.

src/annotation.js
matthewma7: @aashish24 The current color works fine for me, because I used some yellow-orange color in the other project.

src/annotation.js
manthey: On return values, we can be specific -- it *does* output `true`, not just an arbitrary truthy value.  On input parameters we try to be flexible and accept any truthy value.

src/action.js
manthey: We may eventually want other edit actions.  For instance, suppose we supported an action on selecting the annotation in edit mode, then it might be `edit_select` to differentiate it.

aashish24:
@manthey @matthewma7 this is looking very good. Some points for discussions / future work may be (I am interesting in hearing your thoughts on it):

1. I feel that sometimes (depending on how I draw my area, the handlers were little too far

2. When I rotate, the scale handle remain at its place, is that intentional? Should it also rotate?

3. For point annotations, should we think about scaling as well (only scale handle)?

manthey:
@aashish24, in response to your numbered list:

1. I'm not sure what you were saying here.

2. Yes, the scale handle doesn't rotate.  It makes it more obvious that it isn't the rotation handle.  

3. We could scale points, but that seems like a mildly odd thing to do conceptually (it would be easy technically).  If we want to do that, we can make add it in another PR (and, if desired, the default could be to *not* show it).

aashish24:
@manthey for 

1) what I meant that the handles distance from the center of the shape seemed too far to me in some cases. Perhaps I can show it live. 

2) Sounds good. 

3) Sounds good as well. Not a major request,  I felt that it would be nice to be able to scale and move but this is low priority for me.

PR #772, opened 2018-02-05
Update to jQuery 3.x.

manthey:
This shouldn't result in any functional change.

The main difference between jQuery 2.x and jQuery 3.x, at least as impacts GeoJS, is that `Deferred` objects are less synchronous.  Specifically, the `done`, `fail`, and `always` methods can be
synchronous if the deferred is resolved or rejected, but the `then` method is never synchronous.  The major changes are in the tests which relied on the synchronicity for running the tests; they no longer do so, releasing time slices as appropriate.

Since GeoJS `object`s could take native promises, there is code to use `always` if available and `then` if not.  In other places, we rely on jQuery's `Deferred` having the `done` and `always` convenience functions.

### DIFF
--- a/tests/cases/transform.js
+++ b/tests/cases/transform.js
@@ -119,7 +119,7 @@ describe('geo.transform', function () {
       expect(p.forward({x: 10, y: -10, z: 0})).toEqual({x: 10, y: -10, z: 0});
     });
 
-    it('lookup', function () {
+    it('lookup', function (done) {
       var spy = sinon.spy(), request;
       geo.transform.lookup('EPSG:5000').then(spy);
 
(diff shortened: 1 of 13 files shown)


tests/cases/transform.js
```diff
@@ -165,8 +168,11 @@ describe('geo.transform', function () {
         results: []
       }));
 
-      expect(spy.calledOnce).toBe(true);
-      expect(geo.transform.defs.hasOwnProperty('EPSG:5001')).toBe(false);
+      window.setTimeout(function () { // wait for the next time slice
```
aashish24: why we need window timeout now?

manthey: jQuery 3 is always asynchronous for certain requests where jQuery 2 was sometimes synchronous. Specifically, we are making an ajax request to get the transform from a web service.  We mock the web service for the test, so the mock returns immediately.  But, in jQuery 3, the callback is fired in the next time slice rather than synchronously, so we wait 0 ms (just to be in the next time slice) to make sure the callback has been triggered. 

In jQuery 2, the behavior was ajax->mock response->callback->next time slice.

In jQuery 3, the behavior is ajax->mock response->next time slice->callback.

If we were doing an actual request which couldn't be served from browser cache, we would have to wait an unknown amount of time before the callback would be triggered.  It is only because we are mocking the response that the immediate check worked in jQuery 2 (and the check with a 0-duration wait works in jQuery 3).  Rather than use window.setTimeout, we could have used time mocking, but that doesn't make the test any clearer in my opinion.

aashish24: thanks for the detailed explanation @manthey.

PR #775, opened 2018-02-09
Harden some tests; fix a typo.

manthey:
Convert some of the throttle tests to use sinon timers for more reliable testing.

Fix some tutorial tests since Chrome 64 changed srcdoc in iframe usage.

Harden the test to check if a video is ready to play.

### DIFF
--- a/src/gl/polygonFeature.js
+++ b/src/gl/polygonFeature.js
@@ -155,7 +155,7 @@ var gl_polygonFeature = function (arg) {
           });
         }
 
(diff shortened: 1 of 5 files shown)


manthey:
I'll note that in travis, we can't just pin to Chrome 63 for testing.  You can only use `stable` or `beta` (even though `stable` isn't actually stable).

manthey:
I'll go ahead and merge.  If we decide we should have done something different, we can make another PR.

PR #779, opened 2018-02-22
Improve line updates

manthey:
Speed up line style updates.

Before, any change to a line style would recompute the geometry information as well.  Now, if the position of vertices has not changes, the geometry is not recomputed.  Further, many style were being called by function when they were their default values (such as the strokeOffset not being set).  These properly are treated as fixed values, avoiding many function calls.  The polygon feature, when stroked, is better at passing constant values for stroke opacity to the lines.

In practice, this results in a large speed up.  For a condition where there are lines with ~100,000 vertices, the line _build call was reduced from 450-500 ms to 12-15 ms.

### DIFF
--- a/src/gl/lineFeature.js
+++ b/src/gl/lineFeature.js
@@ -34,10 +34,11 @@ var flagsDebug = {  // uses 1 bit
 };
 
 /**
- * Create a new instance of lineFeature
+ * Create a new instance of lineFeature.
  *
  * @class geo.gl.lineFeature
  * @extends geo.lineFeature
+ * @param {geo.lineFeature.spec} arg
  * @returns {geo.gl.lineFeature}
  */
 var gl_lineFeature = function (arg) {
@@ -72,6 +73,11 @@ var gl_lineFeature = function (arg) {
       s_init = this._init,
       s_update = this._update;
 
+  /**
+   * Create the vertex shader for lines.
+   *
+   * @returns {vgl.shader}
+   */
   function createVertexShader() {
     var vertexShaderSource = [
           '#ifdef GL_ES',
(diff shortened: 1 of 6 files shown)


manthey:
@kheffah: This resolves a performance issue reported in https://github.com/DigitalSlideArchive/HistomicsTK/issues/459.

src/polygonFeature.js
```diff
@@ -328,15 +326,18 @@ var polygonFeature = function (arg) {
       strokeStyle: linePolyStyle(polyStyle.strokeStyle),
       strokeColor: linePolyStyle(polyStyle.strokeColor),
       strokeOffset: linePolyStyle(polyStyle.strokeOffset),
-      strokeOpacity: function (d) {
-        return m_this.style.get('stroke')(d[2], d[3]) ? m_this.style.get('strokeOpacity')(d[0], d[1], d[2], d[3]) : 0;
-      }
+      strokeOpacity: util.isFunction(polyStyle.stroke) || !polyStyle.stroke ?
+        function (d) {
```
aashish24: +1

src/gl/lineFeature.js
```diff
@@ -655,23 +723,32 @@ var gl_lineFeature = function (arg) {
...
+   * strokeOffset), the vertex flags are not recomputed -- this helps, as it is
+   * a slow step due to most javascript interpreters not optimizing bit
+   * operations.
    *
-   * @override
+   * @returns {this}
    */
   this._build = function () {
-    createGLLines();
+    createGLLines(m_this.dataTime().getMTime() < m_this.buildTime().getMTime() && m_geometry);
```
aashish24: +1

PR #782, opened 2018-02-27
When available, use the mouse event buttons property.

### LINKED ISSUE #781: On Firefox, mosemove events always indicate the left mouse button is clicked
For example, binding to `geo.event.feature.mouseclick` will always fire events where the property `evt.mouse.buttonsDown.left` is `true`.

The cause appears to be that GeoJS [uses `MouseEvent.which`](https://github.com/OpenGeoscience/geojs/blob/2854ccb087d30aa87fed2984929a29e0a0027940/src/mapInteractor.js#L854) to determine which mouse buttons are pressed. According to [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/which) and [a bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1048294), `MouseEvent.which` does not work properly on Firefox.

[`MouseEvent.buttons`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons) should probably be used instead, though it appears to not be supported by Safari.

manthey:
Firefox erroneously sets the which value to the left button even when it is not held down.  The `buttons` property is correct for browsers that support it, but Safari doesn't, so we have to fall back to using the `which` value in that case.

This fixes #781.

### DIFF
--- a/src/mapInteractor.js
+++ b/src/mapInteractor.js
@@ -850,17 +850,26 @@ var mapInteractor = function (args) {
         m_mouse.buttons[prop] = false;
       }
     }
-    if (evt.type !== 'mouseup') {
+    /* If the event buttons are specified, use them in preference to the
+     * evt.which for determining which buttons are down.  buttons is a bitfield
+     * and therefore can represent more than one button at a time. */
+    if (evt.buttons !== undefined) {
+      m_mouse.buttons.left = !!(evt.buttons & 1);
+      m_mouse.buttons.right = !!(evt.buttons & 2);
+      m_mouse.buttons.middle = !!(evt.buttons & 4);
+    } else if (evt.type !== 'mouseup') {
+      /* If we don't evt.buttons, fall back to which, but not on mouseup. */
       switch (evt.which) {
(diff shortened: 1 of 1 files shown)


src/mapInteractor.js
```diff
@@ -850,18 +850,27 @@ var mapInteractor = function (args) {
...
-          if (evt.which) {
-            m_mouse.buttons[evt.which] = true;
-          }
-          break;
       }
     }
+    /* When handling touch events, evt.which can be a string, in which case
+     * handle a "button" with that name -- a non-integer string will not
+     * evaluate as being between 1 and 3. */
+    if (evt.which && !(evt.which >= 1 && evt.which <= 3)) {
```
brianhelba: Should this be a separate conditional block, whereby one of the other previous conditionals runs with this too?

manthey: It doesn't need to be, as in the touch events where it will get triggered, it can never also be in a `mouseup` event, and we want it to trigger regardless of the buttons conditional.

PR #784, opened 2018-02-28
Update to version 0.15.0

jbeezley:
I arbitrarily choose a minor version bump because of the editable annotations, but it could be a patch release as there are no breaking changes that I'm aware of.


manthey:
I agree with the minor version bump.

matthewma7:
@jbeezley Cool.

PR #791, opened 2018-03-06
Distance and Ramer–Douglas–Peucker functions.

manthey:
Move distance functions from the lineFeature to the util module, making the function names more explicit.  Add a `rdpLineSimply` function for reducing the complexity of a line or polygon.  Add a `rdpSimplifyData` method to the line and polygon features for doing this while setting data.


src/polygonFeature.js
```diff
@@ -392,6 +401,89 @@ var polygonFeature = function (arg) {
...
 
+  /**
+   * Take a set of data, reduce the number of vertices per polygon using the
+   * Ramer–Douglas–Peucker algorithm, and use the result as the new data.
+   * This changes the instance's data, the position accessor, and the polygon
+   * accessor.
+   *
+   * @param {array} data A new data array.
+   * @param {number} [tolerance] The maximum variation allowed in map.gcs
+   *    units.  A value of zero will only remove perfectly colinear points.  If
```
aashish24: I am wondering if the unit should be pixels for the tolerance?

manthey: It defaults to 0.5 pixel at the current zoom.  But if you are overriding this, you might not want the tolerance to be dependent on the current scale of the map.  So this can be made into current zoom pixels via `map.unitsPerPixel(map.zoom()) * <tolerance in pixels>`, or it can be in gcs units.

src/util/index.js
```diff
@@ -781,6 +781,202 @@ var util = module.exports = {
...
   },
 
+  /**
+   * Get the square of the Euclidean 2D distance between two points.
+   *
+   * @param {geo.geoPosition} pt1 The first point.
+   * @param {geo.geoPosition} pt2 The second point.
+   * @returns {number} The distance squared.
+   */
+  distance2dSquared: function (pt1, pt2) {
```
aashish24: should we just have distance2d? It may be worth using turfjs? (https://turfjs.org/)

src/util/index.js
```diff
@@ -781,6 +781,202 @@ var util = module.exports = {
...
+  /**
+   * Get the square of the Euclidean 2D distance between a point and a line
+   * segment.
+   *
+   * @param {geo.geoPosition} pt The point.
+   * @param {geo.geoPosition} line1 One end of the line.
+   * @param {geo.geoPosition} line2 The other end of the line.
+   * @returns {number} The distance squared.
+   */
+  distance2dToLineSquared: function (pt, line1, line2) {
```
aashish24: same here I think distance2dToLine is probably more useful?

tests/cases/lineFeature.js
```diff
@@ -197,6 +197,60 @@ describe('geo.lineFeature', function () {
...
+        line.data(testLines).line(lineFunc);
+        counts = countLines(line.data().map(line.style.get('line')));
+        expect(counts).toEqual({lines: 8, vertices: 23});
+        line.rdpSimplifyData(testLines, undefined, undefined, lineFunc);
+        counts = countLines(line.data().map(line.style.get('line')));
+        expect(counts).toEqual({lines: 8, vertices: 18});
+
+        // use pixel space for ease of picking tolerance values in tests
+        map.gcs('+proj=longlat +axis=enu');
+        map.ingcs('+proj=longlat +axis=esu');
```
aashish24: nice

src/util/index.js
manthey: If we return the non-squared versions, there is a lot of needless sqrt function calls, since mostly we are trying to find the smallest value.

src/polygonFeature.js
aashish24: so the user has to pass t = map.unitsPerPixel(map.zoom()) * <tolerance in pixels> or else we will use map.unitsPerPixel(map.zoom()) * 0.5?

src/util/index.js
manthey: turfjs internally does everything on a spheroid.  Their distance function, for instance, is much heavier because of this, always making multiple trigonometric function calls.  Since we are usually already in a 2d spatial domain, this is vastly slower (and doesn't extend to pixel-space examples in a rational manner, or ellipsoids that are highly eccentric, etc.).

aashish24: sure, perhaps in the future we could consider it.

src/polygonFeature.js
aashish24: I see. I am wondering if we could auto compute it given some initial parameter in a separate PR (as a new feature), not needed now, but wanted to mention it.

PR #793, opened 2018-03-07
Use https for references to openstreetmap.org.

manthey:
Recently, all http requests to openstreetmap.prg get a 307 response to forward them to the https server.  It is more efficient to access the https server directly.  OSM has been running https servers for several years now (at least since 2014), so this should be as stable as OSM is in general.

I haven't figured out how to proxy https requests to karma.  Perhaps we'd have to run karma in https mode, but my first foray into that didn't work.  For now, the one test (the lines example) just uses the http openstreetmap.org reference in the test.


aashish24:
awesome.. this is very quick. I am assuming that you addressed the testing issue?

manthey:
This uses http with a proxy for testing (see PR comment).

Also, Chrome 65 headless with mesa has what looks like a bug that caused travis to fail on existing tests, so this PR is dependent on #792.

PR #798, opened 2018-03-19
Add a scale widget

manthey:
This adds a scale widget and a tutorial to show it off.


manthey:
All but the last commit are from other PRs.  When they are merged, this could be rebased to contain just the relevant commit.

PR #799, opened 2018-03-21
Improve the screenshot by dereferencing CSS urls.

manthey:
If CSS included a URL resource, such as for a font or background image, it was not being dereferenced, which would lead to it not rendering as expected on the screenshot.


aashish24:
this was quick :+1:

PR #802, opened 2018-03-29
Only allow pointer events on children of active layers.

manthey:
This allows multiple active layers to work together.  Before, the topmost active layer will receive all pointer events.

This also fixes an issue in the sceneObject exit function (it should clear the parent, not delete the parent function).  This, in turn, required a fix to the annotation test because the test map is destroyed between tests.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -937,6 +937,8 @@ var map = function (arg) {
       m_this.interactor().destroy();
       m_this.interactor(null);
     }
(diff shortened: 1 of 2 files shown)


aashish24:
interesting that we did not catch this before. I am wondering how did you find about this?

manthey:
@matthewma7 reported this when he created two UI layers.  If we have only one UI layer, the issue wasn't obvious.

PR #803, opened 2018-03-30
Handle dereferencing errors more gracefully.

manthey:
This will log relevant warnings and not fail if a css file is missing.

Check if html to image via svg canvas is supported.

Add some more tests and comments.


aashish24:
is this something @matthewma7 needs to review?

PR #809, opened 2018-04-13
improvement tutorial page layout

matthewma7:
improvement tutorial page layout by not wrapping tutorial title text
And also update hexo version


matthewma7:
@aashish24 
I kind of already pushed this change to the site. 
But I can simulate a before and after:
Before:
image

After:
image

PR #817, opened 2018-05-15
Generalize meshes

manthey:
Generalize the rectangular grid used in the contour feature to a mesh that could also be composed of enumerated triangular or square elements.

This adds more tests to the contour feature and fixes some jsdocs.


manthey:
@aashish24 This now has tests for all of the new code.

src/meshFeature.js
```diff
@@ -0,0 +1,444 @@
...
+   *
+   * @param {object} [vertexValueFuncs] A dictionary where the keys are the
+   *    names of properties to include in the results and the values are
+   *    functions that are evaluated at each vertex with the arguments
+   *    (data[idx], idx).  If a key is named `used`, then if its function
+   *    returns a falsy value for a data point, the vertex associated with that
+   *    data point is removed from the resultant mesh.
+   * @returns {geo.meshFeature.meshInfo} An object with the mesh information.
+   */
+  this.createMesh = function (vertexValueFuncs) {
```
aashish24: @manthey does this needs to be in public API?

manthey: It doesn't need to be.  My reasoning for it to be a public API, is that making so encourages its reuse in other components.

src/contourFeature.js
```diff
@@ -1,93 +1,56 @@
...
- *      data element.  This defaults `z` properties of the data element.  If
- *      the value of a grid point is `null` or `undefined`, that point will not
- *      be included in the contour display.  Since the values are on a grid, if
- *      this point is in the interior of the grid, this can remove up to four
- *      squares.
+ *    feature.
+ * @property {number|function} [style.opacity=1] The opacity on a scale of 0 to
+ *    1.
+ * @property {geo.geoPosition|function} [style.position=data] The position of
+ *    each data element.  This defaults to just using `x`, `y`, and `z`
```
aashish24: very minor document issue (not sure if that matters when this will get rendered). Seems to have lot many blank spaces between the '.' and 'This'.

manthey: There are two popular choices in typography, two spaces between sentences or one.  There are impassioned arguments for each side, and, since it is opinion, there is no strict right or wrong. 
 I use two spaces.  HTML is poor in that it collapses all spaces to a single space, losing the semantic difference between sentence separation and word separation, but in the source, this is still apparent.  In old typography manuals, it used to be argued that word separation should be 1/5 em and sentence separation should be 1/3 em, but that doesn't work in monospaced fonts.

aashish24: I see, I am fine with this as long as we are consistent. We could probably mention this somewhere so that a new developer follows the same.

src/meshFeature.js
aashish24: I see but from the end-user perspective, I do see rare use specially since this does not seem like a self contained method. I would make it _createMesh but if you have strong feelings about keeping it public, I am fine with that.

src/contourFeature.js
manthey: We need to update the docs in general to mentioned all of the desired behavior for jsdocs in general.

src/meshFeature.js
manthey: I'll change it.

PR #818, opened 2018-05-16
Use Firefox headless where appropriate.

manthey:
Significantly reduce the number of external web requests made by Firefox and Chrome during testing.


manthey:
When running headless tests while some web services were slow, the tests were oddly slower.  As the headless tests do not themselves make any external requests, it was surprising.  Both Chrome and Firefox make many external requests when starting, even in headless and incognito modes.  By adding appropriate flags and preferences, many of the requests are no longer made.  When there is a good network connection, this speeds up the overall tests a small amount (1 or 2 percent based on a limited set of runs).  When there is a bad network connection, this has a much bigger impact.

Firefox still makes external connections the first time 2D canvas is used.  Chrome still contacts google.com once.

karma-base.js
```diff
@@ -256,25 +260,98 @@ module.exports = function (config) {
...
+          'browser.safebrowsing.provider.mozilla.updateURL': '',
+          'browser.safebrowsing.provider.google.gethashURL': '',
+          'browser.safebrowsing.provider.google.updateURL': '',
+          'browser.safebrowsing.provider.google4.dataSharingURL': '',
+          'browser.safebrowsing.provider.google4.gethashURL': '',
+          'browser.safebrowsing.provider.google4.updateURL': '',
+          'datareporting.healthreport.uploadEnabled': false,
+          'datareporting.policy.dataSubmissionEnabled': false,
+          'media.gmp-gmpopenh264.autoupdate': false,
+          'media.gmp-manager.url': ''
```
jbeezley: Could you dedup the common parts of these `prefs` objects?

PR #823, opened 2018-05-23
Warn on trying to create an unsupported feature.

manthey:
When a feature is requested that a layer doesn't support, issue a warning.  Don't fail in the geojson reader when this occurs.

### DIFF
--- a/src/featureLayer.js
+++ b/src/featureLayer.js
@@ -45,6 +45,8 @@ var featureLayer = function (arg) {
       featureName, m_this, m_this.renderer(), arg);
     if (newFeature) {
       this.addFeature(newFeature);
+    } else {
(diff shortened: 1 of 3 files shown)


src/featureLayer.js
```diff
@@ -45,6 +45,8 @@ var featureLayer = function (arg) {
       featureName, m_this, m_this.renderer(), arg);
     if (newFeature) {
       this.addFeature(newFeature);
+    } else {
+      console.warn('Layer renderer (' + m_this.rendererName() + ') does not support feature type "' + featureName + '"');
```
aashish24: perhaps the line is little too long (I cannot read end-to-end on my browser) but other than that it looks good.

PR #824, opened 2018-05-23
Improve polygon and line documentation.

manthey:
This should make it clearer what the position, line, and polygon accessors do.  It also explicitly defines a polygon data type.

### DIFF
--- a/src/lineFeature.js
+++ b/src/lineFeature.js
@@ -8,39 +8,39 @@ var util = require('./util');
  * Line feature specification.
  *
  * @typedef {geo.feature.spec} geo.lineFeature.spec
- * @param {object|Function} [position] Position of the data.  Default is
+ * @property {object|function} [position] Position of the data.  Default is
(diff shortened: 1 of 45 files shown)


src/lineFeature.js
```diff
@@ -82,11 +83,13 @@ var lineFeature = function (arg) {
...
-   * Get/set lineaccessor.
+   * Get/set line accessor.
    *
-   * @param {object} [val] if specified, use this for the line accessor
-   *    and return the feature.  If not specified, return the current line.
-   * @returns {object|this} The current line or this feature.
+   * @param {object|function} [val] If specified, use this for the line
+   *    accessor and return the feature.  If not specified, return the current
+   *    line.  If a function, this is passed `(dataElement, dataIndex)` and
+   *    returns an array of vertex elements.
```
aashish24: returns or return? above the sentences read return.

PR #827, opened 2018-05-24
Fix having tutorials update if text is deleted.

manthey:
Tutorials should rerun whenever the contents change.  Some code designed to perform updates when separate code blocks were edited happened to prevent deleting text from triggering the update.

### DIFF
--- a/tutorials/common/tutorials.js
+++ b/tutorials/common/tutorials.js
@@ -192,8 +192,8 @@ function process_block_debounce(selector, debounce) {
     processBlockInfo.lastelem = selector;
     if (!debounce) {
       process_block(selector);
+      return;
     }
-    return;
   }
   processBlockInfo.timer = window.setTimeout(function () {
     process_block(processBlockInfo.lastelem);


tutorials/common/tutorials.js
```diff
@@ -192,8 +192,8 @@ function process_block_debounce(selector, debounce) {
     processBlockInfo.lastelem = selector;
     if (!debounce) {
       process_block(selector);
+      return;
     }
```
aashish24: @manthey could you explain why this change? Looking at the code it is not exactly clear to me.

manthey: When a text area is altered, we want to rerun the tutorial, but we want to debounce it so it doesn't rerun too often (to give you a chance to finish typing).  We trigger this on textarea `change`, `input`, and `propertychange` events (and resetting a field).  The specific bug that this fixes is when no editing has been done, and you click on a text area and just delete a character.  In this case, because the area had not been the most recently altered, it enters this code path.  However, since we are trying to debounce, instead of processing the block, it should fall through and set a timer to run after debouncing.  But, the `return` was in the wrong place, and it never set the timer.  If you clicked on another text area, or typed more, then this branch (from line 188) wasn't entered, and the tutorial would update.  The `return` should only apply if we aren't debouncing and are processing the block immediately.

PR #829, opened 2018-05-25
Switch test progress to use the karma-spec-reporter.

manthey:
Instead of printing test progress like
```
JS 2.1.1 (Linux 0.0.0): Executed 89 of 1022 SUCCESS (0 secs / 0 secs)
```
this prints information like
```
  geo.annotation
    geo.annotation.lineAnnotation
      ✓ _coordinates
```

For coverage tests, I've set it to only print failures.  This will make it easier to read through results on CI.

Also, the colorLegend test was failing when run without coverage but passing when run with coverage.  This has to do with how browsers are invoked, and was the result of a shift the location of the map element.  This location is now used to determine where to perform a mouse test, resulting in identical results regardless of coverage.  By fixing how a mouse event was created in PhantomJS, PhantomJS now gets identical results to Firefox.


aashish24:
@jbeezley could you review it?

PR #834, opened 2018-05-31
Add a parameter to the text feature to limit what gets rendered.

manthey:
If specified, this skips rendering text that is outside of the current viewport by a specified distance.  If you have very large text or text with offsets, this could fail to render the text.  However, in well-formed cases, this can skip rendering text that would not be visible in any case.


src/textFeature.js
```diff
@@ -60,6 +60,12 @@ var feature = require('./feature');
  *      stroke color.  May include opacity.
  * @property {geo.geoColor|function} [style.textStrokeWidth=0] Text stroke
  *      width in pixels.
+ * @property {number|function} [style.renderedZone] If this is a positive
```
aashish24: @manthey i am not entire sure if the name of the variable truly reflects the intention. Would you be open to call it something like renderThreshold or viewportThreshold or something else?

manthey: Of those, I think `renderThreshold` would be clearer, or even a longer combination like `renderViewportThreshold`.  Do you have a preference?

aashish24: renderThreshold seems good enough for me now.  Since we do not use the word viewport otherwise, I would have chosen renderViewportThreshold.

the code LGTM

manthey: Updated to `renderThreshold` and rebased to current master.

PR #837, opened 2018-06-05
Upgrade jasmine and some tutorial testing infrastructure.

manthey:
This upgrades to jasmine 3.1 and also the karma-jasmine-html-reporter.  There are clearer tutorial test failure messages, especially when the test throws an exception.

Fixed some wording in the contour tutorial.

A small performance improvement was made to the gl.lineFeature.


manthey:
You are referring to npm's package-lock.json file, which is essentially opaque to us -- we just generate it so that we get consistent builds.  Different versions of npm frequently change how they store information in the package-lock, and I haven't bothered to learn why.

PR #838, opened 2018-06-05
Add an isoline feature.

### LINKED ISSUE #592: Add Isolines / Contours with labels
Isolines or contour lines are  commonly used for climate and weather datasets.

manthey:
This adds a `Math.log10` polyfill and removes the unused `Math.sinh` polyfill.

Add simple isoline example.

Add an isoline tutorial.  This uses a data set of Oahu's elevation that is lower resolution than the 'dense' set, primarily so that the tutorial will run faster (and test faster).

This resolves #592.

This depends on a variety of open PRs (and includes them as part of this PR) and could be rebased once those are merged it.


aashish24:
@manthey is it possible to update this branch?

manthey:
It still depends on the `update-jasmine` branch.  Now that the `util-wrap-angle` branch is merged, it can be a simple dependency.  If you want, I can also collapse all of the commits into a single commit.  I didn't want to change it without notice if you were already looking at it.

src/isolineFeature.js
```diff
@@ -0,0 +1,891 @@
...
+  /**
+   * Compute the positions for labels on each line.  This can be called to
+   * recompute label positions without needign to recompute isolines, for
+   * instance when the zoom level changes.  Label positions are computed in the
+   * map gcs coordinates, not interface gcs coordinates, since the interface
+   * gcs may not be linear with the display space.
+   *
+   * @returns {this}
+   */
+  this.labelPositions = function () {
```
aashish24: Do we need this as a public function?

src/isolineFeature.js
```diff
@@ -0,0 +1,891 @@
...
+ * @property {geo.meshFeature.meshInfo} mesh The normalized mesh.
+ * @property {array[]} lines An array of arrays.  Each entry is a list of
+ *    vertices that also have a `value` property with the appropriate entry in
+ *    `values`.  If the line should show a label, it will  also have a `label`
+ *    property with the text of the label.
+ * @property {boolean} hasLabels `true` if there are any lines that have
+ *    labels that need to be shown if there is enough resolution.
+ */
+
+/* This includes both the marching triangles and marching squares conditions.
```
aashish24: Nice!

src/isolineFeature.js
```diff
@@ -0,0 +1,891 @@
...
+  };
+
+  /**
+   * Get the last map position that was used for generating labels.
+   *
+   * @returns {object} An object with the map `zoom` and `center` and the
+   *    `labelViewport` used in generating labels.  The object may have no
+   *    properties if there are no labels.
+   */
+  this.lastLabelPositions = function () {
```
aashish24: Same here, not entirely sure if we need this in the public API

src/meshFeature.js
```diff
@@ -349,20 +345,37 @@ var meshFeature = function (arg) {
...
-          remap[i] = -1;
+      for (i = 0; i < numPts; i += 1) {
+        idx = result.index ? result.index[i] : i;
+        if (!vertexValueFuncs.used(data[idx], idx)) {
+          break;
         }
       }
-      result.index.splice(usedPts);
-      if (usedPts !== numPts) {
+      if (i !== numPts) {
```
aashish24: whats changed here?

src/isolineFeature.js
```diff
@@ -0,0 +1,891 @@
...
+ * @property {number} [max] Maximum isoline value.  If unspecified, taken from
+ *    the computed maximum of the `value` style.
+ * @property {number} [count=15] Approximate number of isolines shown through
+ *    the value range.  Used if `spacing` or `values` is not specified.
+ * @property {boolean} [autofit=true] If `count` is used to determine the
+ *    isolines, and this is truthy, the isoline values will be round numbers.
+ *    If falsy, they will include the exact minimum and maximum values.
+ * @property {number} [spacing] Distance in value units between isolines.
+ *    Used if specified and `values` is not specified.
+ * @property {number[]|geo.isolineFeature.valueEntry[]} [values] An array of
```
aashish24: Should we mention that either you can use min,max and count or values? I think it is mostly clear, just wondering if we could use bit more strong wording.

src/isolineFeature.js
manthey: Yes.  By default, we recompute local positions after some amount of zooming or panning.  If a user wants the labels to be updated on different criteria, they would call this function.

src/isolineFeature.js
manthey: Yes.  As with `labelPositions`, this information can be used to decide when to regenerate labels if the default is inappropriate.

src/meshFeature.js
manthey: Some of the mesh feature was refactored to use less memory by not duplicate arrays if not needed.  `usedPts` is no longer computed while generating a preliminary array, as by skipping that array things are faster.  If some vertices are unused, then the previous for loop will have stopped before reaching `numPts`.  In this case we go into the next conditional and exclude the unused vertices.

src/isolineFeature.js
manthey: I hope between what is written and the tutorials everything is clear enough.  Since `values` takes precedence, it doesn't mentioned `spacing` or `count`, whilst those do mention `values`.

PR #843, opened 2018-06-13
vtk-js integration branch

sankhesh:
This branch adds support for a vtkjs layer.

### DIFF
--- a/examples/points/main.js
+++ b/examples/points/main.js
@@ -82,18 +82,18 @@ $(function () {
   var svgColor = 'blue';
 
   // Generate some data for vgl
-  var data = d3.range(2).map(function (i) {
-    return {
-      x: -95,             // longitude
-      y: 39.5 + 4.5 * i,  // latitude
-      c: vglColor,        // fill color
-      opacity: 0.1        // fill opacity
-    };
-  });
+  // var data = d3.range(2).map(function (i) {
+  //   return {
+  //     x: -95,             // longitude
+  //     y: 39.5 + 4.5 * i,  // latitude
+  //     c: vglColor,        // fill color
+  //     opacity: 0.1        // fill opacity
+  //   };
+  // });
   // makePoints(data, vglLayer, vglColor);
 
   // Generate some data for svg
-  data = d3.range(2).map(function (i) {
+  var data = d3.range(1).map(function (i) {
     return {
       x: -101,            // longitude
       y: 39.5 + 4.5 * i,  // latitude
--- a/examples/points/main.js
+++ b/examples/points/main.js
@@ -52,25 +52,25 @@ $(function () {
   });
 
   // Create an osm layer with custom tile url for a white background.
-  map.createLayer(
-    'osm',
-    {
-      url: function () {
-        return 'white.jpg';
-      }
-    }
-  );
+  // map.createLayer(
+  //   'osm',
+  //   {
+  //     url: function () {
+  //       return 'white.jpg';
+  //     }
+  //   }
+  // );
 
-  // Create a gl feature layer
-  var vglLayer = map.createLayer(
-    'feature',
-    {
-      renderer: 'vgl'
-    }
-  );
+  // // Create a gl feature layer
+  // var vglLayer = map.createLayer(
+  //   'feature',
+  //   {
+  //     renderer: 'vgl'
+  //   }
+  // );
 
   // Create an svg feature layer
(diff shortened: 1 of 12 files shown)


sankhesh:
@aashish24

jeffbaumes:
Cool! What does this mean exactly? Can someone attach a screenshot to the PR?

sankhesh:
@jeffbaumes This PR introduces a `vtk-js` renderer to `geo.js`.

In the screenshot that follows, the map is rendered using geo.js and the red spheres are rendered by vtk-js. The color and opacity of the spheres is controlled via vtk-js `vtkProperty` API.

screenshot from 2018-06-14 08-59-49

examples/points/main.js
```diff
@@ -52,28 +52,28 @@ $(function () {
...
   // Create an osm layer with custom tile url for a white background.
-  map.createLayer(
-    'osm',
-    {
-      url: function () {
-        return 'white.jpg';
-      }
-    }
-  );
+  // map.createLayer(
```
manthey: Can you make this a new example rather than modifying the existing points example?

sankhesh: Yep, I need to clean this up.

jeffbaumes:
Nice, so they are actually 3D spheres? Is there any way to know that they are spheres?

aashish24:
> Nice, so they are actually 3D spheres? Is there any way to know that they are spheres?

@sankhesh can you add lighting in your test (just for demo/screenshot?). 

@jeffbaumes they are. Currently we use parallel project by default and that is why they look flat. We will be adding a tiltable option so you could tilt the map and see them in 3D. 

Once we do some more work, our hope with vtk.js backend is that we can leverage vtk for most of the low-level work. Also, VTK backend will enable us to do 2.5D and 3D for other geospatial data formats as well with or without map.

sankhesh:
@manthey This branch is ready for your review.

It needs:
- fixes for webpack build with vtk-js
- automatic computation of camera clipping range
- a baseline image for the new example

Lights still do not work properly with vtk-js. I will look into it on another branch.

aashish24:
@sankhesh thanks. You may have to re-write this comment (WIP: pass camera parameters to vtk). Also, I think you may have to update this branch with current master or @manthey could do that.

package.json
```diff
@@ -16,6 +16,13 @@
     "hammerjs": "^2.0.8"
   },
   "devDependencies": {
+    "babel-core": "^6.0.0",
```
zachmullen: Were the babel dependencies added to accommodate vtk.js? Is its dist package not ES5? @jourdain

src/index.js
```diff
@@ -84,6 +84,7 @@ module.exports = $.extend({
   d3: require('./d3'),
   gl: require('./gl'),
   canvas: require('./canvas'),
+  vtkjs: require('./vtkjs'),
```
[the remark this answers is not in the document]
sankhesh: Ah. I see. You're right then - I am pulling in all of vtk-js.

src/gl/vglRenderer.js
```diff
@@ -244,6 +244,10 @@ var vglRenderer = function (arg) {
       }
       m_lastZoom = map.zoom();
       cam.setViewMatrix(view, true);
+
+      console.debug(view);
+      console.debug(proj);
```
zachmullen: remove

package.json
jourdain: We use preset-env and target only the last 2 version of browsers. So there is some ES6 left that don't get transpiled such as arrow function depending on browser support.

jourdain: I mean I don't know about that project, I was talking about vtk.js.

webpack.config.js
```diff
@@ -43,15 +44,16 @@ module.exports = {
   },
   externals: {
     d3: 'd3',
-    hammerjs: 'Hammer'
+    hammerjs: 'Hammer',
+    vtkjs: 'vtk.js'
```
zachmullen: @aashish24  This is how vtk.js is treated as an optional dependency in the same way that d3 is optional for geojs. There are other ways to do it that can be done at build time rather than runtime if we would prefer that route.

package.json
zachmullen: Thanks, that's what I was asking about

aashish24:
@sankhesh @zachmullen I believe we can close this one since you have the new one opened?

sankhesh:
Closing in favor of #893

PR #851, opened 2018-06-20
Refactor how we fetch external data.

manthey:
Prior to this, a custom cmake script (girder.cmake) was used via a direct URL for each external data file.  After this change, we use cmake's build-in ExternalData module.  This continues to get all external data files listed and automatically untar any .tgz files.

External data file references have been moved from testing/test-data to tests/external-data.  Eventually I'd like there to be only one of the tests and testing directories, as having two makes things confusing.

This also deletes some abandoned cmake scripts and removes some unused data files.

As a benefit, data files are now obtained by sha512, which should guarantee that we have the correct file, even if it changes.  Before, the download url was used.  This makes it so that we only need one entry in the repo per data file instead of two.  The only downside is that ExternalData emits a -hash-stamp file in the download directory.


manthey:
The old code was written by @jbeezley, I think.  The new methods I adapted from other Kitware projects, but that was weeks ago and I don't remember which ones.

PR #854, opened 2018-06-22
Switch to webpack v3.

manthey:
There are some sections for webpack v4 which are commented out (and some sections for v3 that would need to be removed).  When karma-webpack v4 is official, we may want to switch to version 4.


external.config.js
```diff
@@ -1,7 +1,12 @@
 var path = require('path');
-var webpack = require('webpack');
+var UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 
 module.exports = {
+  /* webpack 4
```
aashish24: I believe it should be "webpack 3" - assuming this was left over

external.config.js
```diff
@@ -1,7 +1,12 @@
...
+  mode: 'production',
+   */
+  performance: {hints: false},
```
aashish24: what does this do?

external.config.js
```diff
@@ -18,18 +23,41 @@ module.exports = {
...
+      parallel: true,
+      uglifyOptions: {
+        compress: true,
+        comments: /@(license|copyright)/
+      },
+      sourceMap: true
     })
   ],
+  /* end webpack 3 */
+  /* webpack 4
```
aashish24: are we keeping commented code for the future work?

karma-base.js
```diff
@@ -342,13 +360,17 @@ module.exports = function (config) {
       'jasmine', 'sinon'
     ],
     webpack: {
+      /* webpack 4
```
aashish24: same here.. I thought it should be 3

karma-cov.conf.js
```diff
@@ -41,13 +41,11 @@ module.exports = function (config) {
...
     ]
   };
-  karma_config.webpack.module.preLoaders = [
-    {
-      test: /\.js$/,
-      include: path.resolve('src/'),
-      loader: 'istanbul-instrumenter'
-    }
-  ];
+  karma_config.webpack.module.rules.unshift({
```
aashish24: what is the difference between preloaders vs unshift?

aashish24: looks like API change.

webpack-examples.config.js
```diff
@@ -5,51 +5,55 @@ var path = require('path');
...
 plugins.push(new StringReplacePlugin());
 
 var resolve = {
-  extensions: ['.js', '.css', '.pug', ''],
+  extensions: ['.js', '.css', '.pug'],
   alias: base.resolve.alias
 };
 
 module.exports = {
+  /* webpack 4
```
aashish24: 3?

webpack.config.js
```diff
@@ -19,6 +20,12 @@ var define_plugin = new webpack.DefinePlugin({
 });
 
 module.exports = {
+  /* webpack 4
```
aashish24: 3 or 4, I would just check on it, not sure if this is intentional or left over code.

external.config.js
manthey: There is some preliminary work to use webpack 4 that is commented out.  I can remove it, but wanted to preserve it in some commit for when we move to webpack 4.  Unfortunately, karma-webpack is not webpack 4 compatible *yet*.

external.config.js
manthey: It suppresses certain warning messages.  Webpack, for instance, complains that the proj4 inclusion is large and could be broken into smaller modules, but we want proj4 as it stands, so this is not a useful warning.  Rather than see this on every build, this turns those warnings off.

karma-cov.conf.js
manthey: `unshift` just adds to the beginning of an array.  There are no longer preloaders in webpack 3; you just need to order your loaders correctly.  This places the loader first.

external.config.js
aashish24: that is fine. I was just making sure.

external.config.js
aashish24: thanks

Issue #853, opened 2018-06-22
Improve point clustering interface


manthey:
Currently, clustering points takes a radius, maxZoom, width, and height.  The clustering occurs in feature gcs (usually the same as the map interface gcs), and is cryptically dependent on the width and height of the viewport.  It would be more obvious if clustering could be specified in display pixels or in the map's display gcs and had no dependency on the viewport size.

manthey:
Closed by #983.

PR #858, opened 2018-07-02
Update documentation for pointFeature.

manthey:
This also touches documentation on util/clustering.js, though that could use some refactoring so that clusters are computed in map gcs rather than in feature gcs (usually interface gcs) and the cluster size is given in simpler units (such as display pixels).

This fixes a number of trivial documentation issues in other files, too.


src/gl/pointFeature.js
```diff
@@ -365,13 +390,33 @@ var gl_pointFeature = function (arg) {
...
    */
   this.verticesPerFeature = function () {
     var unit = pointPolygon(0, 0, 1, 1);
     return unit.length / 2;
   };
 
+  /**
+   * Set style(s) from array(s).  For each style, the array should have one
+   * value per data item.  The values are not converted or validated.  Color
+   * values should be {@link geo.geoColorObject}s.  If invalid values are given
```
aashish24: is "s" after the "}" intentional?

src/pointFeature.js
```diff
@@ -2,13 +2,59 @@ var inherit = require('./inherit');
...
+ *
+ * @typedef {object} geo.pointFeature.clusteringSpec
+ * @property {number} [radius=0.05] This is combined with the `width` and
+ *   `height` to determine how close points need to be to each other to be
+ *   clustered.
+ * @property {number} [maxZoom=18] Never cluster above this zoom level.
+ * @property {number} [width=256]
+ * @property {number} [height=256]
+ */
+// TODO: refactor point clustering to (a) not be based on window size, (b) use
```
aashish24: perhaps we could create an issue for this and remove it from here?

src/gl/pointFeature.js
manthey: Yes.  It makes the docs read 'Color values should be `geo.geoColorObject`s.', which seemed more grammatical, but I'm not committed to the 's'.

src/pointFeature.js
manthey: Sure.

manthey: Apparently I already did (see issue #853).  Do you want it removed?  It stands as a warning to anyone using point clustering.

src/gl/pointFeature.js
aashish24: I am fine with this. You could have said that 'Color' values should a a geo.geoColorObject.

src/pointFeature.js
aashish24: I would remove it as it is not consistent and a user would not dig in the code as much I think.

src/gl/pointFeature.js
manthey: I reworded it a little.

src/pointFeature.js
manthey: Done.

PR #866, opened 2018-07-12
Add more support for testing examples.

manthey:
Prior to this, all example testing was done via image comparison, which is relatively heavy to add.  This mimics the testing support in tutorials, where examples can test if their dom or internal state matches expected values.  These are different tests than image comparison, but should facilitate adding basic tests to most examples.

Also, this now loads an example's javascript at the end of the body rather than at the end of the head.  This allows skipping the onready callback.


manthey:
This closes #329, as the infrastructure is fully in place.

PR #869, opened 2018-07-13
Add a rainfall example.

manthey:
This uses isolines and contours to animate annual rainfall in the United States based on NOAA data.

Fix a minor edge case in the contour feature where values near the top of the range could have a faint defect due to the way the sampler2d is used.


src/feature.js
```diff
@@ -646,17 +648,24 @@ var feature = function (arg) {
...
+   *    current bin number.  If `null`, the bin is dynamically computed based
+   *    on order within the parent.  If children are nested, this may not be
+   *    what is desired.
    * @returns {number|this} The current bin number or a reference to `this`.
    */
   this.bin = function (val) {
     if (val === undefined) {
+      if (m_bin === null) {
+        var parent = this.parent(),
+            idx = parent ? parent.children().indexOf(m_this) : -1;
```
aashish24: why not just set it to 0 (instead of -1) here. Also, if we are returning a value even if m_bin is null, should we store this value in m_bin before returning it?

src/gl/contourFeature.js
```diff
@@ -103,7 +103,9 @@ var gl_contourFeature = function (arg) {
           '    } else {',
           '      step = valueVar;',
           '    }',
-          '    clr = texture2D(sampler2d, vec2(step / steps, 0.0));',
+          // our texture is padded on either end by a repeated value to ensure
+          // we interpolate smoothly at the ends.
+          '    clr = texture2D(sampler2d, vec2((step + 1.0) / (steps + 2.0), 0.0));',
```
aashish24: can we not use the GL texture GL_CLAMP_TO_EDGE which would probably give us the same behavior?

src/feature.js
manthey: `indexOf` returns -1 if the value isn't present, so return -1 here results in the same result if there is no parent or this is somehow not a child of its own parent.  Although we then clamp it to zero, I felt handling the missing parent case identically to a missing entry case was clearer then doing otherwise.

No, we shouldn't store the value in `m_bin`.  `null` is the special meaning of return the position based on where it is in the child list.  If you had two children, A, B, both with null m_bin, they would have bin values of 0, 1.  If you inserted a child, so you have A, C, B, they would then have bin values of 0, 1, 2 (so B would go from 1 to 2).  This is desired, and overriding this would break that feature.

aashish24: thanks makes sense to me, given this explanation.

src/gl/contourFeature.js
aashish24: not sure, I will have a look but for now this is good.

PR #876, opened 2018-07-27
Fix the SLD example.

manthey:
Some notation was using a deprecated method of accessing an xml document.  Add a test to make sure this actually reads data from the WMS server.

This relies on a testing infrastructure change in a previous PR.


examples/sld/main.js
```diff
@@ -208,8 +208,7 @@ var utility = {
     for (var i = 0; i < count; i++) {
       this.createMapEntry(xml, palette[i], values[i], 1);
     }
-    var xmlString = (new XMLSerializer())
-      .serializeToString(xml.context);
+    var xmlString = new XMLSerializer().serializeToString(xml[0]);
```
aashish24: is xml[0] preferred now because there is no longer an accessor for the context?

manthey: Yes.  `xml[0]` references the actual xml DOM rather than the jQuery selector of the xml DOM.  The jQuery `<document>.context` accessor was deprecated and finally removed in jQuery 3.

PR #881, opened 2018-07-31
Avoid making an "untagged" tag when deploying.

manthey:
When merging to master, travis is generating an "untagged-<hash>" tag.  I think this might avoid that.


manthey:
@matthewma7 I'm not sure how to test this without merging to master.

PR #892, opened 2018-08-09
Improve the jsdoc index page.

manthey:
This alters the jaguarjs-jsdoc template by replacing the mainpage and other templates and adding some css.  The resulting index page lists the GeoJS version and has links to all classes and typedefs.


aashish24:
It looks great.

PR #894, opened 2018-08-14
Add an idle property to objects.

manthey:
For any object that inherits from geo.object, the `idle` property is true if `onIdle` would call a handler immediately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,12 @@
 
 ## Unreleased
 
+### Features
+- Added an idle property to objects (#894)
+
 ### Changes
 
 - Changed build process: optional dependencies are now included in the bundle by default (#890)
-
 
 ## Version 0.17.0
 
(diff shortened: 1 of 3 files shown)


matthewma7:
Thanks @manthey 
I tried to flood the request, but it seems the queue in the Network tab won't go too long already. Is the osmLayer internally tracking pending requests already?
This PR is good as itself, but do you think I still need it?

manthey:
The tile layer will never have more than 6 tiles pending at a time (this number is configurable).  When you switch url, I don't know if it will queue a new set of 6 tiles before the extant ones complete -- the browser will block if they are from the same server.

I think the PR adds a nice property, so I'd say it should still be added.

matthewma7:
Yes, I observe what you said, the queue won't grow crazy.

This PR is good. Thanks for prompt improvement upon request 😉

PR #897, opened 2018-08-14
Clean up unused code from CMakeLists

zachmullen:
@manthey it's possible this is a bad idea, maybe I don't fully understand the purpose of the code I deleted.


CMakeLists.txt
```diff
@@ -95,8 +83,6 @@ foreach(data_file ${data_files_list})
     )
   endif()
 endforeach()
-# Create a test that gets this target
-add_test(NAME get_data_files COMMAND "${CMAKE_COMMAND}" --build "${CMAKE_BINARY_DIR}" --target data_files)
```
manthey: I think this downloads the data files without having to explicitly run `make`.  Tests that add this as a property ensure that this runs before those tests.

CMakeLists.txt
```diff
@@ -29,18 +29,6 @@ if(NOT NPM_EXECUTABLE)
...
-            ${CMAKE_CURRENT_BINARY_DIR}/test/geojs_test_runner.py
-            ${ARGN}
-  )
-  set_property(TEST "${test_name}" APPEND PROPERTY DEPENDS "notes-reset")
-  set_property(TEST "notes-report" APPEND PROPERTY DEPENDS "${test_name}")
-endfunction()
-
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/tests/runners/geojs_test_runner.py.in
                ${CMAKE_CURRENT_BINARY_DIR}/test/geojs_test_runner.py
 )
```
manthey: This `configure_file` command can be deleted as well.

PR #899, opened 2018-08-16
Allow adjusting camera clipbounds.

manthey:
This was motivated by wanting to set the near and far clip values.  This can be done by `map.camera().clipbounds = {near: <near value>, far: <far value>}`.

This simplifies some of the camera code.  Before, we were generating a display matrix but separately calculating the transform in a series of steps, and similarly we were doing the inverse via a set of steps instead of a matrix multiplication.  This uses the `display` and `world` matrices explicitly in the transforms.

The `css` transform property was ambiguous about the transform origin.  We now explicitly expect a transform-origin of `0 0`, as we use that elsewhere in the project.  This adjusts several tests.

Some of the tests have be rewritten for clarity.

Some of the camera helper functions have been eliminated as they were simple wrappers around matrix functions.

There are now distinct clipbounds for perspective and parallel projections.

The css transform format was rewritten to be more compact and not lose precision in some instances.

Note that the webgl use of the perspective camera projection will need additional refactoring to handle any clipbounds.


src/camera.js
```diff
@@ -881,103 +865,49 @@ camera.projection = {
...
  */
-camera.bounds = {
-  left: -1,
-  right: 1,
-  top: 1,
-  bottom: -1,
-  far: -2,
-  near: -1
+camera.clipbounds = {
+  perspective: {
```
aashish24: typically for perspective, you use angle, aspect and near far only (e.g. see here: http://www.ntu.edu.sg/home/ehchua/programming/opengl/CG_examples.html). I think setting in terms of gcs bounds is also useful so perhaps we are fine for now.

manthey: I fully expect that the perspective mode will change further when we start using it in earnest.

PR #900, opened 2018-08-17
Use babel in the build process.

### LINKED ISSUE #418: Consider using ES6 (EMCA6) feature using Babel

manthey:
This uses babel as a webpack loader to preprocess source files.  It uses `babel-preset-env` to decide what browsers to target, using `browsers: ['defaults']` as the criterion.  Perhaps we want to explicitly target a set of browsers.

This resolves #418.


manthey:
This is probably a good point to introduce a general set of polyfills to continue to support the oldest browsers we want to handle.  Specifically, `babel-preset-env` by itself will ensure language features, but not standard library features (such as Promise support in IE 11).  To handle that, we can use `babel-polyfill` and `karma-phantomjs-shim`.  The `babel-preset-env` options can then be used to remove undesired browsers so the built geo.js package doesn't become needless large.

manthey:
There is something amiss with how coverage is being generated.  I'm looking in to it.

zachmullen:
We use `babel-plugin-istanbul` to get coverage working with babel.

manthey:
Using `babel-plugin-istanbul` or switching from `karma-coverage` to `karma-coverage-istanbul-reporter` both sort of work but have issues.  Specifically, the emitted coverage seems to reference columns that only exist in the transpile code.

manthey:
It turns out my issue with coverage was caused by using Babel 7 and, switching back to Babel 6, on where babel-polyfills was required.  I've switched to Babel 6 and only added the polyfills to the non-lean build (and also require them in testing).

manthey:
@zachmullen Can you review this again, please.

karma-cov.conf.js
```diff
@@ -34,16 +34,17 @@ module.exports = function (config) {
   karma_config.specReporter = {suppressPassed: true, suppressSkipped: true};
   karma_config.coverageReporter = {
     reporters: [
-      {type: 'html', dir: 'dist/coverage/', subdir: subdir_name},
+      // {type: 'html', dir: 'dist/coverage/', subdir: subdir_name},
```
zachmullen: Unless there's a good reason, we should probably remove unused lines rather than commenting them out.

manthey: Will do.  Thanks.

PR #902, opened 2018-08-23
Remove unneeded package installation from travis.

manthey:
Update the CHANGELOG with recent changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,20 +14,6 @@ addons:
   chrome: stable
   apt:
     packages:
-      # I suspect that not all of these are necessary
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -14,20 +14,6 @@ addons:
   chrome: stable
   apt:
     packages:
-      # I suspect that not all of these are necessary
```
aashish24: @manthey so these are now available on travis instances?

manthey: xvfb is fully available and that is sufficient for Firefox and Chrome headless.

PR #904, opened 2018-08-24
Added style typedefs.

manthey:
Moved all style specifications into jsdoc typedefs called `<class>.styleSpec`.

This also adds a link within the style function description that links to the `styleSpec` of the current class (even within classes that inherit `style`).  This resolves #70 and resolves #857 (for all features that have up to date documentation).

PR #905, opened 2018-08-27
Update the fileReader and geojsonReader.

### LINKED ISSUE #297: Improve file reader interfaces
- [ ] Create one feature of each type whenever possible
- [ ] Add on load events
- [ ] Update `map.fileReader` to allow custom style functions

### LINKED ISSUE #822: Exceptions in geojson reader
In the geojson reader, if you use a layer that does not support a necessary feature type, there is a code path which fails to raise an exception.  When an exception is raised, it is an unclear "Cannot read property 'data' of null".

Specifically, passing a valid geojson string with a polygon to the reader when using a canvas renderer will fail silently.  Passing the same geojson as an object will raise the above mentioned exception.

It would be better if both of these raised an exception like "layer does not support 'polygon' feature".  I think we should change the `featureLayer.createFeature` method to raise such an exception (it currently returns `null`).

manthey:
Update the `fileReader` and `geojsonReader` docs.

Remove a function that was untested and incorrect from the fileReader.  Add tests for the fileReader.

Rename the `jsonReader` to `geojsonReader`.  An alias called `jsonReader` is maintained to not break existing code.  The geojsonReader can take style defaults for points, lines, and polygons.

This uses newer language features, such as the spread operator.

This resolves #822.  This resolves #297, though it does not add a load event.  If that is still desired, a new issue should be made.


manthey:
This depends on PR #903 (for proper coverage) and PR #904 (for style jsdoc typedefs).

PR #910, opened 2018-08-30
Use a background color to indicate the target in jsdoc.

manthey:
This makes it easier to identify a selected section when it ends up not being at the top of the display.


manthey:
Example when geo.camera zoom is selected:

image

PR #911, opened 2018-08-30
Catch errors in animation frame callbacks

jbeezley:
Errors in user provided animation frame callbacks caused all future
callbacks in the same animation frame to not run.  This catches any
errors and prints them to `console.error`.


manthey:
@jbeezley I added a test.  If you think the test is good, feel free to merge.

jbeezley:
👍 Thanks.

PR #921, opened 2018-09-05
Add a selectionAPI='auto' mode for features.

### LINKED ISSUE #871: geoOn and selectionAPI
In my couple of times using GeoJS, I've run into this issue where I've needed to rediscover that you need to turn on the `selectionAPI` option in order for mouse events to function. I have a few suggestions:

* Make a better name for `selectionAPI` that indicates that it allows interaction.
* If `selectionAPI` is turned off, emit an warning/error/exception when attempting to perform `geoOn()` with an event that will not function. Currently it silently lets you call `geoOn()` but then does nothing.

manthey:
This is the new default.  If any feature event is bound on the feature, the selectionAPI is enabled.  Binding and unbinding feature events will update the selection.  The selectionAPI can still be a boolean to directly control the behavior.

This resolves #871.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Features
+- Feature selection API is now enabled automatically if any event handlers are bounds to the feature (#921)
+
 ### Changes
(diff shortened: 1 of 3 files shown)


aashish24:
it would be great if @waxlamp and @johnkit could review it.

PR #926, opened 2018-09-10
Add an option to the tutorial editors to keep history in the browser.

manthey:
If `keep=history` is in the tutorial's url, in addition to add the source code to the window's location, the back button will return to the previously source code of the tutorial.

I'm finding this is handy when using the editor to try out experimental code, as those experiments are in the browser's history.  However, this has been left as an option that has to be explicitly enabled, as if you make many small edits, the history is very long and tedious without a good indicator of when something significant or interesting occurred.

### DIFF
--- a/examples/common/utils.js
+++ b/examples/common/utils.js
@@ -19,9 +19,11 @@ var exampleUtils = {
    * location and history.  This will also remove undefined values from the
    * set properites of params.
    *
-   * @param {object} params: the query parameters as a dictionary.
+   * @param {object} params The query parameters as a dictionary.
+   * @param {boolean} [updateHistory] If true, update the browser history.  If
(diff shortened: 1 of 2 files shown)


manthey:
@aashish24, @johnkit, @matthewma7 Request for comments: should this be the default behavior of the editor tutorials?  Or is it too specialized and shouldn't be merged?  Opinions desired.

matthewma7:
This is a feature that I could use myself. 
I personally think there are more benifit than drawbacks. and maybe when we think there are too many unimportant small changes being pushed to history, we could a setTimeout.

manthey:
We already debounce updates to an extent to prevent rerunning the tutorials as you are typing, so it probably is fine.  If no one else chimes in, I'll make it always push to history.

PR #936, opened 2018-10-11
Memoize convertColor to add speed.

manthey:
Also, refactor util/index.js to have color utilities in their own file and other common utilities in util/common.js.


src/util/color.js
```diff
@@ -0,0 +1,336 @@
...
+/**
+ * Store the results of convertColor in dictionary for so repeated requests can
+ * be returned quickly.  If the memoization table gets over a certain size,
+ * just reset it.
+ *
+ * @param {geo.geoColor} origColor The origial color specification.
+ * @param {geo.geoColorObject} resultColor The result of the conversion.
+ * @returns {geo.geoColorObject} The `resultColor`.
+ */
+function memoizeConvertColor(origColor, resultColor) {
```
aashish24: should we call it cacheConvertColor essentially using cache instead of memoize, as it may provide consistent naming convention elsewhere, memoize is fine as well.

src/util/color.js
```diff
@@ -0,0 +1,336 @@
...
+  m_memoizeConvertColor.count += 1;
+  if (m_memoizeConvertColor.count >= m_memoizeConvertColor.maxCount) {
```
aashish24: nice to have maxCount +1

tests/cases/colors.js
```diff
@@ -115,6 +115,28 @@ describe('geo.util.convertColor', function () {
       expect(geo.util.convertColor(undefined)).toBe(undefined);
     });
   });
+  describe('Memoization', function () {
```
aashish24: for some improved reading of the test outputs, I would use Memoization Colors - as we may have more memoization in the future

src/util/color.js
manthey: I'm not sure what you mean here.  If the memoize cache has maxCount values currently, adding another value will exceed it, so we reset before adding another value.

aashish24: I meant that I liked this check

src/util/color.js
aashish24: Okay, lets go with memoization, I do not see the distinction but it seems that others may. +1

PR #940, opened 2018-10-16
Add an example showing live air traffic from the OpenSky Network.

manthey:
This example is hidden, as we probably need to inform the OpenSky Network if we make it obviously public.

This has a lot of different features:
- It uses a web worker to collect data
- Drawing color scales is optimized by overriding style.get and using a   LUT rather than a d3 scale.
- It shows off scaling to a large number of vertices (given time).
- It uses width, color, and opacity to indicate different properties.
- It can show live or static data.

### DIFF
--- a/examples/build-website.js
+++ b/examples/build-website.js
@@ -22,6 +22,17 @@ examples.map(function (json) {
       fs.removeSync(path.resolve(json.output, 'docs', 'public'));
     });
   }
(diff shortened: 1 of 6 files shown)


manthey:
This is an example I mostly wrote some time ago.  I'm making a PR now to make it easier to share.

PR #941, opened 2018-10-18
Overallocate geomBuffer.

manthey:
When reusing an existing buffer for webGL, permit it to be larger than strictly necessary.  If reallocating an existing buffer, allocate some extra space.  For dynamic data sets, this avoids reallocating buffers on every geometry update, substantially reducing garbage collection.

In one examples with ~35,000 lines containing a total of ~1,000,000 vertices where the number of vertices changes periodically, this reduced geometry update time by ~200 ms.

### DIFF
--- a/src/util/common.js
+++ b/src/util/common.js
@@ -368,14 +368,10 @@ var util = {
     }
     /* If we need to allocate a new buffer (smaller or larger), and we have an
      * existing, non-zero-length buffer, allocate a larger than needed buffer.
-     * Add an extra factor of allocateLarger, but if a power-of-two is between
-     * the specified size and the larger permitted size, perfer the power-of-
-     * two. */
+     * Add an extra factor of allocateLarger. */
     var allocate = len;
     if (data instanceof Float32Array && data.length && len && allocateLarger > 0) {
-      allocate = Math.min(
-        Math.floor((allocateLarger + 1) * len),
-        Math.pow(2, Math.ceil(Math.log(len) / Math.log(2))));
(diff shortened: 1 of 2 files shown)


src/util/common.js
```diff
@@ -343,17 +343,41 @@ var util = {
...
+    if (data instanceof Float32Array && (data.length === len || (data.length >= len && data.length <= allow))) {
       return data;
     }
-    data = new Float32Array(len);
+    /* If we need to allocate a new buffer (smaller or larger), and we have an
+     * existing, non-zero-length buffer, allocate a larger than needed buffer.
+     * Add an extra factor of allocateLarger, but if a power-of-two is between
+     * the specified size and the larger permitted size, perfer the power-of-
+     * two. */
+    var allocate = len;
```
aashish24: why do we care for power of two?

manthey: We might not.  In general, many memory allocation schemes have different pools that change when a power-of-two boundary is crossed.  I don't actually know if Firefox or Chrome do (they used to, but they may no longer do so).  By choosing a power-of-two when convenient, it is possible that we will use the memory pool for smaller objects in preference to the larger.  The optimization is a guess -- if the buffer gets reallocated bigger, then the smaller size is wasted.  If it doesn't get bigger, then we might gain an efficiency.

I can take it out if you'd prefer.

aashish24: > I can take it out if you'd prefer.

I have a slight preference to take it out since this behavior is not documents in the 'user documentation' plus it complicates things further without clear benefit.

manthey: Will do.

Issue #942, opened 2018-10-18
Share webGL contexts between layers


manthey:
Currently, having more than 15 webgl layers will exhaust the available webgl contexts.  If webgl layers do not have non webgl layers between them by z-index, a single context can be used instead.  This should be done automatically.  When there are multiple layers, the webgl renderer will need to apply opacity for each layer, rather than using the opacity on the div.

manthey:
Regarding shared webgl contexts, layering, and opacity.

When two layers share a webgl context, I intend that they will also share a depth buffer.  This will be a difference between separate contexts and shared contexts (but, when we support full 3D, I view this as a positive feature).

On tile layers, we render different levels of tiles at slightly different z values, all "just behind" their actual z value.  When we share contexts, multiple tile layers will composite as expected (with layers with higher layer z-indices just above those with lower).  If the shared context has *both* tile and feature layers, then features with the same z value as tiles will always appear above the tiles, even if the tile layer's z-index is above the feature layer's z-index.  To do otherwise, we could either avoid sharing contexts between tile and feature layers OR there will need to be slight z offsets for all layers (not just tile layers) to force them to composite properly.  I think that, for now, we just accept that when tile and feature layers share context, tiles are behind features with the same z value regardless of layer z-index.

Sharing context between layers with different opacities is tricky.  The results of changing feature opacity instead of layer opacity is not the same (for instance, a layer with 100% opacity and overlapping lines each with 50% opacity will form an area with 75% opacity, whereas a layer with 50% opacity and overlapping lines with 100% opacity will have everything at 50% opacity).  Either we will have to render each layer into a frame buffer object or not share contexts between layers with different opacities.

For automatic merging of shared webgl contexts, I think the simplest approach would be to only share contexts between adjacent webgl layers that have the same opacity (and layers with 0 opacity or 0 visibility could be shared, too, as they will skip rendering).

Feedback welcome, as I am about to add code for automatic sharing of webgl context between adjacent webgl layers.

matthewma7:
Hi @manthey, this all sounds reasonable to me. 
Are you thinking about providing an option to enable context reuse or always reuse? Just asking,  I know it could be a lot more work to make it optional.

manthey:
I am thinking that context sharing would be optional.

Currently, you can either create a layer and have it create a renderer instance or you create the layer while specifying an existing renderer instance.  In the second case, layers share the context.  This works correctly on my development branch except for honoring opacity and visibility properly.

[...]

PR #949, opened 2018-11-02
Test layer.js.

manthey:
Move the layer specification to a typedef for better documentation.  Make sure all layer types use typedefs.

Fix issues with the layer id function.

Ensures that canvas and renderer can actually be specified when creating a layer.  This allows layers to share a renderer or a canvas.  For the webgl renderer, sharing a canvas results in flashing, as each layer completely redraws the canvas.  When sharing a renderer, there is no proper support for opacity (probably not for visibility) or z-order, but this is a first step to implementing #942.

Removed some unused code.

### DIFF
--- a/src/d3/d3Renderer.js
+++ b/src/d3/d3Renderer.js
@@ -380,7 +380,7 @@ var d3Renderer = function (arg) {
           .attr('mode', 'normal');
 
       if (!arg.widget) {
-        canvas = m_svg.append('g');
+        canvas = arg.canvas || m_svg.append('g');
       }
 
       shadow = m_defs.append('filter')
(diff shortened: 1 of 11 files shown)


src/d3/d3Renderer.js
```diff
@@ -380,7 +380,7 @@ var d3Renderer = function (arg) {
           .attr('mode', 'normal');
 
       if (!arg.widget) {
-        canvas = m_svg.append('g');
+        canvas = arg.canvas || m_svg.append('g');
```
zachmullen: Should there be a docstring for this? It looks like maybe a new option.

PR #950, opened 2018-11-02
Move more specifications to typedefs.

manthey:


src/fetchQueue.js
```diff
@@ -1,30 +1,35 @@
...
  * objects in the queue completes (resolved or rejected), another item in the
  * queue is processed.  The number of concurrently processing items can be
- * adjusted.  At this time (2015-12-29) most major browsers support 6
- * concurrent requests from any given server, so, when using the queue for
- * tile images, thie number of concurrent requests should be 6 * (number of
- * subdomains serving tiles).
+ * adjusted.
+ *
+ * At this time (2018-11-02) most major browsers support 6 concurrent requests
+ * from any given server, so, when using the queue for tile images, thie number
```
waxlamp:
```suggestion
 * from any given server, so, when using the queue for tile images, the number
```

aashish24: nice catch @waxlamp

Issue #776, opened 2018-02-12
Add tests for PR 774


matthewma7:
https://github.com/OpenGeoscience/geojs/pull/774 needs some tests.

PR #955, opened 2018-11-02
Add testing to simple point tutorial

matthewma7:
This resolves https://github.com/OpenGeoscience/geojs/issues/776


manthey:
@matthewma7 The test failure happens in the PhantomJS test environment because webGL is not available and the feature layer ends up using the canvas renderer, which doesn't have the point feature.  If you change codeblock 2 to specify that the feature layer must support points (i.e.,
```
    var layer = map.createLayer('feature', {features: ['point']});
```
), then the tests will pass.

PR #959, opened 2018-11-14
Fix redirects on the website.

manthey:
The website redirects bare paths to http, but paths ending with / to the same protocol, so build example and tutorial lists with / after the example or tutorial name.


manthey:
@aashish24 Can you click the approve button (you've commented but not actually clicked it)?

PR #961, opened 2018-11-20
Update the jsdocs for the vector feature.

### LINKED ISSUE #448: Improve documentation (Tutorials, API documentation etc.)

manthey:
Tolerate no warnings from eslint.

Fix a few minor other jsdoc issues.

This is the last PR necessary to resolve #448.

### DIFF
--- a/src/vectorFeature.js
+++ b/src/vectorFeature.js
@@ -2,11 +2,47 @@ var inherit = require('./inherit');
 var feature = require('./feature');
 
 /**
- * Create a new instance of class vectorFeature
+ * Object specification for a graph feature.
+ *
+ * @typedef {geo.feature.spec} geo.vectorFeature.spec
+ * @property {geo.vectorFeature.styleSpec} [style] Style object with default
+ *   style options.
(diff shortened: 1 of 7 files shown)


src/vectorFeature.js
```diff
@@ -2,11 +2,47 @@ var inherit = require('./inherit');
...
 
 /**
- * Create a new instance of class vectorFeature
+ * Object specification for a graph feature.
+ *
+ * @typedef {geo.feature.spec} geo.vectorFeature.spec
+ * @property {geo.vectorFeature.styleSpec} [style] Style object with default
+ *   style options.
+ * @property {function|object} [origin={x: 0, y: 0, z: 0}] Origin accessor.
+ * @property {function
```
aashish24: looks like something is missing here

manthey: The line shouldn't of been there.  I've removed it.

aashish24: thanks

PR #962, opened 2018-11-20
Move width and height to the base renderer class definition.

manthey:
This reduces code duplication.  It has been listed as a TODO in the code for a long time.

### DIFF
--- a/src/d3/d3Renderer.js
+++ b/src/d3/d3Renderer.js
@@ -43,8 +43,6 @@ var d3Renderer = function (arg) {
       m_sticky = null,
       m_features = {},
       m_corners = null,
-      m_width = null,
-      m_height = null,
       m_diagonal = null,
       m_scale = 1,
       m_transform = {dx: 0, dy: 0, rx: 0, ry: 0, rotation: 0},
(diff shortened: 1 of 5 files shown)


src/d3/d3Renderer.js
```diff
@@ -209,11 +207,10 @@ var d3Renderer = function (arg) {
         width = map.size().width,
         height = map.size().height;
 
-    m_width = width;
-    m_height = height;
-    if (!m_width || !m_height) {
+    if (!width || !height) {
```
aashish24: in vglRenderer we use m_this.width() to get width and height, why not we use the same here? instead of getting it from map?

manthey: A few lines later is where the layer's width and height is set *from* these values (`m_this._setWidthHeight(width, height);`), so `width` and `height` are the values that will be set.

aashish24: I see, I was expecting the flow between vgl and d3 renderers would be same and may be they are similar. so sounds like we are getting the width and height from the map and set it to the renderer in both cases?

manthey: Yes.

PR #963, opened 2018-11-26
Ignore z coordinates on gl point and polygon features.

manthey:
We currently ignore the z coordinate on line features.  We don't support z coordinates yet, so disable them until we do.


johnkit:
I would appreciate if you would post a new release to npm once this is merged. Thanks.

PR #965, opened 2018-11-27
Rename renderers

### LINKED ISSUE #898: Rename renderers
The `d3` renderer doesn't actually require d3, and would more properly be called the `svg` renderer.  As part of renaming it, remove any dependencies that do exist on d3 (some features will still require d3, such as the color legend).

Do we also want to rename the `vgl` renderer to `webgl` or just `gl`?

manthey:
Rename the d3 renderer to svg.  Alias it to d3 so as to not break existing code.

Rename the vgl renderer to webgl.  Alias it to vgl so as to not break existing code.

Update the baseline images; many were renamed.

This resolves #898.


src/webgl/choroplethFeature.js
```diff
@@ -36,9 +36,9 @@ var gl_choroplethFeature = function (arg) {
    */
   this.draw = function () {
     m_this._update();
-    if (m_gl_polygons) {
-      for (var idx = 0; idx < m_gl_polygons.length; idx += 1) {
-        m_gl_polygons[idx].draw();
+    if (m_polygons) {
```
aashish24: +1

PR #970, opened 2018-12-05
Handle visibility between layers sharing a webgl renderer.

manthey:


aashish24:
@manthey could you explain what is the expected behavior when two layers share webgl context in relation to their visibility?

manthey:
Prior to this PR, toggling the visibility of a layer would affect the visibility of the DOM element of that layer.  When multiple layers share a webgl context, there is only a single DOM element for those layers.  In this case, the features inherit the visibility of their parent layer (while still maintaining their own visibility flag).  Setting a layer to not-visible functionally makes all of the features of the layer not-visible rather than changing the visibility of the DOM element.

PR #973, opened 2019-01-14
Fix "npm run examples".

manthey:
In the quickstart documentation we cite this as a way to serve the examples, but that has been broken for quite a while.  This uses the server.js script to serve the examples tutorials.


matthewma7:
Worked well for me. 👍

PR #975, opened 2019-02-14
Autoshare webgl renderers.

manthey:
Autosharing is now the default.  Renderers can be manually shared, and webgl layers can be manually be moved to different renderer.

To manually move a layer to a different renderer, for instance, given two layers: `var layer1 = map.createLayer('feature', {renderer: 'webgl', autoshareRenderer: false});  var layer2 = map.createLayer('feature', {renderer: layer1.renderer(), autoshareRenderer: false});`, the second layer can be moved to a new renderer by doing: `layer2.switchRenderer(geo.createRenderer('webgl', layer2), true);`.

This requires an update to vgl to work.


aashish24:
@manthey I am still reviewing this branch.. mostly done just has one or two things that I need to clear my self.

manthey:
Can you also review https://github.com/OpenGeoscience/vgl/pull/106, as this depend on it?

aashish24:
done.

PR #983, opened 2019-03-05
Refactor how point clustering radius is used.

manthey:
Before, it was cryptically linked to the display size, which was counter-intuitive.  Now, the point clustering radius is based on display pixels.

This also removes an unnecessary dependency on vgl from the clustering algorithm.

Temporarily disable the test of the sld (WMS) example.  The external server we use for that example is not responding.

This resolves issue #853.


examples/cluster/main.js
```diff
@@ -164,7 +164,9 @@ injectStyle(pointFeature, 'fillColor', 'black');
 injectStyle(pointFeature, 'strokeOpacity', 1);
 injectStyle(pointFeature, 'strokeWidth', 1);
 injectStyle(pointFeature, 'fillOpacity', 0.5);
-pointFeature.clustering({radius: 0.012});
+// Setting the cluster radius to the cluster point style radius is usually
+// a good value.
+pointFeature.clustering({radius: 10});
```
aashish24: nice @manthey this API is so much better

aashish24: API/method

PR #985, opened 2019-03-11
Fix a small bug in canvas quads with fixed scale

manthey:
This can be seen, for instance, in the deep zoom example where the lower partial tiles were not being rendered.

### DIFF
--- a/src/canvas/quadFeature.js
+++ b/src/canvas/quadFeature.js
@@ -174,8 +174,8 @@ var canvas_quadFeature = function (arg) {
         if (!quad.crop) {
           context2d.drawImage(src, 0, 0);
         } else {
-          var cropx = Math.min(src.w, quad.crop.x),
-              cropy = Math.min(src.h, quad.crop.y);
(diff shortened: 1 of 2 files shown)


src/canvas/quadFeature.js
```diff
@@ -174,8 +174,8 @@ var canvas_quadFeature = function (arg) {
         if (!quad.crop) {
           context2d.drawImage(src, 0, 0);
         } else {
-          var cropx = Math.min(src.w, quad.crop.x),
-              cropy = Math.min(src.h, quad.crop.y);
+          var cropx = Math.min(w, quad.crop.x),
```
aashish24: what is the difference between src.w and w? do they point to different values?

manthey: `w` is a local variable that was taken from `src.width` or `src.videoWidth`.  `src.w` was incorrect.

aashish24: got it, thans.

PR #1002, opened 2019-05-09
Update the glsl validator link for travis.

manthey:
According to https://github.com/KhronosGroup/glslang, our previous link should have always worked, but it no longer does.  This pins to a specific version.  We can revert it if the original link is fixed.


manthey:
The original link has been fixed, so closing this PR.

PR #1011, opened 2019-07-10
Add a third state for autoshareRenderer.

manthey:
The default (`true` setting) will only autoshare if it is likely to not result in a visual difference (layer opacity 0 or 1, tile layers below non-tile layers).  The `"more"` setting will perform as before this PR -- autosharing if the opacities are identical and irrespective of the mixing of tile and non-tile layers.  `false`, as before, won't autoshare.

This also adds a map level setting of `autoshareRenderer` which sets the defaults for newly created levels.


matthewma7:
The new tri-state sounds great!

PR #1016, opened 2019-08-21
Improve webgl lineFeature and polygonFeature.

manthey:
Reduce memory transfer to the gpu when only styles have changed in lines or polygons.

Rename some internal variables in the line feature to make it clearer.

Use constants to avoid array lookups in a few spots.


src/webgl/lineFeature.js
```diff
@@ -126,19 +134,19 @@ var webgl_lineFeature = function (arg) {
         strokeOffsetFunc = m_this.style.get('strokeOffset'), strokeOffsetVal,
         miterLimit = m_this.style.get('miterLimit')(data),
         antialiasing = m_this.style.get('antialiasing')(data) || 0,
-        order = m_this.featureVertices(),
+        order = m_this.featureVertices(), orderk0, prevkey, nextkey, offkey,
         orderLen = order.length,
-        posBuf, prevBuf, nextBuf, farBuf, flagsBuf, indicesBuf,
+        posBuf, prevBuf, nextBuf, farBuf, flagsBuf,
```
aashish24: @manthey would it be possible to add some note here on each buffer type specifically the flags buff?

src/webgl/lineFeature.js
```diff
@@ -25,8 +25,10 @@ var flagsLineJoin = {  // uses 3 bits with flagsLineCap
   round: 6,
   'miter-clip': 7
 };
-var flagsNearLineShift = 2, flagsFarLineShift = 5;
-var flagsNearOffsetShift = 8;  // uses 11 bits
+var flagsNearLineShift = 2, flagsNearLineMult = 1 << flagsNearLineShift;
```
aashish24: Same here, if we add some code comment on how these flags are used, it would be be easy to remember and understand the code

manthey: @aashish24 I've added more to the comment above all of the flag definitions.

src/webgl/lineFeature.js
aashish24: thanks

PR #1020, opened 2019-08-28
Tile source changes

### LINKED ISSUE #60: Add API for setting named tile sets
Like openstreetmaps, mapquest, etc.

### LINKED ISSUE #1017: Default tile server for OSM is very slow
Lately, the tile server at https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png has been very slow, which frequently causes the tutorial tests to fail.  We proxy this for other tests, but since it is useful to know if the tutorials are working, it is better to not do so for the tutorial tests.

Either we should switch our default tile server (perhaps to stamen toner-lite), or we need to change what we test for the tutorials.

manthey:
Add an osmLayer source method.

This lets multiple tile sources be defined with a set of properties to make it easier to specify or switch between such sources.  For instance, to use Stamen toner-lite, `osmLayer.source('stamen-toner-lite')` will appropriately set url, subdomains, attribution, minLevel, and maxLevel.  This resolves #60.

This switches the default tile server to stamen-toner-lite.  This resolves #1017.


aashish24:
@manthey not sure where there is a conflict on CHANGELOG but the changes elsewhere looks good.

PR #1023, opened 2019-09-05
Optimize a transform code path.

manthey:
This is used when drawing lines and polygons in pixel coordinates.


aashish24:
there is conflict in change log but other than that the changes looks good.

PR #1030, opened 2019-09-15
Expose the original event data in the action wheel event.

### LINKED ISSUE #1024: geo.event.actionwheel doesn't include wheel delta
I tried to use the `geo.event.actionwheel` to capture the mouse wheel event. However, the event doesn't provide the `deltaX` and `deltaY` information as provided by the MouseWheel event (https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent)

manthey:
This was intended, but since the `event` key was used, and that is populated with the event name, it wasn't actually exposed.

Resolves #1024.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Bug Fixes
 - Mouse wheel events didn't recompute gcs coordinates, so a wheel event without a previous move event coult list the wrong location (#1027)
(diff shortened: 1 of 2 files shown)


manthey:
No.  This adds `eventData`.  It was using `event`, but that was being overridden by the name of the event and so using `event` functionally did nothing.

PR #1031, opened 2019-09-19
Add auto primitiveShape to the webgl point feature.

manthey:
The new default primitiveShape, `auto`, uses a sprite (a webgl point) if points are smaller than the GPU's maximum point size and a triangle if any are larger.  Specific primitive shapes can still be specified, as there are performance reasons for each of them and auto detection has some time cost.

### DIFF
--- a/src/webgl/pointFeature.js
+++ b/src/webgl/pointFeature.js
@@ -220,7 +220,9 @@ var webgl_pointFeature = function (arg) {
       fillVal = fillFunc(item, i) ? 1.0 : 0.0;
       fillOpacityVal = fillOpacityFunc(item, i);
       fillColorVal = fillColorFunc(item, i);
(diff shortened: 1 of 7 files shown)


src/webgl/pointFeature.js
```diff
@@ -222,6 +220,9 @@ var webgl_pointFeature = function (arg) {
       fillVal = fillFunc(item, i) ? 1.0 : 0.0;
       fillOpacityVal = fillOpacityFunc(item, i);
       fillColorVal = fillColorFunc(item, i);
+      if (m_primitiveShapeAuto && ((fillVal && fillOpacityVal) || (strokeVal && strokeOpacityVal)) && radiusVal + (strokeVal && strokeOpacityVal ? strokeWidthVal : 0) > maxr) {
```
aashish24: this line is probably too long and need to be split for consistency

manthey: I split it to multiple lines.

PR #1033, opened 2019-09-24
Change the point shaders so that common code is in shared files.

manthey:
The shader-loader allows importing chunks into other files.  This allows common code to be placed in a shared file and imported by other files.  There are some restrictions: the common chunks must have filenames ending in `glsl`, and the glslangValidator doesn't handle these inclusions on its own.  However, sharing code is of substantial benefit for reducing maintenance issues, so it is worth this change.

Note that there are other webpack shader loaders (e.g., webpack-glsl-loader) that provide similar include functionality but with a different syntax, as glsl doesn't have its own include statements.  As such, if we ever switch the shader loader to another one, the import statements would have to change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,11 +53,12 @@ before_script:
   - export DISPLAY=:99.0
 
 script:
+  # Use the glslangValidator to validate all .frag and .vert files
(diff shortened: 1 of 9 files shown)


aashish24:
@manthey I am looking into this but I am assuming that this is not needed in the 0.19.7

manthey:
@aashish24 Correct.  This just makes maintaining shaders easier when they share code (which is desired for the marker feature).

Issue #1036, opened 2019-10-01
Deepzoom example no longer working


manthey:
The deepzoom example uses an image from NASA's mars website.  The URLs for these images have changed and now have CORS that prevent loading them.

We either need to disable the deepzoom example or find a different public server.

PR #1037, opened 2019-10-01
Disable Deepzoom Example

manthey:
NASA's deepzoom server no longer allows cross-site usage.

Disable the example for now; otherwise, CI is broken.

Reference #1036.


manthey:
@aashish24 We need to fix this so other CI will pass.  Unless you know another deepzoom server off-hand, I recommend we merge this and revisit finding such a server sometime later.

PR #1042, opened 2019-10-25
Fix an issue with Chrome and hidden tabs.

manthey:
When the webpage is on a hidden tab, tiles are not rendered.  This seems to only be a problem with Chrome.  This adds 'hidden' and 'unhidden' events, so any feature can hook into these.

The webgl tile layer refreshes its tile cache when unhidden.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 - Removed extra calls to sceneObject constructors (#1039)
+- Fixed an issue with rendering on hidden tabs in Chrome (#1042)
 
 ## Version 0.19.7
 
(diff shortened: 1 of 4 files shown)


manthey:
This doesn't have automatic test coverage; we'd have to add quite a bit of code to have Chrome with multiple tabs and code coverage.

You can test this manually -- on master, load the tile example.  One a developer console as a separate window.  Switch to another tab.  In the developer console, hit reload.  Wait ~10 seconds for the hidden tab to finish loading tiles.  When you switch back to the tab with the tile example, the map will be blank.

It is a bit tougher to show off the corrected code without building locally -- the tutorial can't be used since it gets refreshed properly after being hidden since it is in an iframe.

This is a problem seen in actual deployments -- open a tab and switch away before the map finishes loading; switching back and those tiles that were loaded while hidden will be blank.

PR #1052, opened 2020-04-16
Speed up the pointInPolygon function.

manthey:
This relies on the distanceToPolygon2D function, but we don't actually care about the distance, so skip some calculations.

### DIFF
--- a/src/util/common.js
+++ b/src/util/common.js
@@ -74,7 +74,7 @@ var util = {
         return false;
       }
     }
-    return util.distanceToPolygon2d(point, inner ? {outer: outer, inner: inner} : outer) > 0;
+    return util.distanceToPolygon2d(point, inner ? {outer: outer, inner: inner} : outer, true) > 0;
   },
 
   /**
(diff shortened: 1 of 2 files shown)


src/util/common.js
```diff
@@ -693,14 +695,14 @@ var util = {
       if (((p0.y > pt.y) !== (p1.y > pt.y)) && (pt.x < (p1.x - p0.x) * (pt.y - p0.y) / (p1.y - p0.y) + p0.x)) {
         inside = !inside;
       }
-      distSq = util.distance2dToLineSquared(pt, p0, p1);
+      distSq = onlySign ? 1 : util.distance2dToLineSquared(pt, p0, p1);
```
aashish24, line 698: so if onlySign is true, the function will return distSq == 1? I did not look at the entire block, just the diff.

manthey: distSq will be 1 (because we don't care about the actual value of the distance), but the returned result will be -1 or +1 depending on the value of the `inside` flag.

aashish24: thanks makes sense.

PR #1066, opened 2020-06-08
More flexible default position accessors.

### LINKED ISSUE #1065: Be more flexible in position accessors
Position accessors have been expected to return an object with `x` and `y` and, optionally, `z`.  In a few spots we also allows position accessors to return an array that is either 2 or 3 numbers.  For consistency and flexibility, allow position accessors to return either of those universally.

manthey:
Before, most position accessors have assumed the data contains an object with `x` and `y` and, optionally, `z`.  Now, for lines, points, polygons, text, meshes, and paths, the default position accessor will also handle a data entry that is an array that is either 2 or 3 numbers.

Closes #1065.


aashish24 (approved):
Looks good.

PR #1072, opened 2021-02-25
Update testing baseline images.

manthey:
There are minor shifts due to browser changes.

This also updates package versions for building the web site to work with newer versions of node.


manthey:
Merging this so testing on other branches can pass.

PR #1073, opened 2021-02-25
Fix an issue with affine transforms

manthey:
Affine transforms would be skipped in certain cases where an optimization path was followed.  Detect this condition and don't use the simpler optimized path when affine transforms are present.

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -20,6 +20,7 @@ var maxTransformCacheSize = 10;
 /* A RegExp to detect if two transforms only different by the middle axis's
  * direction. */
 var axisPattern = new RegExp('^(.* |)\\+axis=e(n|s)u(| .*)$');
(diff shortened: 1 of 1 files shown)


aashish24:
@manthey what is the context here? Did we hit a case that you could mention here?

manthey:
Brandon discovered that webgl polygons weren't being affected by an affine transform while text features were.  The specific problem is that there is a code optimization path for certain transforms to avoid function calls when possible.  If a 3d flat array of coordinates is to be transformed, and we know the transforms are "simple", we can pass it through proj4 using a simpler path that is faster.  However, when an affine transform is specified, we can't use this simpler path.

aashish24:
Okay makes sense, thanks for the context

aashish24 (approved):

PR #1074, opened 2021-03-01
Support mix-blend-mode compositing in screenshots.

manthey:
There are limitations to the current screenshot function.  If layers are composited with mix-blend-mode, svg filters, or some other options, these are not applied.  mix-blend-mode support is added here; other limitations remain.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -1820,13 +1820,13 @@ var map = function (arg) {
           if (layer.renderer() && layer.renderer().api() === 'webgl') {
             layer.renderer()._renderFrame();
           }
-          drawLayerImageToContext(context, opacity, canvasElem, canvasElem[0]);
+          drawLayerImageToContext(context, opacity, canvasElem, canvasElem[0], layer.node().css('mix-blend-mode'));
         });
       });
       if ((layer.node().children().not('canvas').length || !layer.node().children().length) && (!layer.renderer() || layer.renderer().api() !== 'webgl')) {
         defer = defer.then(function () {
           return util.htmlToImage(layer.node(), 1).done(function (img) {
-            drawLayerImageToContext(context, 1, $([]), img);
+            drawLayerImageToContext(context, 1, $([]), img, layer.node().css('mix-blend-mode'));
           });
         });
       }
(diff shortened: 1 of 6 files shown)


manthey:
As an explicit use case, we have a project where a hyperspectral image (fluorescence microscopy) is shown with one channel per layer.  We let the users adjust the color and scaling of each channel in a separate layer and composite them via the `mix-blend-mode: lighter` option.  We want to take a screenshot of the current settings, and this change is needed to composite the results correctly.

aashish24:
Doing the review now

src/map.js
```diff
@@ -1963,10 +1963,14 @@ var map = function (arg) {
    * @param {number} opacity The opacity in the range [0, 1].
    * @param {object} elem A jQuery element that might have a transform.
    * @param {HTMLImageObject} img The image or canvas to draw to the canvas.
+   * @param {string} [mixBlendMode] the mix-blend-mode used to add this layer.
    * @private
    */
-  function drawLayerImageToContext(context, opacity, elem, img) {
+  function drawLayerImageToContext(context, opacity, elem, img, mixBlendMode) {
     context.globalAlpha = opacity;
+    if (mixBlendMode) {
```
aashish24, line 1971: would it be not possible to just have the user pass any one type of globalCompositeOperation (source in/out, over etc?)

tests/headed-cases/blog-lines.js
```diff
@@ -62,6 +62,6 @@ describe('blog-lines example', function () {
     // this permits a large delta to pass on CI.  It visually is rendered
     // correctly, though with seemingly different aliasing choices by the
     // renderer
-    imageTest.imageTest('exampleBlogLines10k', '#map', 0.04, done, null, 1000, 2, '.geojs-map.ready');
+    imageTest.imageTest('exampleBlogLines10k', '#map', 0.04, done, null, 5000, 2, '.geojs-map.ready');
```
aashish24, line 65: why this increase to 5000?

src/map.js
manthey: Yes, any of the globalCompositeOperation values are supported.  They are called mix-blend-mode in css and globalCompositeOperation in canvas.

tests/headed-cases/blog-lines.js
manthey: Travis was timing out on this.  I don't know why, but increasing it allowed the tests to pass.  It is like they are using a different mesa library than before that is slower, but that is just speculation.

aashish24 (approved):

PR #1083, opened 2021-04-22
Pin colorbrewer to 1.3.0.

manthey:
Version 1.5.1 has changed how it is bundled, which no longer works with out current examples.  Until that is refactored, pin to the last known working version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "bootswatch": "^3.4.1",
     "bowser": "^1.9.4",
     "codemirror": "^5.54.0",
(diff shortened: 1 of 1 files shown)


aashish24 (approved):
curious what has changed?

manthey:
The npm package added typescript support in 1.4.0 then changed to packaging with rollup in 1.5.0.  There is no functional difference, but it doesn't resolve in our examples now.

aashish24:
Sounds great,thanks.

PR #1084, opened 2021-04-26
Refactor CI to not use cmake.

manthey:
The principle benefit of using cmake was to download external test data.  This is now simply handled by scripts/datastore.js.  All of the sha512 hash references for test files have been moved there.

The secondary benefit from using cmake was integration with reporting to cdash, but we haven't actually used cdash in any meaningful way.  We don't use it to track status, so it doesn't provide any actual benefit.

The main motivator to moving away from cmake is to make it less complex to shift CI from travis-ci to another service, such as circle-ci (circle-ci has better debugging, such as sshing into it).


aashish24:
@manthey this branch has conflicts that needs to be resolved. Also, can we not use Github actions?

manthey:
`datastore.js` is largely taken from an example provided by node on how to download and verify files.  The actual url processing is taken from how the python package `pooch` does it.  There is actually less code than in the CMakeLists.txt file.

manthey:
I'll rebase and fix the conflict shortly.  We probably can use GitHub actions.  I'm less familiar with setting them up than with circle.  Either way, this PR will make switching to a different CI easier.

aashish24 (approved):

aashish24:
@manthey it seems that the pull request travis ci is stuck?

PR #1088, opened 2021-05-11
Set overflow: hidden on the main div

manthey:
This is a better default, as it prevents some features (like heatmaps) from showing outside of the div.


banesullivan (approved):
Works wonderfully!

| before | after |
|---|---|
|   |   |

PR #1091, opened 2021-06-08
Fix a few minor typos.

manthey:


manthey:
I found these running `codespell`.  Before we ran spell check, but there is a lot of noise (since variable names are often not English words), so it is sometimes hard to see the errors for the noise.  `codespell` still have false signals (and misses stuff), but fewer than just running the editor's spell check.

aashish24:
Nice, yes it seems to be so much better.

aashish24 (approved):

PR #1094, opened 2021-06-10
fix: Publish artifacts with semantic-release releases

manthey:
With semantic release, there is a different technique to publish the build artifacts to the github releases.  This switches to it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,31 +79,7 @@ jobs:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: website/public
         force_orphan: true
-  deploy-releases:
-    if: ${{ startswith(github.ref, 'refs/tags/') }}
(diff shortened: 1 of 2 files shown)


manthey:
@aashish24 If you look at the release page, there is currently a v1.0.3-beta.1 release.  You can see that it has 6 build artifacts:
image
Before, it just showed the file names:
image
With semantic-release it is easy to control the names that appear.  Maybe it would be better to have something like:
`Full JS Distribution (geo.js)`
Thoughts?

manthey:
Closes #1095.

aashish24:
I think Full JS Distribution (geo.js) will appeal better

manthey:
I agree.  I made that change in the second commit on this branch.  I think it will also now update the CHANGELOG correctly, too (but I'm not sure how test it other than by merging).

aashish24 (approved):

PR #1097, opened 2021-06-28
feat: Add a geoOnce function.

manthey:
This handles an event a single time and then deregisters the callback.  Often, a program calls `geoOn` and in the handler calls `geoOff`.  This allows just calling `geoOnce`.  In order to cancel a `geoOnce`, the return value is the function that would have to be passed to `geoOff`.

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -122,6 +122,28 @@ var object = function () {
   };
 
   /**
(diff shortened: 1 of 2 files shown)


aashish24 (approved):
Looks reasonable and I have seen this pattern elsewhere as well

PR #1100, opened 2021-07-08
feat: Allow sharing queues between tile layers.

manthey:
Although you could sort of share the fetch queue between tile layers, it wouldn't really share properly without a custom needed function.  This explicitly supports sharing a fetch queue between tile layers.

This allows, for instance, multiple layers that access the same server to have a single fetch queue so that the number of requests against the server is limited (usually to 6) rather than that number multiplied by the number of layers.

This also adds an additional reference parameter to tile labels which prevents there from being a tile string collision between layers.

The multiframe tutorial now uses this.

### DIFF
--- a/src/tileLayer.js
+++ b/src/tileLayer.js
@@ -291,6 +291,11 @@ var tileLayer = function (arg) {
   Object.defineProperty(this, 'queue', {
     get: function () { return m_this._queue; },
     set: function (queue) {
+      /* The queue's needed function determines if a tile is still needed. A
+       * tile in the queue is needed if it is needed by at least one layer that
+       * is using it.  _tileLayers tracks the layers that share the queue to
+       * allow walking through the layers and check if any layer needs a tile.
+       * When the queue is set, maintain the list of joined tile layers. */
       if (m_this._queue !== queue) {
         if (this._queue && this._queue._tileLayers && this._queue._tileLayers.indexOf(m_this) >= 0) {
           this._queue._tileLayers.splice(this._queue._tileLayers.indexOf(m_this), 1);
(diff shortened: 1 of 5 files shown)


manthey:
@aashish24 This is intended to be the first of a few improvements that will allow better buffering of multiple tile sources.  Follow on work will include an explicit A/B buffered tile layer.

src/tileLayer.js
```diff
@@ -275,6 +283,25 @@ var tileLayer = function (arg) {
...
+  /**
+   * Get/set the queue object.
+   * @property {geo.fetchQueue} queue The current queue.
+   * @name geo.tileLayer#queue
+   */
+  Object.defineProperty(this, 'queue', {
+    get: function () { return m_this._queue; },
+    set: function (queue) {
+      if (m_this._queue !== queue) {
+        if (this._queue && this._queue._tileLayers && this._queue._tileLayers.indexOf(m_this) >= 0) {
```
aashish24, line 295: I have some idea why this is necessary but wondering if you could provide some explanations?

manthey: There is a `needed` function that determines if a tile is still needed.  A tile is needed if it is needed by at least one layer that is using it.  The `_tileLayers` array is just a way to make sure we can enumerate the layers that are sharing the queue, and walk through the layers and ask if any layer needs a tile.  Setting the queue (or exiting the layer) needs to maintain the book keeping for this list of layers.

manthey: I can add that as a comment.

aashish24: yeah that makes sense. So in the next line why are we setting this index to 1? The tileLayer is already has this queue before but the index may have been zero or above zero. Just wondering.

manthey: We're not.  We are removing the layer from the old queue's _tileLayers list it was part of.  The splice(index, 1) says remove 1 value at its current location.  We then push in on the new queue's _tileLayers list at the end of the function.

aashish24: Ah thanks for reminding the function execution, okay that makes sense.

aashish24 (approved):

PR #1101, opened 2021-07-12
feat: Improve when tile layers are idle.

manthey:
Add a setting to allow tile layers to be considered idle either when all tiles are finished loading or as soon as all tiles in the view are loaded.  Before this was always equivalent to all tiles.  The new default is just tiles in the view, as this is more relevant to typical use.

feat: Allow removing promises from the idle monitor.

This will facilitate tile layers reporting that they are idle when all tiles in the view are loaded even if tiles out of the view are still loading.

### DIFF
--- a/src/tileLayer.js
+++ b/src/tileLayer.js
@@ -72,6 +72,11 @@ var featureLayer = require('./featureLayer');
  *   bottom-up (the ingcs does not matter, only the gcs coordinate system).
  *   When falsy, this inverts the gcs y-coordinate when calculating local
  *   coordinates.
+ * @property {string} [idleAfter='view'] Consider the layer as idle once a
+ *   specific set of tiles is loaded.  'view' is when all tiles in view are
+ *   loaded.  'all' is when tiles in view and tiles that were once requested
+ *   have been loaded (this corresponds to having all network activity
+ *   finished).
  */
 
 /**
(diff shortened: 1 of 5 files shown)


manthey:
@aashish24 This continues improvements that will build to an explicit A/B buffered tile layer.

src/tileLayer.js
```diff
@@ -1358,6 +1366,17 @@ var tileLayer = function (arg) {
           m_this._purge(zoom, true);
         }
       );
+    // for tiles that aren't in view, remove them from the list of tiles that
+    // are needed to be loaded to be considered idle.
+    if (m_this._options.idleAfter !== 'all') {
+      for (const hash in m_promisedTiles) {
+        const tile = m_promisedTiles[hash];
+        if (tiles.indexOf(tile) < 0) {
```
aashish24, line 1374: So is that the case that all the tiles that are not in view have index < 0, just wanted to make sure I understand the logic here.

manthey: Yes.  If a value is not in an array, indexOf returns -1.  The tiles array is the list of tiles in the view.

aashish24: Okay thanks.

aashish24 (approved):

PR #1105, opened 2021-08-06
feat: Add a baseQuad option to tileLayers.

manthey:


aashish24 (approved):
Looks reasonable to me. It would be good have a example added as well.

PR #1106, opened 2021-09-07
perf: Add a continuousCloseProximity option to annotations

manthey:


manthey:
@boucaud, If you set `continuousCloseProximity` to `true`, it will always close a dragged polygon on mouse up.  You should be able to test this with the build artifacts produced by the actions.

src/annotation.js
```diff
@@ -1152,8 +1154,18 @@ function continuousVerticesProcessAction(m_this, evt, name) {
       }
       vertices[vertices.length - 1] = evt.mouse.mapgcs;
       vertices.push(evt.mouse.mapgcs);
-      return true;
+      update = true;
     }
+    if ((ccp || ccp === 0) && evt.event === geo_event.actionup &&
+        (cpp === true || layer.displayDistance(vertices[0], null, evt.mouse.map, 'display') <= cpp)) {
```
boucaud, line 1160: Should this be `ccp === true` ?

Right now, it works but I have to set continuousPointProximity to true as well.

manthey: Yes, my mistake.

boucaud:
Thanks ! I can confirm it works.

PR #1107, opened 2021-09-23
adds basemaps from Carto

marySalvi:
Adds base maps from:
   
   https://carto.com/help/building-maps/basemap-list/#carto-vector-basemaps

Attribution modeled after documentation at:

https://carto.com/help/working-with-data/attribution/#basemaps

### DIFF
--- a/src/osmLayer.js
+++ b/src/osmLayer.js
@@ -131,6 +131,9 @@ let StamenAttribution = 'Map tiles by <a href="http://stamen.com">Stamen ' +
   'CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap' +
   '</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.';
 
+/* Per Carto's website regarding basemap attribution: https://carto.com/help/working-with-data/attribution/#basemaps */
+let CartoAttribution = '<a href="https://carto.com"> Carto</a> ' + 'Contributors <a href="https://www.openstreetmap.org/"> OpenStreetMap</a>'
+
 /**
  * This is a list of known tile sources.  It can be added to via
  * `geo.osmLayer.tilesource[<key>] = <object>`, where the object has `url`,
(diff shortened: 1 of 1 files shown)


manthey:
Since we switch to semantic-release, can you prefix the name of the PR with `perf: ` -- this will make a point release when the branch is merged,

One of our examples shows off the list of defined tilesets (see examples/measure/index.pug).  Can you add these to it?

Perhaps as a separate PR, we could add preferred human-readable labels to these definitions and then populate the example from the dictionary of values.

src/osmLayer.js
```diff
@@ -139,6 +142,18 @@ let StamenAttribution = 'Map tiles by <a href="http://stamen.com">Stamen ' +
  * @type {object}
  */
 osmLayer.tileSources = {
+  'darkMatter-with-labels': {
```
manthey, line 145: For consistency, should this be `dark-matter-with-labels` rather than camel case?

marySalvi: Yes, yes it should.

marySalvi:
I can certainly add these to the example and make those changes.

marySalvi:
I added name to the tileSources definition and changed the example to load the tileSources dynamically. Let me know what you think.

src/osmLayer.js
```diff
@@ -139,51 +142,100 @@ let StamenAttribution = 'Map tiles by <a href="http://stamen.com">Stamen ' +
  * @type {object}
  */
 osmLayer.tileSources = {
+  'dark-matter-with-labels': {
+    url: ' https://{s}.basemaps.cartocdn.com/rastertiles/dark_all/{z}/{x}/{y}.png',
+    attribution: CartoAttribution,
+    name: 'Dark Matter With Labels',
```
manthey, line 148: My only recommendation is to prefix each of the Carto names with "Carto ".

manthey (approved):

PR #1141, opened 2022-01-04
build: Switch to webpack 5.

### LINKED ISSUE #1132: Update tooling
Many of the development packages used by GeoJS are stale.  We should update them all to modern versions.  For ease of development, we should abandon using PhantomJS -- it isn't maintained and both Chrome and Firefox have sensible headless versions for testing.

manthey:
This has many fewer stale npm packages (so many fewer warnings).  Tests can use all of modern javascript.  This should be easier to maintain in the future.

Note the built bundles are somewhat smaller due to modernization of some packages.

This is part of and closes #1132.


manthey:
@aashish24 This brings the build environment up to date with modern components.  We could still upgrade bootstrap (used in the website), vtkjs, d3 (though that could be construed as a breaking change).  Also, in the last few days, jasmine made a major update, but changing to that should be a separate PR.

manthey:
Some notes about the process of upgrading from Webpack 3 to Webpack 5 along with all associated packages (karma, babel, various loaders).

- Webpack plugins changed.  We have a simple plugin for Karma that fails on any warning so that all of our tests are guaranteed to run.  This has to be refactored.
- The minification process for javascript is slow.  Uglify moving to Terser is definitely an improvement.
- Polyfills are no longer explicitly mentioned or called.  Thank you, Babel.
- Internals for webpack have changed to check if optional packages are present.
- The url-loader and file-loader parts of webpack have become simpler, if more cryptic.
- The expose-loader/import-loader chain changed a lot.  The correct syntax wasn't obvious.  It would be nice to repackage vgl to not expect global namespace entries.

PR #1152, opened 2022-01-19
perf: Add support for webgl pixelmaps.

manthey:
This also adds a 2D texture lookup table.

This is a necessary antecedent to adding tiled pixelmaps.


manthey:
In the canvas pixelmap renderer, any change to a color on the pixelmap requires rerendering much of the canvas.  In this webgl implementation, we regenerate the lookup table for the pixelmap.  This could be sped up further by only regenerating the part that has changed -- but if colors are largely duplicated, this will be fairly fast in any case.

naglepuff (approved):

PR #1153, opened 2022-01-19
feat: Support tiled pixelmaps.

manthey:
This creates a pixelmapLayer that is much like an osmLayer mixed with a pixelmap feature.

This needs:
- [x] tests
- [x] a tutorial or example
- [x] better documentation
- [x] harden against non-webgl pixelmaps (since those aren't supported)
- [x] Check if selecting a point returns the correct index
- [x] Test with a larger array

It would be nice to expose a function that would report the maximum value, but that would be expensive -- it would have to fetch all tiles and operate on each of them.


src/pixelmapFeature.js
```diff
@@ -134,6 +116,42 @@ var pixelmapFeature = function (arg) {
...
+    m_this.modified();
+    if (m_modifiedIndexRange === undefined) {
+      m_modifiedIndexRange = [idx, idx];
+    }
+    if (idx < m_modifiedIndexRange[0]) {
+      m_modifiedIndexRange[0] = idx;
+    }
+    if ((idx2 || idx) > m_modifiedIndexRange[1]) {
+      m_modifiedIndexRange[1] = (idx2 || idx);
+    }
```
naglepuff, line 151: What are the downstream consequences of calling code not abiding by the `idx < idx2` constraint? Is there a need to protect against this scenario when building `m_modifedIndexRange`?

manthey: If you pass idx2 < idx, the resulting tracked range will be `min(m_modifiedIndexRange[1], idx), max(m_modifiedIndexRange[2], idx2)`.  In the case where the index range had not previously been set, the functional result will be that no update will be required.  If the index range has previously been set, then this will possibly extend the lower end of the range but not extend the upper end.

I don't think we need to guard against it, as it isn't clear what the caller would expect if asking to update from an inverted range.  Nothing untoward will happen -- it just won't update what they expect.

There are many functions that don't strictly check input -- for instance, the indices are expected to be integers, but we don't confirm that, and, if idx is not an integer, the program throw an error.  My inclination is to leave it without an explicit check.

naglepuff: That seems reasonable to me.

naglepuff (approved):

PR #1161, opened 2022-01-27
perf: Update option for addAnnotation.

manthey:
Add an option to skip updating the layer when adding an annotation.


manthey:
@boucaud This is a fairly simple change, so I'm happy to merge it once CI has passed.

boucaud:
Thank you David !

Could we do the same thing for removeAnnotation() ?

manthey:
> Could we do the same thing for removeAnnotation() ?

Oddly, we already had that flag on `removeAnnotation`.

PR #1192, opened 2022-03-28
feat: Add square, ellipse, and circle annotation classes

manthey:
Also, if shift is held down while drawing a selection rectangle, constrain it to a square.

There is a general constraint function that can constrain to a list of aspect ratios, too.


manthey:
image

PR #1198, opened 2022-05-06
feat: Polygon operations

manthey:
This adds some polygon operation utilities.  The utilities are in `geo.util.polyops` and are `union`, `difference`, `intersect`, and `xor`.  Basic usage is `union(poly1, poly2, opts)`, where the polygons can be in a variety of formats including taken from a polygon feature.  The options allow a variety of control.  Internally, we use the polybool library, which (with the right branch) has proven to be reliable in the past.

This shows intersection two polygon features, one shown in blue and one in red, with the output in green:
Animation


manthey:
The next step is to add support for annotation layers.  This is a necessary antecedent to supporting more complex annotation editing options, such as paint brush and wand tools, and, in general behaving in expected manners with polygons with holes.

PR #1289, opened 2023-09-02
[build] Update get_tiles.py to use recommend OSM tile URL

Firefishy:
See: https://github.com/openstreetmap/operations/issues/737


manthey:
Thank you, @Firefishy.

PR #1342, opened 2024-11-15
perf: Stop using jquery extend

manthey:
Pure javascript is faster.

jquery.extend has shallow and deep versions.  All shallow versions were replaced with Object.assign.  All deep versions use a new method called `deepMerge` with seems to run in about 40% of the time the jquery deep extend runs.


manthey:
Confirmed by external source.

PR #1372, opened 2025-04-18
fix: Change modifier keys for wheel actions

manthey:
Since 2021-09 it has become the unofficial standard to send pinch events on touchpads as wheel events with the ctrl key held down.  This should correspond to zoom.  GeoJS actions predate this and had used ctrl+wheel as rotate (and wheel by itself as zoom).  This changes it so shift+wheel is rotate, and wheel+(no shift) regardless of ctrl state is zoom.

### DIFF
--- a/src/mapInteractor.js
+++ b/src/mapInteractor.js
@@ -229,7 +229,7 @@ var mapInteractor = function (args) {
       }, {
         action: geo_action.zoom,
         input: 'wheel',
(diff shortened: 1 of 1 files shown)


manthey:
See https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent for when this was adopted by different browsers

PR #1378, opened 2025-05-23
fix: Update a style from an array could fail

manthey:
When updating a style from a buffer taking into account that there are multiple vertices per pixel, the fallback lookup could be wrong.  The initial update worked, but subsequent updates could fail.


manthey:
This was unneeded.

PR #1381, opened 2025-07-08
perf: Speed up cluster zoom transition

manthey:
In a test of 1.3M points, this saves 500-700 ms per zoom transition in one test.  The clustering took a maximum of 94ms for the zoom transition; the remaining time is in the webGL point transformation and property fetches.  The next improvement would be to cache those values per zoom level.


manthey:
@annehaley This change helps, but at high zoom levels with a lot of points we are still needlessly recomputing some things.

PR #1397, opened 2025-09-04
perf: Add more information to annotation mode change events

manthey:
Include the last state of an active annotation and a reason


manthey:
@marySalvi FYI.

PR #1400, opened 2025-09-09
perf: Add more information to annotation mode change events

manthey:
Specifically, this add "oldCoordinates" to the event. If there was an active annotation before the mode change, these are the annotation's coordinates before the change.  This will be an empty list if an annotation in create state had not been started and a non-empty list if it is partially created.

### DIFF
--- a/src/annotationLayer.js
+++ b/src/annotationLayer.js
@@ -707,9 +707,10 @@ var annotationLayer = function (arg) {
       } else {
         m_keyHandler.unbind('esc');
       }
-      let oldState;
+      let oldState, oldCoordinates;
       if (m_this.currentAnnotation) {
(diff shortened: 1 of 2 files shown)


src/annotationLayer.js
```diff
@@ -757,8 +758,9 @@ var annotationLayer = function (arg) {
           m_this.map().interactor().addAction(action);
         });
       }
+      console.log({mode: m_mode, oldMode, oldState, oldCoordinates, reason});
```
marySalvi, line 761: @manthey Was this meant to stay in?

manthey: Nope.  And pre-commit and CI both should have refused it...

PR #1476, opened 2026-04-21
fix contextrestored webgl renderer call

BryonLewis:
Noticed a frequent issue in BatAI with this specific error:
`d.renderWindow(...)._init is not a function`

With this stack trace:
```
    m_contextRenderer.setResetScene(false);
    canvas.get(0).addEventListener('webglcontextlost', function (evt) {
      return evt.preventDefault();
    }, false);
    canvas.get(0).addEventListener('webglcontextrestored', function () {
      return m_viewer.renderWindow()._init();
    }, false);
    if (m_viewer.renderWindow().renderers().length > 0) {
      m_contextRenderer.setLayer(m_viewer.renderWindow().renderers().length);
    }
    m_this.canvas(canvas);
```

The specific line is here: https://github.com/OpenGeoscience/geojs/blob/master/src/webgl/webglRenderer.js#L96

I'm guessing it can be happening with annotators and BatAI because of the rather large webGL context it can jump between in a SPA (very large images tiled together along with a decent amount of line annotations).

The problem is that it apppears that the `m_viewer.renderWindow()` doesn't have a function called `._init()`
See: https://github.com/OpenGeoscience/geojs/blob/master/src/vgl/renderWindow.js
Specifically, there seems to be a function called `_setup` and not `_init` at: https://github.com/OpenGeoscience/geojs/blob/master/src/vgl/renderWindow.js#L131

It looks like items outside of `/vgl` use `_init()` frequently but items inside of that folder seem to use `_setup` more often.

20260423 Update:

- Added dirty context management to the shader programs and renderers
- This will probably redraw the map when context is restored as well
- Added a test for the functionality of losing and restoring the webgl context through the `WEBGL_lose_context` extension for WebGL: https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_lose_context

### DIFF
--- a/src/vgl/mapper.js
+++ b/src/vgl/mapper.js
@@ -38,8 +38,12 @@ vgl.mapper = function (arg) {
    */
   this.deleteVertexBufferObjects = function (renderState) {
     var i;
+    if (renderState && renderState.m_contextChanged) {
+      // The old context is already gone, so deleting its buffers is invalid.
+      return;
+    }
     var context = m_context;
-    if (renderState) {
+    if (renderState && !m_context) {
       context = renderState.m_context;
     }
     if (context) {
--- a/src/vgl/renderWindow.js
+++ b/src/vgl/renderWindow.js
@@ -126,9 +126,11 @@ vgl.renderWindow = function (canvas) {
    * Create the window.
    *
    * @param {vgl.renderState} renderState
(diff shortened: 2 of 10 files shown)


manthey (approved):
Excellent.  Thanks for working through this.

PR #1497, opened 2026-08-24
Translate fixed shapes

bnmajor:
[Screencast from 08-25-2026 12:41:33 PM.webm](https://github.com/user-attachments/assets/1b909b84-f678-43cd-a878-4861391515c2)

### DIFF
--- a/src/annotation/annotation.js
+++ b/src/annotation/annotation.js
@@ -1358,20 +1358,10 @@ function continuousVerticesProcessAction(m_this, evt, name) {
  *   wide or half as wide as it is tall.  Sizes (e.g., {width: 400, height:
  *   500}) snap to that size.
  * @returns {Function} A function that can be passed to the mapIterator
- *   selectionConstraint or to an annotation constraint function.  If any sizes
- *   are given, the returned function is stateful: it remembers where the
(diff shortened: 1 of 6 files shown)


manthey:
This works as requested, but somehow revealed that I had behavior opinions I didn't realize I had until using it.

For non-fixed size shapes, we either click and drag or click multiple times.

For click-and-drag this works mostly as I expect, but I sort of expected either I would drag the center rather than a corner or it would sort of follow the mouse such that if the mouse would leave the annotation it would follow.  

For the click opposite corners, this is accentuated even more -- my first try was to click what I assumed was the top left and then move down and to the right. Since on the web-Mercator map we have a right-handed coordinate system, we end up creating the bottom left first.

I'm not sure if we should change the current behavior, and, if so, what it should be.

manthey:
Having tried these all out, the center option feels the most reasonable.  I'd love to have the other commits left in as part of the history of this, though.

bnmajor:
@manthey I've kept the changes as a part of the git history and added tests and updated the tutorial documentation. The behavior now uses the center option.

manthey (approved):

